PR #36, opened 2022-11-07
support ingress secret&skip mcp validate

johnlanni:
1. skip mcp validate (TODO: need optimize ingress convert logic)
2. allow higress gateway access secrets in other namespace

### DIFF
--- a/helm/higress/values.yaml
+++ b/helm/higress/values.yaml
@@ -17,6 +17,8 @@ gateway:
     # If enabled, roles will be created to enable accessing certificates from Gateways. This is not needed
     # when using http://gateway-api.org/.
     enabled: true
+    # If enabled, gateway can use cross namespace secrets
+    crossNamespaceSecret: true
 
   serviceAccount:
     # If set, a service account will be created. Otherwise, the default is used
--- a/helm/higress/values.yaml
+++ b/helm/higress/values.yaml
@@ -17,8 +17,6 @@ gateway:
     # If enabled, roles will be created to enable accessing certificates from Gateways. This is not needed
     # when using http://gateway-api.org/.
     enabled: true
-    # If enabled, gateway can use cross namespace secrets
-    crossNamespaceSecret: true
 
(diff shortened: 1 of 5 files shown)


helm/higress/values.yaml
```diff
@@ -17,6 +17,8 @@ gateway:
     # If enabled, roles will be created to enable accessing certificates from Gateways. This is not needed
     # when using http://gateway-api.org/.
     enabled: true
+    # If enabled, gateway can use cross namespace secrets
+    crossNamespaceSecret: true
```
SpecialYang, line 21: Where is the parameter used for?

Kaushikrudra (approved):

helm/higress/values.yaml
johnlanni: The helm template will use this to determine whether to install the ClusterRole resource.

johnlanni: @SpecialYang The option is removed to avoid ambiguity about the cross namespace usage of secret.

istio/1.12/patches/istio/20221107-support-ingress-secret.patch
```diff
@@ -0,0 +1,37 @@
+diff -Naur istio/pilot/pkg/model/credentials/ali_resource.go new/pilot/pkg/model/credentials/ali_resource.go
+--- istio/pilot/pkg/model/credentials/ali_resource.go	2022-11-07 21:40:10.000000000 +0800
++++ new/pilot/pkg/model/credentials/ali_resource.go	2022-11-07 19:03:22.000000000 +0800
```
Lynskylate, line 3: 后面修改为higress_resource?

Lynskylate (approved):

istio/1.12/patches/istio/20221107-support-ingress-secret.patch
johnlanni: OK

PR #40, opened 2022-11-08
optimize hpa & pdb version select

johnlanni:

### DIFF
--- a/helm/istiod/values.yaml
+++ b/helm/istiod/values.yaml
@@ -2,7 +2,7 @@
 
 ## Discovery Settings
 pilot:
-  autoscaleEnabled: true
+  autoscaleEnabled: false
   autoscaleMin: 1
   autoscaleMax: 5
   replicaCount: 1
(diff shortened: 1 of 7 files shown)


iutx:
👍It's handled very quickly. 
The same commit.I started by trying to abstract the common library package, referenced by higress and istio. However, istio officials have already processed it, and only need add an hpa template to higress.

https://github.com/iutx/higress/commit/83a08286f8775683f30e6fd2ac11f98dca889e80

helm/istiod/values.yaml
```diff
@@ -221,6 +221,9 @@ meshConfig:
   # No hurry to do this in 1.6, we're trying to prove the code.
 
 global:
+  # whether to use autoscaling/v2 template for HPA settings
+  # for internal usage only, not to be configured by users.
+  autoscalingv2API: true
```
SpecialYang, line 226: Can we hide it? Can we determine which version should be used based-on k8s cluster version?

SpecialYang (changes requested):

SpecialYang (approved):

PR #51, opened 2022-11-10
change route name

johnlanni:

change the ingress route name


SpecialYang (approved):
Currently, we don't use the route name for the specific route so this pr is fine. I think we should document it.

PR #52, opened 2022-11-10
add istio patch to fix xds push

johnlanni:
The previous hack implementation caused LDS to be pushed empty


SpecialYang (changes requested):
Don't understand why use the mcp generator directly. Since the resources were sent to mcp store, why not use the xds generator?

johnlanni:
> Don't understand why use the mcp generator directly. Since the resources were sent to mcp store, why not use the xds generator?

Although these two  `Resource` types are all convert to `[]Any` in the DiscoveryResponse, xds generator generates the `Resource` which is defined by envoy's api in `discovery/v3`, and mcp generator generates the `Resource` which is defined by istio's api in `mcp/v1alpha1`.

SpecialYang:
What's protocol between envoy and istio? xds or mcp-over-xds?

johnlanni:
> What's protocol between envoy and istio? xds or mcp-over-xds?

I understand your doubts, in fact, istio will not call pushMcpXds here, only higress will, here is higress multiplexing the logic related to xds

SpecialYang (approved):
We should reconsider this way in the future. It's better to build our xds server.

PR #108, opened 2022-12-10
Support nacos discovery

johnlanni:


SpecialYang:
Maybe we should have a design document like yuque workspace. So reviewers can easily get the core concept and review codes.

johnlanni:
> Maybe we should have a design document like yuque workspace. So reviewers can easily get the core concept and review codes.

Good suggestion, otherwise it would be difficult for external developers to get involved. 
As for the review, let’s discuss this implementation face to face next week. 😄

SpecialYang:
We can create the public workspace in public yuque (https://www.yuque.com/).  I know nacos community does the same thing.

SpecialYang (approved):

PR #155, opened 2023-02-01
Default set ingress class as higress and enable ingress status.

SpecialYang:
As a new ingress controller,  we should have our own ingress class. 
For migration case, we also have settings for explicitly changing ingress class.

Associated issuse https://github.com/alibaba/higress/issues/145.


johnlanni:
https://github.com/alibaba/higress/blob/main/test/ingress/conformance/tests/httproute-simple-same-namespace.yaml

@SpecialYang This test case needs to be updated.

johnlanni (approved):
LGTM

PR #157, opened 2023-02-02
add wasi stub patch

johnlanni:


Lynskylate (approved):
LGTM.

But I think fork should be used instead of patch code modification as soon as possible

PR #160, opened 2023-02-06
e2e: add host match routing e2e case

Xunzhuo:
Signed-off-by: bitliu <bitliu@tencent.com>

Add hostname match in the same namespace e2e test case.


johnlanni (approved):
LGTM

PR #213, opened 2023-02-21
docs: translate the document of jwt-auth plugin readme to English. 

### LINKED ISSUE #73: jwt-auth 插件的 README 文档翻译
https://github.com/alibaba/higress/blob/main/plugins/wasm-cpp/extensions/jwt_auth

PerforMance308:
translate the document of jwt-auth plugin readme to English. 

fixes #73


plugins/wasm-cpp/extensions/jwt_auth/README_EN.md
```diff
@@ -0,0 +1,399 @@
...
+
+**Payload**
+
+The payload is where the actual information is stored. The details are defined below:
+
+```text
+iss: The issuer of the token. This indicates who created the token and is a string value.
+sub: The subject identifier. This is the unique identifier for the end user provided by the issuer, and is no longer than 255 ASCII characters, and is case-sensitive within the issuer's scope.
+aud: The audience(s) of the token, which is an array of case-sensitive strings.
+exp: The expiration time of the token, which is a Unix timestamp in seconds, and any token that is issued after that time is invalid.
```
CH3CHO, line 91: `any token that is issued after that time is invalid`, I don't think it is correct.

plugins/wasm-cpp/extensions/jwt_auth/README_EN.md
```diff
@@ -0,0 +1,399 @@
...
+iat: The time at which the token was issued, also a Unix timestamp in seconds.
```
CH3CHO, line 92: at which or when? And do we need an "is" before "also"?

plugins/wasm-cpp/extensions/jwt_auth/README_EN.md
```diff
@@ -0,0 +1,399 @@
...
+jti: The unique identifier for the token, and the value is unique for every token created by the issuer. It is usually a cryptographically random value to prevent conflicts. This component adds a random entropy that an attacker cannot obtain to the structured token, making it more difficult for the token to be guessed or replayed.
```
CH3CHO, line 93: I would prefer `idenifier "of" the token`.

plugins/wasm-cpp/extensions/jwt_auth/README_EN.md
```diff
@@ -0,0 +1,399 @@
...
+```
+
+Custom fields for a user system can also be added, such as the example below adding a "name" field for the user's nickname:
```
CH3CHO, line 96: `Custom fields for a user system`, what does "for" mean here, and why it's "a user system"?

plugins/wasm-cpp/extensions/jwt_auth/README_EN.md
```diff
@@ -0,0 +1,399 @@
...
+**Signature**
+
+This part is a string that consists of the Base64-encoded header and Base64-encoded payload concatenated with a period, followed by the encryption of the resulting string using the algorithm specified in the header (where $secret represents the user's private key).
+
+```js
+var encodedString = base64UrlEncode(header) + '.' + base64UrlEncode(payload);
+var signature = HMACSHA256(encodedString, '$secret');
+```
+
+These three parts are then concatenated using periods to form the complete JWT string as shown in the example in section 1.3.2.
```
CH3CHO, line 118: It sounds strange because we are still in section 1.3.2 here. The original text says "就构成了 1.3.2 节最开始的JWT示例". Do you think the meaning of "最开始" is mandatory here?

plugins/wasm-cpp/extensions/jwt_auth/README_EN.md
```diff
@@ -0,0 +1,399 @@
...
+
+#### 1.3.3 Time validity
+
+The gateway will verify the exp field in the token. Once this field has expired, the gateway will consider the token invalid and reject the request directly. The expiration time value must be set.
+
+1.3.4 Several Characteristics of JWT
```
CH3CHO, line 124: Do you need a header style for 1.3.4?

plugins/wasm-cpp/extensions/jwt_auth/README_EN.md
```diff
@@ -0,0 +1,399 @@
...
+
+1. By default, JWT is not encrypted and cannot write secret data into JWT.
+2. JWT can not only be used for authentication but also for exchanging information. Using JWT effectively can reduce the number of times servers query the database.
+3. The biggest drawback of JWT is that the server cannot invalidate a token during use, or change the token's permission, because the server does not keep the session state. That is, once JWT is issued, it will always be valid before it expires, unless the server deploys additional logic.
+4. JWT contains authentication information itself. Once leaked, anyone can obtain all permissions of the token. To reduce theft, the expiration time of JWT should be set relatively short. For some more important permissions, users should be authenticated again when using them.
+5. To reduce theft, JWT should not be transmitted in plaintext using the HTTP protocol, and the HTTPS protocol should be used for transmission.
```
CH3CHO, line 130: For list items like these, should we end them with a stop or semicolon?

plugins/wasm-cpp/extensions/jwt_auth/README_EN.md
```diff
@@ -0,0 +1,399 @@
...
+
+## 2. How to apply the JWT plugin to protect the API of the user system
+
+### 2.1 Generate a pair of JWK (JSON Web Key)
+
+**Method 1: Generate online**
+
+Users can generate the private and public keys used for token generation and verification on this website https://mkjwk.org. The private key is used for the authorization service to issue JWT, and the public key is configured into the JWT plugin for the gateway to verify the signature of the request. Note that the jwks format configuration used by the gateway requires the Public Key in the figure below to be placed in the keys structure, such as: `{"keys":[{"kty":"RSA","e":"AQAB",...}]}`
```
CH3CHO, line 138: `jwks`, `Public key`. I think the letter cases should be consistent throughout the document.

plugins/wasm-cpp/extensions/jwt_auth/README_EN.md
```diff
@@ -0,0 +1,399 @@
...
+```
+
+# Common Error Codes
+| 
+HTTP Status Code | Error Message               | Reason Description|
+| ----------- | ---------------------- | -------------------------------------------------------------------------------- |
+| 401         | Jwt missing            | The JWT is not provided in the request header. |
+| 401         | Jwt expired            | The JWT has expired. |
+| 401         | Jwt verification fails | The JWT payload verification failed, such as when the iss claim does not match. |
+| 403         | Access Denied          | Access to the current route is denied. |
```
CH3CHO, line 399: Why are we using title case for "Access Denied", but not for other messages above? And the word JWT, as an abbreviation, should be always in uppercase.

plugins/wasm-cpp/extensions/jwt_auth/README_EN.md
```diff
@@ -0,0 +1,399 @@
...
+    }
+```
+
+# Common Error Codes
+| 
+HTTP Status Code | Error Message               | Reason Description|
+| ----------- | ---------------------- | -------------------------------------------------------------------------------- |
+| 401         | Jwt missing            | The JWT is not provided in the request header. |
+| 401         | Jwt expired            | The JWT has expired. |
+| 401         | Jwt verification fails | The JWT payload verification failed, such as when the iss claim does not match. |
```
CH3CHO, line 398: `such as when`, I don't think it is right from the grammar standpoint, although we can live with it.

CH3CHO (changes requested):

plugins/wasm-cpp/extensions/jwt_auth/README_EN.md
PerforMance308: It is generally recommended to end each sentence with a period for consistency and clarity when using bullet points.

plugins/wasm-cpp/extensions/jwt_auth/README_EN.md
```diff
@@ -0,0 +1,399 @@
+# Description
+The `jwt-auth` plugin implements authentication and authorization based on JWT(JSON Web Tokens), supports parsing JWTs from URL parameters, request headers, and Cookie fields from HTTP requests, and verifies whether the token has permission to access.
+
+The difference between this plugin and the JWT authentication in `Security Capabilities->Authentication and Authorization` is that it provides additional capabilities for identifying the caller's identity, supporting the configuration of different JWT credentials for different callers.
```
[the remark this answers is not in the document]
PerforMance308: This make sense, I changed requester to caller.

plugins/wasm-cpp/extensions/jwt_auth/README_EN.md
PerforMance308: I changed `jwks` to `JWKs` and Public key to lower case to be consistent throughout the document.

plugins/wasm-cpp/extensions/jwt_auth/README_EN.md
```diff
@@ -0,0 +1,399 @@
...
+
+8. The gateway returns the business response to the client.
+
+In this entire process, the gateway uses the token authentication mechanism to realize the ability of users to use their own user system to authorize their API. Next, we will introduce the structured token Json Web Token (JWT) used by the gateway to implement token authentication.
+
+### 1.3 JWT
+
+#### 1.3.1 Introduction
+
+Json Web Token (JWT) is an open standard RFC7519 based on JSON for executing a type of claim to pass between network applications. JWT can generally be used as an independent identity verification token, which can contain user identification, user roles, and permissions information, making it easier to obtain resources from the resource server, and can also add some other necessary declarative information for other business logic, especially suitable for the login scenario of distributed sites.
```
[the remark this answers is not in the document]
PerforMance308: Logic is countable, but in the context of `business logic`, `business logic` refers to the set of rules and processes. "Logics" can be used as a plural form of "logic", but less commonly used in this context, in most cases, "business logic" is the more appropriate and commonly used term.

plugins/wasm-cpp/extensions/jwt_auth/README_EN.md
```diff
@@ -88,12 +88,12 @@ The payload is where the actual information is stored. The details are defined b
 iss: The issuer of the token. This indicates who created the token and is a string value.
 sub: The subject identifier. This is the unique identifier for the end user provided by the issuer, and is no longer than 255 ASCII characters, and is case-sensitive within the issuer's scope.
 aud: The audience(s) of the token, which is an array of case-sensitive strings.
-exp: The expiration time of the token, which is a Unix timestamp in seconds, and any token that is issued after that time is invalid.
-iat: The time at which the token was issued, also a Unix timestamp in seconds.
-jti: The unique identifier for the token, and the value is unique for every token created by the issuer. It is usually a cryptographically random value to prevent conflicts. This component adds a random entropy that an attacker cannot obtain to the structured token, making it more difficult for the token to be guessed or replayed.
+exp: The expiration time of the token, after which the token will be invalidated, is a integer declaration representing the number of seconds since January 1, 1970.
```
CH3CHO, line 91: "an" integer.

CH3CHO: And I think Unix timestamp in seconds is good enough. Otherwise, you need to specifiy both time and timezone, which is 00:00 UTC.

plugins/wasm-cpp/extensions/jwt_auth/README_EN.md
```diff
@@ -88,12 +88,12 @@ The payload is where the actual information is stored. The details are defined b
...
+iat: The issuance time of the token, is a integer declaration representing the number of seconds since January 1, 1970.
```
CH3CHO, line 92: See comments above.

plugins/wasm-cpp/extensions/jwt_auth/README_EN.md
```diff
@@ -393,7 +393,7 @@ consumers:
...
 HTTP Status Code | Error Message               | Reason Description|
 | ----------- | ---------------------- | -------------------------------------------------------------------------------- |
-| 401         | Jwt missing            | The JWT is not provided in the request header. |
-| 401         | Jwt expired            | The JWT has expired. |
-| 401         | Jwt verification fails | The JWT payload verification failed, such as when the iss claim does not match. |
-| 403         | Access Denied          | Access to the current route is denied. |
\ No newline at end of file
+| 401         | JWT missing            | The JWT is not provided in the request header. |
+| 401         | JWT expired            | The JWT has expired. |
+| 401         | JWT verification fails | The JWT payload verification failed, such as the iss iss mismatch. |
```
CH3CHO, line 398: Why is there two "iss"es here?

CH3CHO (changes requested):

CH3CHO (approved):
LGTM. Thanks.

PR #254, opened 2023-03-24
Add header control

SpecialYang:
Issue https://github.com/alibaba/higress/issues/165.


Xunzhuo (commented):
I think we need e2e for response header operation too.

johnlanni:
> I think we need e2e for response header operation too.

It is an abiliity that has been verified internally. I think it may be better to leave e2e to the community to complete😄

johnlanni (approved):
LGTM

PR #260, opened 2023-03-27
Fix the problem that ignoreUriCase does not work when the path type is prefix

### LINKED ISSUE #258: The path prefix match uses a regular pattern, which results in ignoring case not work.
Now: 

prefix match of "/abc" -> regex match of "/abc(\\/.*)?"

Expect: 

prefix match of "/abc" -> prefix match of "/abc/"   and  exact match of "/abc"

Charlie17Li:

fix prefix path match bug

fixes #258 

1. the `/abc` can match `/abc` and `/abcd` in virtual servcie
2. the `/abc/` can match `/abc/*`, but cannot match `/abc` in virtual servcie
3. but `/abc/` should also match `/abc` in [ingress documetation](https://kubernetes.io/docs/concepts/services-networking/ingress/#examples)
4. so I both convert them(`/abc, /abc/`) to prefix `/abc/` and exact `/abc`


johnlanni:
cc @SpecialYang

SpecialYang (changes requested):
We respect the spec of Ingress not VirtualService.

pkg/ingress/kube/ingressv1/controller.go
```diff
@@ -539,9 +542,22 @@ func (c *controller) ConvertHTTPRoute(convertOptions *common.ConvertOptions, wra
 							MatchType: &networking.StringMatch_Prefix{Prefix: path},
 						}
 					} else {
-						path = strings.TrimSuffix(path, "/")
+						// the path maybe prefix /xxx or prefix /xxx/
+						// we both convert it to prefix /xxx/ and exact /xxx
+						var extraExactPath string
+						if path[len(path)-1] != '/' { // prefix /xxx
```
johnlanni, line 548: Why not trim the possible suffix "/" first, so the logic will look clearer

johnlanni (changes requested):

pkg/ingress/kube/ingressv1/controller.go
```diff
@@ -594,6 +610,14 @@ func (c *controller) ConvertHTTPRoute(convertOptions *common.ConvertOptions, wra
...
 		}
 
+		for _, httpPath := range rule.HTTP.Paths {
+			convertPath(httpPath)
+		}
+
+		for _, httpPath := range extraExactPaths {
+			convertPath(httpPath)
+		}
+
```
johnlanni, line 620: It's a bit tricky here, you'd better convert the ingress path in one step

johnlanni (changes requested):

Charlie17Li:
> We respect the spec of Ingress not VirtualService.

Yes, so I convert prefix `/abc/` to prefix `/abc/` + exact `/abc`, otherwise it will not follow ingress spec

pkg/ingress/kube/ingressv1/controller.go
Charlie17Li: done

johnlanni:
> It has already been supported in istio-envoy [istio/istio#38233](https://github.com/istio/istio/pull/38233). Not sure whether the current verison of higress supports.

Not supported now, higress is based on envoy 1.20. Wait for the overall upgrade to support this

pkg/ingress/kube/annotations/canary.go
```diff
@@ -133,39 +133,41 @@ func ApplyByHeader(canary, route *networking.HTTPRoute, canaryIngress *Ingress)
...
-			canaryConfig.Header: {
-				MatchType: &networking.StringMatch_Exact{
-					Exact: "always",
-				},
-			},
-		}
-		if canaryConfig.HeaderValue != "" {
-			canary.Match[0].Headers = map[string]*networking.StringMatch{
+	for _, match := range canary.Match {
+		if canaryConfig.Header != "" {
```
johnlanni, line 137: This `if` should be above.

johnlanni (changes requested):

pkg/ingress/kube/ingressv1/controller.go
```diff
@@ -507,47 +506,44 @@ func (c *controller) ConvertHTTPRoute(convertOptions *common.ConvertOptions, wra
 		}
 
 		wrapperHttpRoutes := make([]*common.WrapperHTTPRoute, 0, len(rule.HTTP.Paths))
+
 		for _, httpPath := range rule.HTTP.Paths {
+			var pathType common.PathType
```
SpecialYang, line 511: Also should add this support for ingress v1beta1.

pkg/ingress/kube/ingressv1/controller.go
```diff
@@ -507,47 +506,44 @@ func (c *controller) ConvertHTTPRoute(convertOptions *common.ConvertOptions, wra
...
-				httpMatch.Uri = &networking.StringMatch{
-					MatchType: &networking.StringMatch_Regex{Regex: httpPath.Path + ".*"},
-				}
-			} else {
-				switch *httpPath.PathType {
-				case ingress.PathTypeExact:
+			// if the pathType is Prefix, it will have 2 httpMatches, the first is the type of Exact and the next is Prefix
+			// so if the pathType is Prefix, the wrapperHttpRoute.OriginPathType is still Prefix, though it has been assigned twice
+			httpMatches := c.generateHttpMatches(pathType, httpPath.Path, wrapperVS)
+			for _, httpMatch := range httpMatches {
```
SpecialYang, line 533: Here no need to traverse the httpMatches. We treat these two (at most) routes as one route for RouteCache to show for console.

The users of console still get the only one route, just the prefix match.

Charlie17Li: I'm not sure if I misunderstood your point🤔, but my understanding is to not traverse through httpMatches, but to only keep the prefix match. 

But for example, if it's `prefix=/abc` or `prefix=/abc/` in Ingress, and we only keep the prefix match, then in the `wrapperHttpRoute.HTTPRoute.Match`, there will only be `prefix=/abc/`. In the current Higress version, this maybe not match the request `path=/abc`.

pkg/ingress/kube/ingressv1/controller.go
Charlie17Li: ok

pkg/ingress/kube/ingress/controller.go
```diff
@@ -526,46 +525,43 @@ func (c *controller) ConvertHTTPRoute(convertOptions *common.ConvertOptions, wra
 
 		wrapperHttpRoutes := make([]*common.WrapperHTTPRoute, 0, len(rule.HTTP.Paths))
 		for _, httpPath := range rule.HTTP.Paths {
+			var pathType common.PathType
+			if wrapper.AnnotationsConfig.NeedRegexMatch() {
```
SpecialYang, line 529: Here we can determine the value of original path.

Charlie17Li: get, so I should convert the original path to httpMatch here and remove generateHttpMatches func. then we don't need to tarverse http matchs below.

pkg/ingress/kube/ingress/controller.go
```diff
@@ -532,40 +531,25 @@ func (c *controller) ConvertHTTPRoute(convertOptions *common.ConvertOptions, wra
...
-							MatchType: &networking.StringMatch_Regex{Regex: regexp.QuoteMeta(path) + common.PrefixMatchRegex},
-						}
-					}
+					pathType = common.Prefix
 				}
 			}
-			wrapperHttpRoute.OriginPath = path
-			wrapperHttpRoute.HTTPRoute.Match = []*networking.HTTPMatchRequest{httpMatch}
+			wrapperHttpRoute.OriginPathType = pathType
+			wrapperHttpRoute.OriginPath = httpPath.Path
```
SpecialYang, line 547: For backward compatibility, we should assgin the trimmed path to OriginPath. i.e, the path `/path/` should be `/path`.

```
case ingress.PathTypePrefix:
	if path != "/" {
      path = strings.TrimSuffix(path, "/")
}
```

Charlie17Li: get

pkg/ingress/kube/ingress/controller.go
```diff
@@ -532,40 +531,25 @@ func (c *controller) ConvertHTTPRoute(convertOptions *common.ConvertOptions, wra
...
+
+			httpMatches := c.generateHttpMatches(pathType, httpPath.Path, wrapperVS)
+			for _, httpMatch := range httpMatches {
+				wrapperHttpRoute.HTTPRoute.Match = append(wrapperHttpRoute.HTTPRoute.Match, httpMatch)
```
SpecialYang, line 551: Can we simplely assgin httpMatches to `wrapperHttpRoute.HTTPRoute.Matc` ?

Charlie17Li: get!

pkg/ingress/kube/ingressv1/controller.go
```diff
@@ -507,47 +506,37 @@ func (c *controller) ConvertHTTPRoute(convertOptions *common.ConvertOptions, wra
...
 				HTTPRoute:     &networking.HTTPRoute{},
 				WrapperConfig: wrapper,
 				Host:          rule.Host,
 				ClusterId:     c.options.ClusterId,
 			}
-			httpMatch := &networking.HTTPMatchRequest{}
 
-			path := httpPath.Path
+			var pathType common.PathType
+			var originPath string
```
SpecialYang, line 519: originPath := httpPath.Path

pkg/ingress/kube/ingressv1/controller.go
```diff
@@ -507,47 +506,37 @@ func (c *controller) ConvertHTTPRoute(convertOptions *common.ConvertOptions, wra
...
 			if wrapper.AnnotationsConfig.NeedRegexMatch() {
-				wrapperHttpRoute.OriginPathType = common.Regex
-				httpMatch.Uri = &networking.StringMatch{
-					MatchType: &networking.StringMatch_Regex{Regex: httpPath.Path + ".*"},
-				}
+				pathType = common.Regex
+				originPath = httpPath.Path
```
SpecialYang, line 522: Can remove.

pkg/ingress/kube/ingressv1/controller.go
```diff
@@ -507,47 +506,37 @@ func (c *controller) ConvertHTTPRoute(convertOptions *common.ConvertOptions, wra
...
-					wrapperHttpRoute.OriginPathType = common.Prefix
-					// borrow from implement of official istio code.
-					if path == "/" {
-						wrapperVS.ConfiguredDefaultBackend = true
-						// Optimize common case of / to not needed regex
-						httpMatch.Uri = &networking.StringMatch{
-							MatchType: &networking.StringMatch_Prefix{Prefix: path},
-						}
+					pathType = common.Prefix
+					if httpPath.Path != "/" {
```
SpecialYang, line 530: If path is '/', the origin path will be empty. So you should assgin first, just like mentioned above.

pkg/ingress/kube/ingressv1/controller.go
Charlie17Li: get

SpecialYang (approved):
Looks good. Thanks.

johnlanni (approved):

PR #263, opened 2023-03-31
feautre: support ext-authz by annotation  

zhangcly:
parse higress anno and transfer to envoyfilter(ext-authz and rbac http filter) .
test cases added.
the issue haven't been resloved yet. 

#207


SpecialYang:
It is awsome 👍🏻.

But I want to know how you control the route match order. Given one request /api/test, if we have two ingress, /api with prefix matching and /api/test with exact matching, how you control this request matching the right rbac and external authz policy because these resources have their own matching list.

johnlanni:
> It is cool 👍🏻.
> 
> But I want to know how you control the route match order. Given one request /api/test, if we have two ingress, /api with prefix matching and /api/test with exact matching, how you control this request matching the right rbac and external authz policy because these resources have their own matching list.

You are right, this design to implement ext-authz via ingress annotation was my mistake, it seems we need a CRD to do this.
And there is another option: it can be extended based on envoy rbac [permission](https://www.envoyproxy.io/docs/envoy/v1.20.3/api-v3/config/rbac/v3/rbac.proto#envoy-v3-api-msg-config-rbac-v3-permission) matcher to achieve matching based on route name

PR #274, opened 2023-04-04
feat: add support for `hgctl gateway-config` to retrieve higress gateway config

Xunzhuo:

feat: add support for `hgctl gateway-config` to debug higress gateway xds config

1. support retrieve all configDump from specific higress gateway pod
2. support retrieve bootstrap config from specific higress gateway pod
3. support retrieve cluster config from specific higress gateway pod
4. support retrieve endpoint config from specific higress gateway pod
5. support retrieve listener config from specific higress gateway pod
6. support retrieve route config from specific higress gateway pod

xref: #229

image

image

examples:
image

### DIFF
--- a/pkg/cmd/hgctl/configBootstrap.go
+++ b/pkg/cmd/hgctl/configBootstrap.go
@@ -1,0 +1,65 @@
+// Copyright (c) 2022 Alibaba Group Holding Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package hgctl
+
+import (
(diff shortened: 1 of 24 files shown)


johnlanni:
cc @SpecialYang @gengleilei

pkg/cmd/hgctl/configBootstrap.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+func bootstrapConfigCmd() *cobra.Command {
+	configCmd := &cobra.Command{
+		Use:     "bootstrap <pod-name>",
```
johnlanni, line 26: Higress is focused on gateways, maybe we don't need to specify pod name? Just select the random pod of higress gateway?

johnlanni (changes requested):

pkg/cmd/hgctl/configBootstrap.go
Xunzhuo: This is a operating tools, which means operators/developers may debug some certain higress gateway pod, like if the number of replicas is more than one. A random pod is not a good idea to debug.

Xunzhuo: A better idea is to make this optional. If the pod name is empty, config retrieves from one of the higress gateway pods in a namespace.

johnlanni (approved):
LGTM

Issue #71, opened 2022-11-30
key-auth 插件的 README 文档翻译


johnlanni:
key-auth 插件的 README 文档翻译
https://github.com/alibaba/higress/blob/main/plugins/wasm-cpp/extensions/key_auth

Fkbqf:
Is this task to translate the key-auth plugin's README document into English?

CH3CHO:
> Is this task to translate the key-auth plugin's README document into English?

You bet!

PR #279, opened 2023-04-07
README translation of key-auth plugin

Fkbqf:

README translation of the key-auth plugin

#71

### DIFF
--- a/plugins/wasm-cpp/extensions/key_auth/README_EN.md
+++ b/plugins/wasm-cpp/extensions/key_auth/README_EN.md
@@ -1,0 +1,123 @@
+# Features
+The `key-auth` plug-in implements the authentication function based on the API Key, supports parsing the API Key from the URL parameter or request header of the HTTP request, and verifies whether the API Key has permission to access.
+
+# Configuration field
+
+|   Name      | Data Type       |               Parameter requirements                    | Default|     Description                                                                                          |
+| ----------- | --------------- | --------------------------------------------------------| ------ | ---------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


plugins/wasm-cpp/extensions/key_auth/README_EN.md
```diff
@@ -0,0 +1,123 @@
...
+**Set the API Key in the url parameter**
+```bash
+curl  http://xxx.hello.com/test?apikey=2bda943c-ba2b-11ec-ba07-00163e1250b5
+```
+**Set the API Key in the http request header**
+```bash
+curl  http://xxx.hello.com/test -H 'x-api-key: 2bda943c-ba2b-11ec-ba07-00163e1250b5'
+```
+
+After the authentication is passed, a `X-Mse-Consumer` field will be added to the header of the request. In this example, its value is `consumer1`, which is used to identify the name of the caller
```
CH3CHO, line 82: Before letter "X", we should use "an" instead of "a". And there should be a stop (".") at the end of sentence.

plugins/wasm-cpp/extensions/key_auth/README_EN.md
```diff
@@ -0,0 +1,123 @@
...
+
+### The following requests will deny access：
+
+**The request does not provide an API Key, returns 401**
```
CH3CHO, line 86: What is the subject of "returns 401"?

CH3CHO (changes requested):

plugins/wasm-cpp/extensions/key_auth/README_EN.md
Fkbqf: > What is the subject of "returns 401"?

sry, corrected

PR #286, opened 2023-04-10
feat: support full path regex annotation

### LINKED ISSUE #264: Implement full path regex annotation
## Why do you need it?
Now we implent the `nginx.ingress.kubernetes.io/use-regex` annotation to compatible with nginx ingress, however this is a partial path regex:

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: test-ingress-3
  annotations:
    nginx.ingress.kubernetes.io/use-regex: "true"
spec:
  ingressClassName: nginx
  rules:
  - host: test.com
    http:
      paths:
      - path: /foo/bar/[A-Z0-9]{3}
        pathType: ImplementationSpecific
        backend:
          service:
            name: test
            port: 
              number: 80
```
The ingress controller would define the following location blocks (in this order) within the NGINX template for the test.com server:

```bash
location ~* "^/foo/bar/[A-Z0-9]{3}" {
  ...
}
```
This means that the regex for the path `/foo/bar/[A-Z0-9]{3}` is actually `^/foo/bar/[A-Z0-9]{3}.*`, which confuses users in some scenarios.

## How could it be?

Implement the higress full path regex annotation: 

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: test-ingress-3
  annotations:
    higress/full-path-regex: "true"
spec:
  ingressClassName: nginx
  rules:
  - host: test.com
    http:
      paths:
      - path: /foo/bar/[A-Z0-9]{3}
        pathType: ImplementationSpecific
        backend:
          service:
            name: test
            port: 
              number: 80
```
The regex for the path `/foo/bar/[A-Z0-9]{3}` is actually `^/foo/bar/[A-Z0-9]{3}$`

Charlie17Li:
support full path regex annotation

fixes #264

### DIFF
--- a/pkg/ingress/kube/annotations/annotations.go
+++ b/pkg/ingress/kube/annotations/annotations.go
@@ -76,14 +76,14 @@ func (i *Ingress) NeedRegexMatch() bool {
 		return false
 	}
 
-	return i.Rewrite.RewriteTarget != "" || i.IsPrefixRegexMatch() || i.IsStrictRegexMatch()
+	return i.Rewrite.RewriteTarget != "" || i.IsPrefixRegexMatch() || i.IsFullPathRegexMatch()
 }
 
 func (i *Ingress) IsPrefixRegexMatch() bool {
 	return i.Rewrite.UseRegex
 }
 
-func (i *Ingress) IsStrictRegexMatch() bool {
+func (i *Ingress) IsFullPathRegexMatch() bool {
 	return i.Rewrite.FullPathRegex
 }
 
--- a/pkg/ingress/kube/annotations/annotations.go
+++ b/pkg/ingress/kube/annotations/annotations.go
@@ -76,7 +76,15 @@ func (i *Ingress) NeedRegexMatch() bool {
 		return false
 	}
 
-	return i.Rewrite.RewriteTarget != "" || i.Rewrite.UseRegex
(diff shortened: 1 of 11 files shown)


johnlanni:
@Charlie17Li  Hi, the e2e test cases seems failed.

Charlie17Li:
sorry, I'll fix it soon

pkg/ingress/kube/annotations/annotations.go
```diff
@@ -76,7 +76,15 @@ func (i *Ingress) NeedRegexMatch() bool {
...
 
-	return i.Rewrite.RewriteTarget != "" || i.Rewrite.UseRegex
+	return i.Rewrite.RewriteTarget != "" || i.IsPrefixRegexMatch() || i.IsStrictRegexMatch()
+}
+
+func (i *Ingress) IsPrefixRegexMatch() bool {
+	return i.Rewrite.UseRegex
+}
+
+func (i *Ingress) IsStrictRegexMatch() bool {
```
johnlanni, line 86: StrictRegex -> FullPathRegex

Unified name FullPathRegex will be easier to understand

johnlanni (changes requested):

pkg/ingress/kube/annotations/annotations.go
Charlie17Li: ok

johnlanni (approved):
LGTM

PR #308, opened 2023-04-28
feat: Use new wasm-go-builder with oras in the Makefile

CH3CHO:

Use new wasm-go-builder with oras in the Makefile


johnlanni (approved):
LGTM

PR #313, opened 2023-05-06
Add environment variable goproxy

Fkbqf:

Add environment variable goproxy  for wasm-go/Dockerfile

### DIFF
--- a/plugins/wasm-go/Dockerfile
+++ b/plugins/wasm-go/Dockerfile
@@ -1,6 +1,8 @@
 ARG BUILDER=higress-registry.cn-hangzhou.cr.aliyuncs.com/plugins/wasm-go-builder:go1.19-tinygo0.27.0-oras1.0.0
 FROM $BUILDER as builder
 
+ENV GOPROXY https://goproxy.cn
+
 ARG PLUGIN_NAME=hello-world
 
 WORKDIR /workspace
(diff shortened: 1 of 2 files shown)


plugins/wasm-go/Dockerfile
```diff
@@ -1,6 +1,8 @@
 ARG BUILDER=higress-registry.cn-hangzhou.cr.aliyuncs.com/plugins/wasm-go-builder:go1.19-tinygo0.27.0-oras1.0.0
 FROM $BUILDER as builder
 
+ENV GOPROXY https://goproxy.cn
```
CH3CHO, line 4: This may not be a good idea for developers outside the Chinese mainland.

CH3CHO: Check out this: https://github.com/alibaba/higress/blob/main/plugins/wasm-go/Makefile

Allowing executors to configure their own value might be better.

Fkbqf: > Check out this: https://github.com/alibaba/higress/blob/main/plugins/wasm-go/Makefile
> 
> Allowing executors to configure their own value might be better.

But for domestic developers, if you don’t add this option, it’s not very friendly, I didn’t think carefully, you’re right

johnlanni:
Maybe we should remove `go mod tidy` in Dockerfile, then GOPROXY all depends on local workspace.

CH3CHO:
> Maybe we should remove `go mod tidy` in Dockerfile, then GOPROXY all depends on local workspace.

Agree. And we can execute `go mod tidy` locally instead of inside the builder container.

johnlanni (approved):
LGTM

PR #323, opened 2023-05-10
ci: enable building files caching

Ink-33:
Cache go dependencies and compile caches, submodules and prebuild artifacts. The workflow can be completed within **ten minutes** after the cache is ready. 

no

no

When the cache is ready, a significant reduction in workflow runtime can be observed.

Two tricks are used here to accomplish expectations:
1. `.git/modules` is cached to ensure that `git submodule update -init` runs correctly.
2. `git stash` will be run after the cache is pulled to avoid patches files being overwritten by the cache.

These two points require additional testing to avoid unexpected errors. I will reply here with the test results after a brief test.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -23,6 +23,28 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
+    - name: Setup Golang Caches
+      uses: actions/cache@v3
+      with:
+        path: |-
+          ~/.cache/go-build
+          ~/go/pkg/mod
+        key: ${{ runner.os }}-go-${{ github.run_id }}
+        restore-keys: ${{ runner.os }}-go
+
+    - name: Setup Submodule Caches
+      uses: actions/cache@v3
+      with:
+        path: |-
+            envoy
+            istio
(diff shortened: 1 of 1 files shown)


Ink-33:
updated upstream

.github/workflows/build-and-test.yaml
```diff
@@ -67,14 +111,39 @@ jobs:
...
+            istio
+            external
+            .git/modules
+        key: ${{ runner.os }}-submodules-${{ github.run_id }}
+        restore-keys: ${{ runner.os }}-submodules
+          
+    - run: git stash # restore patch
+
     - name: "Run Ingress Conformance Tests"
       run: GOPROXY="https://proxy.golang.org,direct" make ingress-conformance-test
```
Xunzhuo, line 145: Can we add this into ingress-conformance-test ?

Xunzhuo:
I want to see the difference before and after enabling cache. It takes about 22 mins before caching. Let me rerun build and test workflow and see how long it takes now.

Xunzhuo (approved):

Xunzhuo:
22 min -> 15 min : )

PR #333, opened 2023-05-18
feat: Refactor skywalking configuration structure

CH3CHO:

1. Combine all configuration properties into a single object
2. Only enable skywalking feature when enabled is true and service.address is not empty.


johnlanni (approved):
LGTM

PR #386, opened 2023-06-18
feat: Pilot debug APIs support remote anonymous requests

CH3CHO:

Make pilot debug APIs support remote anonymous requests, so we don't need to generate an access token on purpose  for Higress Console in standalone deployment.


johnlanni (approved):
LGTM

PR #395, opened 2023-06-25
fix: split `REGISTRY` into `BUILDER_REGISTRY` and `REGISTRY`

ESWZY:

As the documentation says, we can 

> The registry address of the generated image, e.g. `example.registry.io/my-name/`. Note that the REGISTRY value should end with /.

But in the Makefile, `REGISTRY` specifies both the registry of builder and the registry of artifact.

For most users, they may not want to set the builder registry and output image registry to be the same, so it is necessary to separate them.


johnlanni (approved):
Thanks

PR #440, opened 2023-07-17
fix go mod path in waf plugin

spacewander:

Use higress's SDK, not its fork.

Only fix the import path

The CI should pass


johnlanni (approved):
Thanks!

PR #462, opened 2023-08-03
update wasm go sdk to tinygo 0.28.1

rinfx:
1. replace tinygogc to avoid memory leak in tinygo 0.28.1
2. update waf plugin to adapt tinygo 0.28.1


johnlanni (approved):
LGTM

Issue #383, opened 2023-06-13
Support Eureka service registry


CH3CHO:
## Why do you need it?

Eureka is a well-known service registry product. As Higress has supported loading services from Nacos and Zookeeper, we should support Eureka as well.

## How could it be?

Use Eureka SDK. You can refer to https://github.com/alibaba/higress/tree/main/registry/nacos

Charlie17Li:
I'm interested in this. Let me try it, thx:)

PR #464, opened 2023-08-07
feat: support eureka registry

Charlie17Li:
support eureka service registry

fixes: #383

### DIFF
--- a/registry/eureka/watcher.go
+++ b/registry/eureka/watcher.go
@@ -214,7 +214,10 @@ func (w *watcher) subscribe(service *fargo.Application) error {
 		defer w.UpdateService()
 
 		if len(service.Instances) != 0 {
-			se := generateServiceEntry(service)
+			se, err := generateServiceEntry(service)
+			if err != nil {
+				return err
+			}
 			w.cache.UpdateServiceEntryWrapper(makeHost(service.Name), &memory.ServiceEntryWrapper{
 				ServiceName:  service.Name,
 				ServiceEntry: se,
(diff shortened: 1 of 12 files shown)


johnlanni:
@Charlie17Li Please add copyright header to the new go files.

johnlanni:
@Charlie17Li Please fix the e2e

Charlie17Li:
> @Charlie17Li Please fix the e2e

ok

registry/eureka/watcher.go
```diff
@@ -0,0 +1,296 @@
...
+
+	return result
+}
+
+func generateServiceEntry(app *fargo.Application) *v1alpha3.ServiceEntry {
+	portList := make([]*v1alpha3.Port, 0)
+	endpoints := make([]*v1alpha3.WorkloadEntry, 0)
+
+	for _, instance := range app.Instances {
+		protocol := common.HTTP
```
johnlanni, line 267: We should support other protocols, such as gRPC, which you can parse from the service's metadata if present.

johnlanni (changes requested):

registry/eureka/watcher.go
Charlie17Li: maybe we can get protocol type from `instance.Metadata`

```
if val, _ := instance.Metadata.GetString("protocol"); val != "" {
			protocol = common.ParseProtocol(val)
		}
```

johnlanni: That's ok

johnlanni: You should check the protocol is valid, if not, should use the http as default.

johnlanni (approved):
Great! LGTM

PR #482, opened 2023-08-11
A plugin that implements token parsing and authentication function based on wasm-go

cf1998:
A plugin that implements token parsing and authentication function based on wasm-go


johnlanni:
能否参考这个：https://github.com/alibaba/higress/tree/main/plugins/wasm-go#e2e%E6%B5%8B%E8%AF%95

写一个e2e测试用例呢

PR #486, opened 2023-08-14
fix: Always use the latest CRDs file when deploying helm chart to OSS

CH3CHO:

Always use the latest CRDs file when deploying helm chart to OSS.

To facilitate local testing and deployment, I keep the `helm/core/crds/customresourcedefinitions.gen.yaml` in place and update it to the latest content.


johnlanni (approved):
LGTM

PR #517, opened 2023-09-11
feat: add request block regular matching and modify block_ Urls are e…

### LINKED ISSUE #512: Request Block URL cannot match properly
## Why do you need it?
 Is your feature request related to a problem? Please describe in details
1. URL cannot achieve regular matching

2. URL matching cannot be achieved. For example, my expectation is to only block/env, but I do not want to block/env/info

yingjianjian:
Modify the default matching mode in request block

Previously, as long as the URL contained characters from block urls, it was considered a successful match. Logically, this should not be the case. Currently, block urls have been changed to precise matching, and block regexp urls have been added to match regular urls

fixes #512


johnlanni (changes requested):

johnlanni:
Please update the E2E test：
https://github.com/alibaba/higress/blob/main/test/e2e/conformance/tests/request-block.yaml
https://github.com/alibaba/higress/blob/main/test/e2e/conformance/tests/request-block.go

plugins/wasm-go/extensions/request-block/main.go
```diff
@@ -104,7 +119,13 @@ func onHttpRequestHeaders(ctx wrapper.HttpContext, config RequestBlockConfig, lo
 			requestUrl = strings.ToLower(requestUrl)
 		}
 		for _, blockUrl := range config.blockUrls {
-			if strings.Contains(requestUrl, blockUrl) {
+			if requestUrl == blockUrl {
```
johnlanni, line 122: Why change this? This change will make this plugin incompatible with previous version.

johnlanni (changes requested):

plugins/wasm-go/extensions/request-block/main.go
yingjianjian: This may require discussing whether using regular matching is more reasonable in the form of substring matching. I feel that retaining two matching modes (precise matching and regular matching) is more reasonable

johnlanni: I think we should talk about the new configuration format while maintaining configuration compatibility. I suggest you design an example of the new configuration format, and we will discuss it based on this.

johnlanni (changes requested):

plugins/wasm-go/extensions/request-block/main.go
yingjianjian: Reserve block-url for string inclusion matching, add new blocks_ Exact_ URLs for precise URL matching

johnlanni (approved):
LGTM Thanks！

johnlanni:
@yingjianjian Hi, please resolve the conflicts.

PR #523, opened 2023-09-14
feat: Add update function to get-higress.sh

CH3CHO:

1. Add "-u" flag to get-ingress.sh to enable update mode.
1. The current deployment is packed before starting the update.
1. Only change image tags during update.
1. User can apply updates when Higress is running, but need to restart higress to use the latest version.

image

Sibling PR: https://github.com/higress-group/higress-standalone/pull/37

### DIFF
--- a/tools/get-higress.sh
+++ b/tools/get-higress.sh
@@ -16,6 +16,8 @@
 
 export VERSION
 
+MODE="install"
+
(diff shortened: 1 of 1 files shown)


johnlanni (approved):
LGTM

PR #543, opened 2023-09-19
feat: Use istio to provide Gateway API support

CH3CHO:

Use istio to provide Gateway API support.

For newly added envs:
- PILOT_ENABLE_GATEWAY_API: If this is set to true, support for Kubernetes gateway-api (github.com/kubernetes-sigs/gateway-api) will be enabled. In addition to this being enabled, the gateway-api CRDs need to be installed. Default to true in istio.
- PILOT_ENABLE_GATEWAY_API_STATUS: If this is set to true, gateway-api resources will have status written to them. Default to true in istio.
- PILOT_ENABLE_GATEWAY_API_DEPLOYMENT_CONTROLLER: If this is set to true, gateway-api resources will automatically provision in cluster deployment, services, etc. Default to false in istio.

Gateway API resources for testing:
```yaml
apiVersion: gateway.networking.k8s.io/v1alpha2
kind: Gateway
metadata:
  name: higress-gateway
  namespace: higress-system
spec:
  gatewayClassName: istio
  listeners:
  - name: default
    hostname: "*.test.com"
    port: 80
    protocol: HTTP
    allowedRoutes:
      namespaces:
        from: All
---
apiVersion: gateway.networking.k8s.io/v1alpha2
kind: HTTPRoute
metadata:
  name: http
  namespace: default
spec:
  parentRefs:
  - name: higress-gateway
    namespace: higress-system
  hostnames: ["www.test.com"]
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /
    filters:
    - type: RequestHeaderModifier
      requestHeaderModifier:
        add:
        - name: my-added-header
          value: added-value-higress
    backendRefs:
    - name: home-service # This needs to be changed to a valid local K8s service reference
      port: 8080
```

image


Xunzhuo:
When set PILOT_ENABLE_GATEWAY_API to true, will automatically install GWAPI related Resources(CRDs)?

CH3CHO:
> When set PILOT_ENABLE_GATEWAY_API to true, will automatically install GWAPI related Resources(CRDs)?

So far, no. Should we?

Xunzhuo:
So when you test this feature, you manually install gwapi resources?

Xunzhuo:
So it will confuse people if they install the latest version of gwapi.

Two options:
+ Add docs to tell which latest gwapi version higress(binded istio) supports.
+ Include GWAPI resources into Helm Chart, and install them when enabling gwapi features.

WDYT @CH3CHO @johnlanni

johnlanni:
@Xunzhuo I agree, we will add instructions like the "Supporting Istio CRD" section [here](https://higress.io/en-us/docs/ops/deploy-by-helm/).

johnlanni (approved):

PR #562, opened 2023-09-25
fix: hide auto-generated completion command when run hgctl --help

2456868764:
when run hgctl --help, it will show : 
```
Available Commands:
  completion     generate the autocompletion script for the specified shell

```
i have tested completion command which  can not work well , so hide it first.


Xunzhuo:
Nope, it can work.

PR #620, opened 2023-11-02
Add detecting higress installed by helm or not before install

2456868764:
Add detecting higress installed by helm or not before install


pkg/cmd/hgctl/installer/installer_k8s.go
```diff
@@ -35,6 +35,14 @@ type K8sInstaller struct {
 }
 
 func (o *K8sInstaller) Install() error {
+	// check if higress is installed by helm
```
johnlanni, line 38: It would be better for us to first resolve the namespace from the configuration file and then check if higress is installed in the specific namespace.

johnlanni (approved):

PR #665, opened 2023-12-05
opt(hgctl/dashboard): avoid printing error messages cannot open browser

### LINKED ISSUE #652: `hgctl dashboard` command fails on Linux without xdg-open command installed

- [x] I have searched the [issues](https://github.com/alibaba/higress/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description

`hgctl dashboard` command fails without xdg-open command installed.

### Ⅱ. Describe what happened

8b17c306605c44b608a3878e6189fcee

There might be three options:
1. Show a friendly error message.
2. Open browser with something else.
3. Support port-forward only mode without opening browser directly, especially when running the command in an SSH session.

See Ⅱ.

### Ⅵ. Environment:

- Higress version:  v1.3.1
- OS : Alibaba Cloud Linux 3.2104 LTS 64bits
- Others:

sjcsjc123:

Avoid printing error messages when commands do not exist.

fixes #652 

This is related to the presence or absence of a GUI.

This is executed in the centos container of Docker.

image

This is executed in local.

image

### DIFF
--- a/pkg/cmd/hgctl/dashboard.go
+++ b/pkg/cmd/hgctl/dashboard.go
@@ -361,7 +361,6 @@ func ClosePortForwarderOnInterrupt(fw kubernetes.PortForwarder) {
 }
 
 func openBrowser(url string, writer io.Writer, browser bool) {
-
 	fmt.Fprintf(writer, "%s\n", url)
 
 	if !browser {
@@ -379,7 +378,6 @@ func openBrowser(url string, writer io.Writer, browser bool) {
 	default:
 		fmt.Fprintf(writer, "Unsupported platform %q; open %s in your browser.\n", runtime.GOOS, url)
 	}
-
 }
 
 func openCommand(writer io.Writer, command string, args ...string) {
--- a/pkg/cmd/hgctl/dashboard.go
+++ b/pkg/cmd/hgctl/dashboard.go
@@ -390,6 +390,7 @@ func openCommand(writer io.Writer, command string, args ...string) {
 			return
 		}
 		fmt.Fprintf(writer, "Failed to open browser; open %s in your browser.\nError: %s\n", args[0], err.Error())
+		return
 	}
 
 	err = exec.Command(command, args...).Start()
--- a/pkg/cmd/hgctl/dashboard.go
+++ b/pkg/cmd/hgctl/dashboard.go
@@ -372,11 +372,11 @@ func openBrowser(url string, writer io.Writer, browser bool) {
 
 	switch runtime.GOOS {
 	case "linux":
-		err = openCommand("xdg-open", url)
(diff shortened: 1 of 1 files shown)


CH3CHO:
Why are there only a plain URL showing in the terminal?

sjcsjc123:
Is it okay to add and print this information?
image

CH3CHO:
> Is it okay to add and print this information? image

Could you post a screenshot of the final output? Thanks.

sjcsjc123:
Of course.
image

pkg/cmd/hgctl/dashboard.go
```diff
@@ -385,3 +385,13 @@ func openBrowser(url string, writer io.Writer, browser bool) {
 		fmt.Fprintf(writer, "Failed to open browser; open %s in your browser.\nError: %s\n", url, err.Error())
 	}
 }
+
+func openCommand(command string, args ...string) error {
+	_, err := exec.LookPath(command)
+	if err != nil {
+		return nil
```
CH3CHO, line 392: Why don't we return `err` but a `nil` here?

Xunzhuo:
We already have a flag for it: `--browser=false`

pkg/cmd/hgctl/dashboard.go
sjcsjc123: When the command cannot be found, it will return err. At this time, simply print a reminder and return nil. If it returns err, there will be a step to print err later.

sjcsjc123:
> We already have a flag for it: `--browser=false`

By default, browser is set to true, and users may not notice this flag. Running the default command without a GUI interface will also print the error. This is just my suggestion.

pkg/cmd/hgctl/dashboard.go
CH3CHO: I think if a function cannot perform the task it claims to perform, an error should be returned. And if there is no error returned, how should we determine whether to show the corresponding message then?

sjcsjc123: image

How about this? If the command cannot be found, it will prompt that the browser cannot be opened; Otherwise, return err.

CH3CHO: LGTM

CH3CHO: Which one is better?
1. Display an error message in openCommand and return nil
2. Return an error in openCommand function and show the corresponding error message in the caller

sjcsjc123: I tend to print it out in the opencommand method and then change it to a method with no return parameters.

Xunzhuo (approved):
Current approach looks good, this is an opt on error msg but not a bug fix.

Issue #676, opened 2023-12-08
Unify image customization methods in the core Helm chart


CH3CHO:
**If you are reporting *any* crash or *any* potential security issue, *do not*
open an issue in this repo. Please report the issue via [ASRC](https://security.alibaba.com/)(Alibaba Security Response Center) where the issue will be triaged appropriately.**

- [X] I have searched the [issues](https://github.com/alibaba/higress/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description

There are two pods with three containers in the Higrese Core Helm chart. The containers are controller, pilot and gateway. And the ways used to customize their images are quite different.

1. Controller ([Link to source code](https://github.com/alibaba/higress/blob/main/helm/core/templates/controller-deployment.yaml#L187))
    ```yaml
    ["{{ .Values.hub }}/{{ .Values.controller.image }}:{{ .Values.controller.tag | default .Chart.AppVersion }}"](image: "{{ .Values.hub }}/{{ .Values.controller.image }}:{{ .Values.controller.tag | default .Chart.AppVersion }}")
    ```
2. Pilot ([Link to source code](https://github.com/alibaba/higress/blob/main/helm/core/templates/controller-deployment.yaml#L34C1-L38C11))
    ```yaml
    {{- if contains "/" .Values.pilot.image }}
          image: "{{ .Values.pilot.image }}"
    {{- else }}
          image: "{{ .Values.pilot.hub | default .Values.global.hub }}/{{ .Values.pilot.image | default "pilot" }}:{{ .Values.pilot.tag | default .Chart.AppVersion }}"
    {{- end }}
    ```
3. Gateway ([Link to source code](https://github.com/alibaba/higress/blob/main/helm/core/templates/deployment.yaml#L71))
    ```yaml
    image: "{{ .Values.hub }}/{{ .Values.gateway.image }}:{{ .Values.gateway.tag | default .Chart.AppVersion }}"
    ```

### Ⅱ. Describe what you expected to happen

It would be better to have a unified way to customize component images, for a better debug or test experience.

Uncle-Justice:
Is the difference now caused by the different ways of setting in `values.yaml`? 

If so, i think we can first unify the formats of coressponding fields like `hub` and `image` for these three containers in `values.yaml`, then we can easily unify this issue.

For example: 

```yaml
[...]

PR #687, opened 2023-12-12
refactor: unify image customization methods

Uncle-Justice:

#676 

i made wasm-plugin test to check if the PR works. It showed no errors.

i followed the settings in `values.yaml`, the default image for controller is "higress" in customization. 

I mention this because the other two containers keep their name and image consistent. pilot's default image is "pilot". gateway's image is "gateway". The naming methods of images may be a little bit worth concerning.


johnlanni (approved):
LGTM

Issue #668, opened 2023-12-06
Hgctl has no prebuilt binary for windows


johnlanni:
## Why do you need it?

When we execute this command on windows:

`curl -Ls https://raw.githubusercontent.com/alibaba/higress/main/tools/hack/get-hgctl.sh | VERSION=latest bash`

This error occurs:

![image](https://github.com/alibaba/higress/assets/6763318/9c2a1e47-2d31-4336-9fe9-19c0bd36cdee)

## Other related information

1. action：https://github.com/alibaba/higress/blob/main/.github/workflows/latest-release.yaml
2. makefile target：
https://github.com/alibaba/higress/blob/a554ee1cebdb60bb166d3abcbc287bd68e940d23/Makefile.core.mk#L90-L94
3. install script：
https://github.com/alibaba/higress/blob/main/tools/get-higress.sh

johnlanni:
@Ink-33 Please add a comment here and I can assign this issue to you.

Ink-33:
please assign me, thanks

PR #739, opened 2023-12-28
tools(hgctl): add windows build

Ink-33:
Add windows build for hgctl.

fixs #668  

Run `make build-hgctl-multiarch` and can see prebuilt binnary for windows/amd64 and windows/arm64

Some signal like `syscall.SIGUSR1` is removed for windows build by using conditional compilation directive because windows doesn't have those signal.

Meanwhile, the distributions of hgctl for windows use zip to archive for convenience, but this cause that `get-hgctl.sh` needs to be modified. In my opinion, we can simply add a note in documents for windows user to download `hgctl` manually, ~~and keep the download script unchanged.~~ 

Now windows user can also use `get-hgctl.sh`  to install `hgctl`. The default install loction is `%USERPROFILE%/hgctl/bin`.


Ink-33:
Please help me review~
Since there is no windows prebuilt binary uploaded, the script cannot use now except mocking the download url😂

johnlanni (approved):
LGTM

PR #880, opened 2024-03-25
fix: Consider the new ingress package is available as default

CH3CHO:

Consider the new ingress package is available as default, just in case anything goes wrong during initialization, then Higress falls into v1beta1/Ingress and can't recover before restarting.


johnlanni (approved):
LGTM

PR #920, opened 2024-04-23
feat: Write the original host header before changed by Wasm plugin into access log

CH3CHO:

1. Allow Wasm plugins to use `X-ENVOY-ORIGINAL-HOST` header to store the host header by update.
2. Use the same access log settings in the `envoy.yaml` file generated by `hgctl plugin test create` command.


johnlanni (approved):
LGTM

PR #976, opened 2024-05-19
feat: Enhance the feature of ai-proxy plugin

CH3CHO:

1. Support using the incremental streaming feature of Qwen itself.
2. Skip stream response body processing if the response isn't a text/event-stream content.
3. Support tools call in Qwen.
4. Add some debug logs.
5. Support toggling response streaming with "stream" field in request body when using Qwen.
54fea8f9687017c44a5a044f1959fa03
6. Disable route re-calculation in onHttpRequestHeaders.
7. Fix the incorrect message sequence when enabling chat context.
1. Support enabling search function provided by Qwen.
1. Support configuring fileIds as context in Qwen.
1. Sync README file changes.


johnlanni (approved):
LGTM, Thanks!

PR #991, opened 2024-05-23
fix: Fix the incorrect usage of DisableReroute

CH3CHO:

Move the invocation of ctx.DisableReroute function from onHttpResponseHeaders to onHttpRequestHeaders.


johnlanni (approved):
LGTM

PR #1009, opened 2024-05-28
fix: Fix tool_calls compatibility issues with LobeChat

CH3CHO:

1. Add index field to toolCall.
2. Always return id and name fields in functionCall even if they are empty.
3. Make sure a response message doesn't contain non-empty content and tool_calls at the same time.
4. Return token usages in the last stream chunk.


johnlanni (approved):
LGTM

PR #1045, opened 2024-06-18
fix: Accommodate the incomplete function name in the initial event from Qwen

CH3CHO:

Sometimes in the initial push event from Qwen with a tool call, the function name isn't complete. If we push the event directly to the client, it may cause the tool call failed to be executed. So we wait until the arguments are coming, which means the function name is complete, and push the event then.


johnlanni (approved):
LGTM

johnlanni (approved):
LGTM

PR #1067, opened 2024-06-28
feat: Load 3rd-party images from higress image repo

CH3CHO:

Load 3rd-party images from higress image repo.

### DIFF
--- a/helm/core/values.yaml
+++ b/helm/core/values.yaml
@@ -343,7 +343,7 @@ global:
     enabled: false
     promtail:
       image:
-        repository: grafana/promtail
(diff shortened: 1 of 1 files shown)


johnlanni:
This should require support for multiple architectures. Shall we address this by creating a manifest?

CH3CHO:
> This should require support for multiple architectures. Shall we solve this throught docker manifest?

I'll see if we can do it.

CH3CHO:
> > This should require support for multiple architectures. Shall we solve this throught docker manifest?
> 
> I'll see if we can do it.

I've synced all image arcitectures found in Docker Hub to the Higress image repo.

PR #1097, opened 2024-07-06
feat: Support model prefix mapping in ai-proxy

CH3CHO:

1. Support model prefix mapping in ai-proxy.
2. If the mapped model name is an empty string, use the original model name instead.


johnlanni (approved):
LGTM

PR #1116, opened 2024-07-11
feat: Add statusCodeDetails info when returning response in Wasm plugins directly

CH3CHO:

Add statusCodeDetails info when returning response in Wasm plugins directly.


johnlanni (approved):
LGTM

PR #1121, opened 2024-07-15
feat: add annotation for mirror svc

robynron:
add new ingress annotation feature.
`higress.io/mirror-target-service`: Route traffic that meets all routing rules under the ingress to the specified service.
`higress.io/mirror-percentage`: Route traffic that meets all routing rules under the ingress to the specified services by percentage.


SpecialYang (changes requested):
看看加下e2e test

Xunzhuo (commented):
A question for this feature, how do we specify the mirror filter to a specific rule? Or does it mean all rules backend will have a same mirror svc?

johnlanni:
@robynron 解决一下冲突吧

robynron:
> A question for this feature, how do we specify the mirror filter to a specific rule? Or does it mean all rules backend will have a same mirror svc?

mirror all rules under a ingress to a new service.

johnlanni (approved):
LGTM

PR #1144, opened 2024-07-22
feat: Make higress-core and higress-gateway as the default container

CH3CHO:

Make higress-core and higress-gateway as the default container, so it would be easier for users to run `kubectl logs` and `kubectl exec` commands.

image

image

Q: Why don't I use the `kubectl.kubernetes.io/default-container` mentioned here (https://github.com/kubernetes/enhancements/blob/master/keps/sig-cli/2227-kubectl-default-container/README.md)?

A: Although we get almost the same result after adding this annotation, the first line output, `Defaulted container "higress-core" out of: higress-core, discovery` will disappear, making users more difficult to know all the container names.

### DIFF
--- a/helm/core/templates/controller-deployment.yaml
+++ b/helm/core/templates/controller-deployment.yaml
@@ -32,6 +32,64 @@ spec:
       securityContext:
         {{- toYaml .Values.controller.podSecurityContext | nindent 8 }}
       containers:
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.controller.securityContext | nindent 12 }}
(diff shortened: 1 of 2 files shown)


fengxsong:
I thing annotations can do that thick, ref: https://kubernetes.io/docs/reference/labels-annotations-taints/#kubectl-kubernetes-io-default-container

CH3CHO:
> I thing annotations can do that thick, ref: https://kubernetes.io/docs/reference/labels-annotations-taints/#kubectl-kubernetes-io-default-container

I'm aware of this. Please check out my explanation in "Ⅴ. Special notes for reviews" and let me know what do you think. Thanks.

cr7258 (approved):
LGTM🐉

PR #1155, opened 2024-07-25
feat: Always buffer request body in ai-proxy plugin

CH3CHO:

1. Always buffer request body in ai-proxy plugin, so it can be used by fallback routing.
2. Always include usage data in the response when sending streaming requests to OpenAI.


johnlanni (approved):
LGTM

PR #1161, opened 2024-07-29
fix: Use the official tinygo package to build Wasm go plugin builder

CH3CHO:

Use the official tinygo package to build Wasm go plugin builder, since the current tinygo version we used, 0.28.1, no longer needs our customization.


johnlanni (approved):
LGTM

PR #1189, opened 2024-08-08
fix bug of empty config plugin still start

Jing-ze:
fix bug of empty config plugin still start


johnlanni (approved):
LGTM

PR #1253, opened 2024-08-26
fix: Use shallow mode when cloning submodules

CH3CHO:

Use shallow mode when cloning submodules to avoid the following error:

```
error: rpc failed; curl 18 transfer closed with outstanding read data remaining
```


johnlanni (approved):
LGTM

PR #1332, opened 2024-09-23
add istio workload sds

johnlanni:

Although the workload secret is not utilized, including it prevents the istio agent from logging error messages.


CH3CHO (approved):
LGTM

PR #1361, opened 2024-10-02
feat: sse-timing plugin for tracing events

jizhuozhi:

Add sse-timing plugin for tracing timestamp that each event received in higress. An alternative implementation of the server-timing protocol (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing). It's useful to profile AI backend latency which uses event-stream MIME type.

And implemention an EventStream for common SSE(text/event-stream) parsing.

None

I'm sorry I don't know how to add WASM testcase, we could discuss in this PR.

There is a docker-compose.yml in plugin directory, with a go server backend which provides sse response, up it and then just use broswer to open http://localhost:10000, you could see events in dev-tools.


johnlanni:
cc @007gzs

007gzs:
Please add e2e test cases. Reference https://github.com/alibaba/higress/tree/main/test/e2e/conformance/tests

jizhuozhi:
Will, I have seen the test-suits in https://github.com/alibaba/higress/tree/main/test/e2e/conformance/tests, it seems just supporting static response assertion as server-timing is streaming injecting dynamic metainfo into response body, and no sse server as upstream (I just found echo-server and echo-body in deployments). Does any solution or example here?

plugins/wasm-rust/extensions/sse-timing/src/lib.rs
```diff
@@ -0,0 +1,195 @@
...
+        }
+        HeaderAction::Continue
+    }
+
+    fn on_http_response_headers(
+        &mut self,
+        _num_headers: usize,
+        _end_of_stream: bool,
+    ) -> HeaderAction {
+        match self.get_http_response_header("Content-Type") {
```
johnlanni, line 118: The E2E test failed because it modifies the body; hence, it is necessary to remove the 'Content-Length' header during the header processing stage.

jizhuozhi: According to text/event-stream MIME type, the origin server should not set Content-Length in response headers, I will change custom-response to support streaming response

johnlanni: Using custom response to simulate responses will definitely add the content-length header. In real scenarios, when using the text/event-stream MIME type, it's not necessarily chunked transfer either; for example, if a proxy does caching in between, the content-length header might also be added. Therefore, the logic should still include compatibility handling and uniformly remove the content-length header.

plugins/wasm-go/extensions/custom-response/main.go
```diff
@@ -71,7 +72,10 @@ func parseConfig(gjson gjson.Result, config *CustomResponseConfig, log wrapper.L
 			config.contentType = "text/plain; charset=utf-8"
 		}
 	}
-	config.headers = append(config.headers, [2]string{"content-type", config.contentType})
+	config.isStream = gjson.Get("is_stream").Bool()
+	if !config.isStream {
+		config.headers = append(config.headers, [2]string{"content-type", config.contentType})
```
johnlanni, line 77: content-type?

plugins/wasm-go/extensions/custom-response/README_EN.md
```diff
@@ -12,12 +12,13 @@ Plugin Execution Phase: `Authentication Phase`
...
+|------------------|-----------------|--------------|---------------|---------------------------------------------------------------------------------------------------------------------|
+| status_code      | number          | Optional     | 200           | Custom HTTP response status code                                                                                    |
+| headers          | array of string | Optional     | -             | Custom HTTP response headers, keys and values separated by `=`                                                      |
+| body             | string          | Optional     | -             | Custom HTTP response body                                                                                           |
+| enable_on_status | array of number | Optional     | -             | Match original status codes to generate custom responses; if not specified, the original status code is not checked |
+| is_stream        | bool            | Optional     | false         | If stream response, will not set Content-Length header                                                              |
```
johnlanni, line 21: In fact, it is not possible to control whether the content-length header is set, because when the plugin calls sendHTTPResponse, it will callback to the Envoy logic, which automatically adds this header

johnlanni (changes requested):
As discussed above, the plugin needs to be compatible with the potential presence of the content-length response header, but modifying the body can lead to inconsistent lengths. Simply removing the content-length response header during the response header phase is sufficient, and Envoy will automatically switch to chunked transfer mode.

jizhuozhi:
Fine, I'm tired to fix test suits such as custom-response plugin. I will move to another PR https://github.com/alibaba/higress/pull/1416 which makes sse-timing as an example then no need to add test cases.

PR #1377, opened 2024-10-10
feat: implements Promise for dispatch callback

jizhuozhi:

In the current envoy WASM plugins, if we need IO request such as HTTP/GRPC/Redis, we must register request to envoy event loop and assigned a token via such as `dispatch_http_call` and WASM plugin should yield current request lifetime using `Action::Pause`. When the IO request completed, the envoy will callback to WASM plugin via `on_http_call_response`, and plugin should dispatch response using token (or ignored if single IO request). If we want to share something between `dispatch_http_call` and `on_http_call_response`, we must share them in plugin context fields, it's not a suitable scope.

Here is an example from [proxy-wasm-rust-sdk](https://github.com/higress-group/proxy-wasm-rust-sdk/blob/main/examples/http_auth_random/src/lib.rs)
```rust
impl HttpContext for HttpAuthRandom {
    fn on_http_request_headers(&mut self, _: usize, _: bool) -> Action {
        self.dispatch_http_call(
            "httpbin",
            vec![
                (":method", "GET"),
                (":path", "/bytes/1"),
                (":authority", "httpbin.org"),
            ],
            None,
            vec![],
            Duration::from_secs(1),
        )
        .unwrap();
        Action::Pause
    }

    fn on_http_response_headers(&mut self, _: usize, _: bool) -> Action {
        self.set_http_response_header("Powered-By", Some("proxy-wasm"));
        Action::Continue
    }
}

impl Context for HttpAuthRandom {
    fn on_http_call_response(&mut self, _: u32, _: usize, body_size: usize, _: usize) {
        if let Some(body) = self.get_http_call_response_body(0, body_size) {
            if !body.is_empty() && body[0] % 2 == 0 {
                info!("Access granted.");
                self.resume_http_request();
                return;
            }
        }
        info!("Access forbidden.");
        self.send_http_response(
            403,
            vec![("Powered-By", "proxy-wasm")],
            Some(b"Access forbidden.\n"),
        );
    }
}
```

So there is three major problem we need to resolve:
1. How to easier dispatch request and callback response via token.
2. How to share something between dispatch request and callback response with smaller scope.
3. How to make code more fluid instead of spreading logic in different places

In Rust async programming, normally we use `async/await` for IO request, but in envoy WASM plugin, there is no executor to poll future. A suitable solution is providing JavaScript style Promise (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise). With Promise, we could write all logic in single function 
```Rust
impl HttpContext for HttpAuthRandom {
    fn on_http_request_headers(&mut self, _: usize, _: bool) -> Action {
        let token = self.dispatch_http_call(
            "httpbin",
            vec![
                (":method", "GET"),
                (":path", "/bytes/1"),
                (":authority", "httpbin.org"),
            ],
            None,
            vec![],
            Duration::from_secs(1),
        )
        .unwrap();
        let promise = Promise::new();
        // make relation between promise and request (token)
        promise.then(|(_, _, body_size, _)| {
            if let Some(body) = hostcalls::get_http_call_response_body(0, body_size) {
                if !body.is_empty() && body[0] % 2 == 0 {
                    info!("Access granted.");
                    hostcalls::resume_http_request();
                    return;
                }
            }
            info!("Access forbidden.");
            hostcalls::send_http_response(
                403,
                vec![("Powered-By", "proxy-wasm")],
                Some(b"Access forbidden.\n"),
            );
        })
        Action::Pause
    }

    fn on_http_response_headers(&mut self, _: usize, _: bool) -> Action {
        self.set_http_response_header("Powered-By", Some("proxy-wasm"));
        Action::Continue
    }
}
```
It seems more fluid then writing in callback of `on_http_call_response`, but there is no executor for promise to trigger state transferring. We can use `on_http_call_response` as trigger simply
```rust
struct HttpAuthRandom {
    promise: Rc<Promise<(u32, usize, usize, usize)>>
}

impl HttpContext for HttpAuthRandom {
    fn on_http_request_headers(&mut self, _: usize, _: bool) -> Action {
      // ...
      // make relation between promise and request (token)
      self.promise = promise.clone();
      // ...
    }
}

impl Context for HttpAuthRandom {
    fn on_http_call_response(&mut self, _token_id: u32, _num_headers: usize, _body_size: usize, _num_trailers: usize) {
        self.promise.fulfill((_token_id, _num_headers, _body_size, _num_trailers))
    }
}
```

As for making relationship between multi tokens and promises, we could just using `HashMap` with `insert/remove` (maybe it should be embed in SDK but not belongs to this PR)
```rust
struct HttpAuthRandom {
    m: HashMap<u32, Rc<Promise<(u32, usize, usize, usize)>>
}

impl HttpContext for HttpAuthRandom {
    fn on_http_request_headers(&mut self, _: usize, _: bool) -> Action {
      // ...
      // make relation between promise and request (token)
      self.m.insert(token, promise.clone());
      // ...
    }
}

impl Context for HttpAuthRandom {
    fn on_http_call_response(&mut self, _token_id: u32, _num_headers: usize, _body_size: usize, _num_trailers: usize) {
        let promise = self.m.remove(token);
        promise.fulfill((_token_id, _num_headers, _body_size, _num_trailers))
    }
}
```


plugins/wasm-rust/examples/http-dispatch-example/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+        &mut self,
+        _token_id: u32,
+        _num_headers: usize,
+        _body_size: usize,
+        _num_trailers: usize,
+    ) {
+        self.http_dispatcher
+            .callback(_token_id, _num_headers, _body_size, _num_trailers)
+    }
+}
```
johnlanni, line 90: This logic should be placed on the SDK side, so the user's code doesn't need to worry about this detail

plugins/wasm-rust/src/dispatcher.rs
```diff
@@ -0,0 +1,54 @@
...
+
+impl HttpDispatcher {
+    pub fn dispatch(
+        &mut self,
+        upstream: &str,
+        headers: Vec<(&str, &str)>,
+        body: Option<&[u8]>,
+        trailers: Vec<(&str, &str)>,
+        timeout: Duration,
+    ) -> Rc<Promise<(u32, usize, usize, usize)>> {
```
johnlanni, line 34: Based on this, further encapsulation is needed, such as supporting the extraction of domain and path from URL and cluster, as well as supporting get/post, etc

johnlanni (changes requested):
Let's integrate the functionality in this [PR's](https://github.com/alibaba/higress/pull/1367/files#diff-fa0f17c97782fd1be62475c4c6d8cb80b2dab405c35b4a7d9f3a18262de38d20) plugin wrapper further. Hopefully, we can complete the httpcall feature for rust in this PR. For users, it should be very simple to use

johnlanni:
cc @007gzs

PR #1381, opened 2024-10-10
feat: Support route fallback by default

CH3CHO:

Support route fallback by default.

1. Enable Istio CRD by default.
2. Insert custom response filter by default.


johnlanni (approved):
LGTM

PR #1397, opened 2024-10-16
fix: Fix clean targets in Makefile

CH3CHO:

Some folders were not deleted in the clean targets. Add them to the Makefile.


johnlanni (approved):
LGTM

PR #1419, opened 2024-10-22
fix: do not create ingressclass when it's empty

fengxsong:

when there are only one higress-controller watch all the ingress objects, we do not need to create the ingressclass object.


CH3CHO (approved):
LGTM. Thanks.

PR #1441, opened 2024-10-28
feat: lazy formatted log

jizhuozhi:

Current logging in wasm-rust needs to be formatted before level checking, it means even log level is less than configured, but the cost of formatting is wasted.

Provides a serials of `format log` methods, which using `Argument` as args and formatting message after level checking success.

# Ⅱ. Does this pull request fix one issue?

None

None

None

None


plugins/wasm-rust/src/log.rs
```diff
@@ -68,4 +69,45 @@ impl Log {
     pub fn critical(&self, msg: &str) {
         self.log(LogLevel::Critical, msg)
     }
+
```
007gzs, line 72: examples can be added in the comments, such as log.debugf (format_args! ("Hello, {}!", "world");

007gzs (commented):
log和logf有必要合并下么？比如
```
fn log(&self, level: LogLevel, msg: &str) {
    self.logf(level, format_args!("{}", msg));
}
```
现在通用格式化`format!("[{}] {}", self.plugin_name, msg);` 和level转换写在了两个地方，如果后面想改格式的话需要改多处

007gzs (approved):
LGTM

PR #1483, opened 2024-11-06
Implement Rust Wasm Plugin Build & Publish Action

### LINKED ISSUE #1469: Implement Rust Wasm Plugin Build & Publish Action
## Why do you need it?

We have supported the Build & Publish Action for Go Wasm plugins:

https://github.com/alibaba/higress/blob/main/.github/workflows/build-and-push-wasm-plugin-image.yaml

Need to extend this to support the Build & Publish of Rust Wasm plugins

007gzs:

fixes #1469

### DIFF
--- a/.github/workflows/build-and-push-wasm-plugin-image.yaml
+++ b/.github/workflows/build-and-push-wasm-plugin-image.yaml
@@ -35,7 +35,7 @@ jobs:
       TINYGO_VERSION: 0.28.1
       ORAS_VERSION: 1.0.0
     steps:
-      - name: Set plugin_type、 plugin_name and version from inputs or ref_name
+      - name: Set plugin_type, plugin_name and version from inputs or ref_name
         id: set_vars
         run: |
           if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
--- a/.github/workflows/build-and-push-wasm-plugin-image.yaml
+++ b/.github/workflows/build-and-push-wasm-plugin-image.yaml
@@ -3,9 +3,16 @@ name: Build and Push Wasm Plugin Image
 on:
   push:
     tags:
-    - "wasm-go-*-v*.*.*" # 匹配 wasm-go-{pluginName}-vX.Y.Z 格式的标签
+    - "wasm-*-*-v*.*.*" # 匹配 wasm-{go|rust}-{pluginName}-vX.Y.Z 格式的标签
   workflow_dispatch:
     inputs:
+      plugin_type:
+        description: 'Type of the plugin'
+        required: true
+        type: choice
+        options:
+          - go
+          - rust
       plugin_name:
         description: 'Name of the plugin'
         required: true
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-push-wasm-plugin-image.yaml
```diff
@@ -23,32 +30,40 @@ jobs:
     env:
       IMAGE_REGISTRY_SERVICE: ${{ vars.IMAGE_REGISTRY || 'higress-registry.cn-hangzhou.cr.aliyuncs.com' }}
       IMAGE_REPOSITORY: ${{ vars.PLUGIN_IMAGE_REPOSITORY || 'plugins' }}
+      RUST_VERSION: 1.82
       GO_VERSION: 1.19
       TINYGO_VERSION: 0.28.1
       ORAS_VERSION: 1.0.0
     steps:
-      - name: Set plugin_name and version from inputs or ref_name
+      - name: Set plugin_type、 plugin_name and version from inputs or ref_name
```
jizhuozhi, line 38: don't use full width character in english comments

plugins/wasm-rust/DockerfileBuilder
```diff
@@ -0,0 +1,38 @@
+
```
jizhuozhi, line 1: why do we need another Dockerfile to build?

007gzs: I referred to the builder used for packaging and publishing in the go-lang https://github.com/alibaba/higress/blob/main/plugins/wasm-go/DockerfileBuilder

.github/workflows/build-and-push-wasm-plugin-image.yaml
007gzs: done

plugins/wasm-rust/DockerfileBuilder
jizhuozhi: Dockerfile is enough, no need additional files

jizhuozhi: Both Dockerfile and DockerfileBuilder are for building an artifact named plugin.wasm, but the oras is a client which aims to publish artifact to registry. It means no matter which language is, the only different is how to build an artifact but using same way to publish artifact. 

In a word, I think the oras should be a independent image, but not mixed in builders. 

cc @johnlanni

johnlanni: builder mainly facilitates developers by eliminating the need to download corresponding versions of oras, as well as Rust build tools, etc

jizhuozhi: For developers, devcontainer might be the best solution and I still think it should keep simple for OCI building

johnlanni: Well, you're right, developers just need the Dockerfile, but it seems that the DockerfileBuilder used for building images allows the workflow for publishing wasm plugins to remain consistent with Go. From this perspective, it still makes sense

johnlanni (approved):
LGTM

PR #1564, opened 2024-12-03
fix: qwen stream issue

cr7258:

修复 qwen 流式响应的 header 设置问题。

CleanShot 2024-12-03 at 12 03 20@2x


CH3CHO (approved):
LGTM

PR #1621, opened 2024-12-23
fix: Make opa listen to all IPs

CH3CHO:

Make opa listen to all IPs, otherwise, it will listen to localhost by default.

image


johnlanni (approved):
LGTM

PR #1640, opened 2025-01-03
fix: Enlarge the default retry timeout in ai-proxy

CH3CHO:

Enlarge the default retry timeout in ai-proxy since retry requests are not streaming.


johnlanni (approved):
LGTM

PR #1644, opened 2025-01-06
ai-cache update body buffer limit size

johnlanni:

The default buffer size limit is 32k, which can be easily exceeded, leading to a 413 error.


CH3CHO (approved):
LGTM

PR #1645, opened 2025-01-06
fix ai proxy checkStream

johnlanni:

`wrapper.HttpContext` is an interface, there's no need to use a pointer here


CH3CHO (approved):
LGTM

PR #1707, opened 2025-01-23
optimize body bufferlimit set in ext-auth plugin

johnlanni:
1. Unify the setting to prohibit re-routing, which was previously only considered for handling the request body;
2. Avoid setting the request buffer size to 0


hanxiantao:
LGTM

PR #1746, opened 2025-02-08
update ai-cache extension

mirror58229:
1. Change `ctx.DontReadRequestBody` to `ctx.DontReadResponseBody` in `onHttpRequestBody`
2. Move the logic for checking `ERROR_PARTIAL_MESSAGE_KEY` to `onHttpResponseBody`. The previous logic would not work correctly because a response may have only one header but multiple bodies.
3. Update the documentation to reflect that the cache database will be used first when both configured.


johnlanni (approved):
LGTM

PR #1790, opened 2025-02-20
fix: Fix a bug in openaiCustomUrl support

CH3CHO:

1. Support openaiCustomUrl like "https://openai.xxx.com".
2. Break the plugin initialization if error occurs when creating the provider instance.
3. Log all the errors got in parse config functions


johnlanni (approved):
LGTM

PR #1815, opened 2025-02-25
feat: Support pushing multi-arch images to a custom image registry

CH3CHO:

Support pushing multi-arch images to a custom image registry, so user can fork higress to his own GitHub repo, and use GitHub actions to build higress images.

1. `.github/workflows/build-image-and-push.yaml` is shown completed modified due to the line break change from CRLF to LF. Actually, only `Build Docker Image and Push` steps are changed.
2. Because the names of pilot and gateway images are determined by `DOCKER_TARGETS` env and the env value is fixed in the upstream repository, we are unable to change them. So extra images with name `pilot` and `proxyv2` will be pushed to the custom repository with the same image repository URL. For example, if user configures to push images with name `hub.foo.bar/higress/higress-pilot`, both `hub.foo.bar/higress/higress-pilot` and `hub.foo.bar/higress/pilot` will be pushed, referring to the same digest.


Xunzhuo:
Can you give a fork as an example?

CH3CHO:
> Can you give a fork as an example?

https://github.com/CH3CHO/higress/

https://github.com/CH3CHO/higress/actions/runs/13521147267

CH3CHO:
> https://github.com/CH3CHO/higress/
> 
> https://github.com/CH3CHO/higress/actions/runs/13521147267

Are you able to see the action run details of my own repo?

PR #1821, opened 2025-02-26
feat: Support only watching key resources in one namespace

CH3CHO:

Support only watching key resources in one namespace. User can use `--watchNamespace` parameter to restrict watching logics to a certain namespace.


johnlanni (approved):
LGTM

PR #1828, opened 2025-02-27
fix: Disable helm-docs action since it's still under development

CH3CHO:

Disable helm-docs action since it's still under development.


johnlanni (approved):
LGTM

PR #1842, opened 2025-03-02
fix: Fix the incorrect reasoning content concat logic in ai-proxy

CH3CHO:

Wrap reasoning content with `<think></think>` when preprending it to the original assistant message.

image

image

image


johnlanni (approved):
LGTM

PR #1903, opened 2025-03-15
optimize retry&failover logic

johnlanni:
Made the following changes:
1. Unified the handling mechanism for retry and failover, both based on processing the header and body before modification.
2. When all tokens are unavailable, randomly select one to use, enhancing fault tolerance.
3. Retries are also supported in streaming scenarios.
4. Fix the issue of not being able to get the path in non-header stages.


cr7258 (approved):
LGTM 🐉

PR #1914, opened 2025-03-18
fix: Skip reading non-JSON request and response bodies in ai-proxy

CH3CHO:

Skip reading non-JSON request and response bodies in ai-proxy, since we only support application/json and incorrct modifications may break the request or response


cr7258 (approved):
LGTM 🐉

PR #1927, opened 2025-03-19
feat:add mocking plugin

jujiale:
this pr add a very simple plugin, which could mock response in testing period which could return different result accroding plugin config. because other gateway have similar plugin, so I added it for dev period test.

this is a wasm plugin, very simple logic, I find other golang plugin didn't have test case.

the readme.md has described how to use it.

none

because I am a novice in higress, if the code have some problem, please help me to review it, I will fix. thanks


CH3CHO:
So what's the difference between custom-response plugin and this one? Is it possible to add more features to custom-response plugin to meet the requirement instead of adding a new plugin?

jujiale:
> So what's the difference between custom-response plugin and this one? Is it possible to add more features to custom-response plugin to meet the requirement instead of adding a new plugin?

I paid no more attention to custom-response plugin, your opinion is right, I will added feature in it.

PR #1943, opened 2025-03-24
chore: Remove redundant get-higress.sh

CH3CHO:

Remove redundant get-higress.sh. And we shall use the one in its original repository: https://github.com/higress-group/higress-standalone/blob/main/src/get-higress.sh


johnlanni (approved):
LGTM

PR #1952, opened 2025-03-26
feat: Support conditional LLM model mapping

CH3CHO:

Support conditional LLM model mapping, so user can configure difference mapping rules for different consumers, a.k.a. consumer-level model aliases.


plugins/wasm-cpp/extensions/model_mapper/plugin.cc
```diff
@@ -66,7 +68,66 @@ bool PluginRootContext::parsePluginConfig(const json& configuration,
...
     }
     auto model_mapping = it->get<Wasm::Common::JsonObject>();
-    if (!JsonObjectIterate(model_mapping, [&](std::string key) -> bool {
+    if (!parseModelMappingRule(model_mapping, rule.default_rule_)) {
+      return false;
+    }
+  }
+
+  if (!JsonArrayIterate(
+          configuration, "conditionalModelMappings", [&](const json& conditional_item) -> bool {
```
johnlanni, line 77: Need to add unit tests, and also need to provide documentation

johnlanni (changes requested):

plugins/wasm-cpp/extensions/model_mapper/plugin.cc
CH3CHO: Done

PR #1953, opened 2025-03-26
feat: add mcpServer in config map

Jing-ze:
add mcpServer in config map

add unit test


johnlanni (approved):
LGTM

PR #1987, opened 2025-03-31
feat: support service delete event trigger for tool and some fix

luoxiner:
Feat: Support service delete event triger
Fix: use stdio instead of file logs and use log level error


johnlanni (approved):
LGTM

PR #1989, opened 2025-04-01
fix: Golang filter supports skipping processing at the body stage.

Jing-ze:
add skip golang filter


johnlanni (approved):
LGTM

PR #2031, opened 2025-04-09
Feat dynamic tool reset

luoxiner:
Support delete tools when server or config delete and dynamic reset tool list when tool delete in config


johnlanni (approved):
LGTM

Issue #2000, opened 2025-04-01
token限流插件 修改McpBridge后会导致redis认证为空


Lewisyixin:
- bug概述

修改McpBridge后会导致token限流插件的redis认证为空

- 复现步骤
1. 配置了外部服务类型的redis服务
`k -nhigress-system get mcpbridges.networking.higress.io -oyaml
apiVersion: networking.higress.io/v1
kind: McpBridge
metadata:
  creationTimestamp: "2025-03-31T15:58:21Z"
  generation: 29
  name: default
  namespace: higress-system
  resourceVersion: "243407"
  uid: a3071821-737e-4214-97e6-6211580e928b
spec:
  registries:
  - domain: 10.56.23.202:6379
    name: redis-out
    port: 6379
    protocol: http
    type: static`
2. 配置好ai-token-ratelimit插件(此过程略，按照官网文档配置就行)
此时限流插件是能正常运行的，并且每次发送请求，能看到higress->redis的数据包是有auth命令的
![Image](https://github.com/user-attachments/assets/6fc8b141-ac33-4222-80b0-57d21a59a5ac)
3. 修改步骤1中(即McpBridge配置)的port配置，例如从6379改到80，再改回6379(有此类操作的原因是因为默认从dashboard上配置，只支持http协议，redis配置出来之后是80端口，所以需要手动修改一下，才发现的问题)
此时再去发送客户端请求，会发现higress->redis包命令已经不带auth参数了，并且网关也会报相关错误

![Image](https://github.com/user-attachments/assets/e8e7e354-e79f-4cd5-842d-8a0271df60aa)

![Image](https://github.com/user-attachments/assets/55154535-6b38-41b8-93a5-1d4cf0990369)
4. 但是将ai-token-ratelimit插件关闭再打开后恢复正常

想确认下这个算不算bug
### Ⅵ. Environment:

- Higress version:   higress-2.0.7
- OS :  centos7的k8s集群中,k8s 版本 v1.29.15
- Others:

johnlanni:
你修改了端口后，有修改插件配置吗？如果不修改插件配置，连的还是老的服务地址。按道理应该都不会连通，因为端口变了

Lewisyixin:
> 你修改了端口后，有修改插件配置吗？如果不修改插件配置，连的还是老的服务地址。按道理应该都不会连通，因为端口变了

[...]

PR #2032, opened 2025-04-09
fix: check NOAUTH response and init redis

Fengxq2014:
1. 增加clusterConfMap存储Init方法传入的redis配置
2. 解析redis返回内容之后判断`NOAUTH`关键字
3. 使用clusterConfMap中的配置进行初始化，如果初始化成功重新调用redisCall
4. 防止无限递归，增加ShareData进行判断是否在递归

fix: #2000

### DIFF
--- a/plugins/wasm-go/pkg/wrapper/redis_wrapper.go
+++ b/plugins/wasm-go/pkg/wrapper/redis_wrapper.go
@@ -20,9 +20,11 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"strings"
 
 	"github.com/google/uuid"
 	"github.com/higress-group/proxy-wasm-go-sdk/proxywasm"
+	"github.com/higress-group/proxy-wasm-go-sdk/proxywasm/types"
 	"github.com/tidwall/resp"
 )
 
@@ -125,7 +127,19 @@ func WithDataBase(dataBase int) optionFunc {
 	}
 }
 
+var clusterConfMap map[string]redisConfig
+
+type redisConfig struct {
+	clientClusterName string
+	username          string
+	password          string
+	timeout           uint32
+}
+
 func NewRedisClusterClient[C Cluster](cluster C) *RedisClusterClient[C] {
+	if clusterConfMap == nil {
+		clusterConfMap = make(map[string]redisConfig)
+	}
 	return &RedisClusterClient[C]{
 		cluster: cluster,
 		checkReadyFunc: func() error {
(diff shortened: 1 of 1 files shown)


plugins/wasm-go/pkg/wrapper/redis_wrapper.go
```diff
@@ -139,39 +153,68 @@ func RedisCall(cluster Cluster, respQuery []byte, callback RedisResponseCallback
...
 				}
 			}
-			if callback != nil {
-				callback(responseValue)
+		}
+		key := "redisInitialized" + requestID
+		data, cas, err := proxywasm.GetSharedData(key)
+		if strings.HasPrefix(responseValue.String(), "NOAUTH") && (errors.Is(err, types.ErrorStatusNotFound) || data == nil) {
+			if config, ok := clusterConfMap[cluster.ClusterName()]; ok {
+				for initTimes := 1; initTimes <= 2; initTimes++ {
```
johnlanni, line 193: Why init twice here?

plugins/wasm-go/pkg/wrapper/redis_wrapper.go
```diff
@@ -139,39 +153,68 @@ func RedisCall(cluster Cluster, respQuery []byte, callback RedisResponseCallback
...
+					responseValue = value
+					proxywasm.LogDebugf("redis call end, request-id: %s, respQuery: %s, respValue: %s",
+						requestID, base64.StdEncoding.EncodeToString([]byte(respQuery)), base64.StdEncoding.EncodeToString(response))
 				}
 			}
-			if callback != nil {
-				callback(responseValue)
+		}
+		key := "redisInitialized" + requestID
+		data, cas, err := proxywasm.GetSharedData(key)
```
johnlanni, line 190: It is not recommended to determine whether it is in recursion through sharedData, as this operation will add a lock at the underlying level

plugins/wasm-go/pkg/wrapper/redis_wrapper.go
```diff
@@ -139,39 +153,68 @@ func RedisCall(cluster Cluster, respQuery []byte, callback RedisResponseCallback
...
+		if strings.HasPrefix(responseValue.String(), "NOAUTH") && (errors.Is(err, types.ErrorStatusNotFound) || data == nil) {
```
johnlanni, line 191: The logic here seems incorrect as well; I think you want to reinitialize and retry only when data is nil and there is a NOAUTH error

plugins/wasm-go/pkg/wrapper/redis_wrapper.go
johnlanni: Using a variable in a closure would be more reasonable, for example: var authRetried bool

johnlanni (changes requested):

plugins/wasm-go/pkg/wrapper/redis_wrapper.go
Fengxq2014: 防止有网络波动情况，第一次初始化失败，所以进行了重试

plugins/wasm-go/pkg/wrapper/redis_wrapper.go
Fengxq2014: 使用`sharedData`是考虑了并发的情况，如果用变量的话，会不会有并发的问题？

Fengxq2014: 还有个问题是，如果reids重新初始化成功，会重新调用`proxywasm.DispatchRedisCall(cluster.ClusterName(), respQuery, redisRespCallback(cluster, respQuery, callback, requestID))`，这样`redisRespCallback`存在递归，如果变量定义在`redisRespCallback`中就回被重置

PR #2035, opened 2025-04-09
feat: support config store and redis configuration optional in mcp server

Jing-ze:
[support config handler and store in mcp server](https://github.com/alibaba/higress/commit/196b8d78277368daa9260bab9b242135b3a70d33)
[feat: support mcp server redis configuration optional](https://github.com/alibaba/higress/commit/ad64743e9c8c4ce2e81d9c62db8252720806c0b1)
[fix bug of redis not enabled response body](https://github.com/alibaba/higress/commit/d87580b86ae23b9792061b364ea48bc5ceb20d76)
[feat: add internal IP check for config path access](https://github.com/alibaba/higress/pull/2035/commits/f823b435d576015f11e73b814d4dd5e99203b3a9)
[update mcp-server.so to golang-filter.so](https://github.com/alibaba/higress/pull/2035/commits/5477f481d3db73b3534f3838521c61f77353acbd)
[fix: support panic recover in go routine and pprof in mcp server](https://github.com/alibaba/higress/pull/2035/commits/ed17fcaafbad058282b8238ffa4a525dce8a00ba)


plugins/golang-filter/mcp-server/handler/rate_limit_handler.go
```diff
@@ -0,0 +1,102 @@
...
+func (h *MCPRatelimitHandler) HandleRatelimit(path string, method string, body []byte) bool {
+	parts := strings.Split(path, "/")
+	if len(parts) < 3 {
+		return false
+	}
+	serverName := parts[1]
+	uid := parts[2]
+
+	// Build rate limit key using serverName and uid
+	limitKey := fmt.Sprintf("limit:%s:%s", serverName, uid)
```
johnlanni, line 70: window and limit also need to be part of the key

johnlanni (changes requested):

plugins/golang-filter/mcp-server/filter.go
```diff
@@ -99,6 +122,26 @@ func (f *filter) DecodeHeaders(header api.RequestHeaderMap, endStream bool) api.
 	}
 	if !strings.HasSuffix(url.parsedURL.Path, f.config.ssePathSuffix) {
 		f.proxyURL = url.parsedURL
+		if f.config.enableUserLevelServer {
```
johnlanni, line 125: When enabling user-granularity server, access should be denied when the path does not carry a uid or when the configuration corresponding to the uid cannot be found

Jing-ze: When the configuration for the uid is not found, it suppose to be use a default trial API-Key other than access denied?

johnlanni (approved):
LGTM

PR #2043, opened 2025-04-11
feat:add GetContextId func for HttpContext

hzhswyz:
Expose the contextId property of CommonHttpCtx so that the proxywasm.SetEffectivenessContext function can be called in tickFunc


johnlanni (approved):
LGTM

PR #2046, opened 2025-04-11
fix: fix param mapping use %v instead of %s

luoxiner:
Fix param mapping in mcp nacos registry, use %v instead of %s for param format


johnlanni (approved):
LGTM

PR #2061, opened 2025-04-15
feat: Support building waf plugin using Makefile

CH3CHO:

Update the wasm-go building Dockerfile to support building waf plugin.


rinfx (approved):
LGTM

PR #2070, opened 2025-04-15
rm plugin id after use

rinfx:
`_plugin_id_` is only used to mark a plugin, should be removed after the plugin id is extracted, to avoid the sdk thinking 
 the plugin has global configuration


johnlanni (approved):
LGTM

PR #2074, opened 2025-04-16
fix: support mcp server database reconnect and fix tool/list method denied

Jing-ze:
support mcp server database reconnect
fix bug of tool/list method denied due to uid missing


johnlanni (approved):
LGTM

PR #2090, opened 2025-04-19
fix: update module replacements

Similarityoung:

Modified the replace directive in go.mod to use a relative path (../..) instead of an absolute path (/Users/zhangty/tmp/higress/plugins/wasm-go).


CH3CHO (approved):
LGTM. Thanks.

PR #2120, opened 2025-04-24
fix: Refactor MCP Server into MCP Session and MCP Server

Jing-ze:
Refactor MCP Server into MCP Session and MCP Server
Support refresh TTL in each get config request


plugins/golang-filter/mcp-session/common/utils.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+func NewRequestURL(header api.RequestHeaderMap) *RequestURL {
+	method, _ := header.Get(":method")
+	scheme, _ := header.Get(":scheme")
+	host, _ := header.Get(":authority")
+	path, _ := header.Get(":path")
+	internalIP, _ := header.Get("x-envoy-internal")
+	baseURL := fmt.Sprintf("%s://%s", scheme, host)
+	parsedURL, _ := url.Parse(path)
```
johnlanni, line 27: The error needs to be handled; if parsing fails, parsedURL will return nil, and similar places also need to handle errors and nil

johnlanni (changes requested):

plugins/golang-filter/mcp-server/filter.go
```diff
@@ -1,104 +1,41 @@
...
-	f.path = url.parsedURL.Path
-
-	// Check if request matches any rule in match_list
-	if !internal.IsMatch(f.config.matchList, url.host, f.path) {
-		f.skip = true
-		api.LogDebugf("Request does not match any rule in match_list: %s", url.parsedURL.String())
-		return api.Continue
-	}
+	url := common.NewRequestURL(header)
+	f.path = url.ParsedURL.Path
```
johnlanni, line 24: For example, using url.ParsedURL here may cause a crash

johnlanni:
image
This dependency should be replaced with: https://github.com/higress-group/envoy

plugins/golang-filter/mcp-session/config.go
```diff
@@ -0,0 +1,143 @@
...
+const ConfigPathSuffix = "/config"
+const DefaultServerName = "higress-mcp-server"
+
+var GlobalSSEPathSuffix = "/sse"
+
+type config struct {
+	matchList             []common.MatchRule
+	enableUserLevelServer bool
+	rateLimitConfig       *handler.MCPRatelimitConfig
+	defaultServer         *common.SSEServer
```
zth9: Is this config globally shared? Putting defaultServer into the config may result in the issue of defaultServer overlapping in concurrent situations. @Jing-ze

PR #2237, opened 2025-05-14
Fix : add fail strategy for wasmplugin generated by mcp server

Erica177:


registry/memory/cache.go
```diff
@@ -104,6 +104,7 @@ func (s *store) GetAllConfigs(kind config.GroupVersionKind) map[string]*config.C
 			Priority:        &wrapperspb.Int32Value{Value: 30},
```
johnlanni, line 104: 这个优先级改到999吧

johnlanni (changes requested):

registry/memory/cache.go
Erica177: done

johnlanni (approved):
LGTM

PR #2300, opened 2025-05-22
[HIGRESS#2284] Enhance the e2e testing of the moder-router plugin bas…

undertaker86001:

Related to [HIGRESS#2284](https://github.com/alibaba/higress/issues/2284)

- Provider Header Addition Functionality
  - Extract provider information from model parameters (e.g., "qwen" from "qwen/qwen-long")
  - Add provider information to specified request headers
  - Simultaneously modify the model parameter in the request body, removing the provider prefix
- Model Header Addition Functionality
  - Add the complete model parameter value directly to specified request headers
  - Do not modify the original request body
 - Custom modelKey Configuration
  - Support extracting model information from different fields in the request body (e.g., "engine")
  - Verify that custom field name configurations work correctly
- Path Matching Functionality
  - Matching Path Suffix Processing: Test that when request path suffixes match the configured enableOnPathSuffix list, the 
plugin processes requests normally. For example, when configured with ["/v1/embeddings", "/v1/chat/completions"], requests to ```/v1/embeddings``` paths will be processed, and model parameters will be extracted and added to request headers. 

- Non-matching Path Skip Processing: Test that when request path suffixes are not in the configured enableOnPathSuffix list, the plugin skips processing. For example, requests to ```/v1/models``` paths will not be processed, no additional request headers will be added, and requests will be forwarded to backend services unchanged.

- Wildcard Path Matching: Support using "" as a wildcard to match all paths. When configured as "", all requests will be processed without path restrictions.

- Composite Functionality Testing
  - Test behavior when both ```addProviderHeader ```and ```modelToHeader ``` are configured simultaneously
  - Verify that the plugin can extract both provider information and complete model parameters, adding them to different request headers respectively
  - Ensure that model parameters in the request body are correctly modified (provider prefix removed)
- Route-level Configuration
  - Test configuration matching based on routing rules, verifying that different Ingresses can apply different model-router configurations
  - Ensure the plugin can correctly apply corresponding configuration rules based on the target host of requests

### DIFF
--- a/test/e2e/conformance/tests/cpp-wasm-model-router.yaml
+++ b/test/e2e/conformance/tests/cpp-wasm-model-router.yaml
@@ -131,3 +131,4 @@ spec:
         modelToHeader: "x-higress-llm-model"
       ingress:
         - higress-conformance-ai-backend/wasmplugin-model-router-combined
+  url: oci://higress-registry.cn-hangzhou.cr.aliyuncs.com/plugins/model-router:1.0.0
--- a/test/e2e/conformance/tests/cpp-wasm-model-router.yaml
+++ b/test/e2e/conformance/tests/cpp-wasm-model-router.yaml
@@ -1,3 +1,17 @@
+# Copyright (c) 2022 Alibaba Group Holding Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


test/e2e/conformance/tests/cpp-wasm-model-router.yaml
```diff
@@ -0,0 +1,133 @@
...
+---
+apiVersion: extensions.higress.io/v1alpha1
+kind: WasmPlugin
+metadata:
+  name: model-router
+  namespace: higress-system
+spec:
+  defaultConfigDisable: true
+  phase: UNSPECIFIED_PHASE
+  priority: 100
```
johnlanni, line 104: The OCI address configuration for the plugin is missing

johnlanni (changes requested):

test/e2e/conformance/tests/cpp-wasm-model-router.yaml
undertaker86001: it's solved

PR #2351, opened 2025-05-30
fix: set "EnableSemanticCachefalse" to false when no vector configured in ai-cache

mirror58229:

Set "EnableSemanticCachefalse" to false when no vector configured in ai-cache, which avoids error log in `handleResponse`

https://github.com/alibaba/higress/blob/main/plugins/wasm-go/extensions/ai-cache/core.go#52


johnlanni (approved):
LGTM

Issue #2396, opened 2025-06-11
[bug] AWS Bedrock signature mismatch when using modelId with colon (:) — path is over-encoded


xiandan-erizo:
- [x] I have searched the [issues](https://github.com/alibaba/higress/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description

When using Higress with the AI Proxy plugin to forward requests to AWS Bedrock, I encountered a signature mismatch error when calling the `converse` API with a standard (non-ARN) modelId such as: `anthropic.claude-3-5-sonnet-20240620-v1:0`

The same request works perfectly with `boto3`, using the same AK/SK and region.

---

### Ⅱ. Describe what happened

The request returns 403 from AWS, with the following error message:

The request signature we calculated does not match the signature you provided.

The Canonical String for this request should have been
POST
/model/anthropic.claude-3-5-sonnet-20240620-v1%25253A0/converse
host:bedrock-runtime.us-east-1.amazonaws.com
x-amz-date:20250611T072336Z
...

The String-to-Sign should have been
AWS4-HMAC-SHA256
...

From this, we can see that the colon (`:`) in the `modelId` has been triple-encoded:  
`: → %3A → %253A → %25253A`

This leads to signature mismatch on AWS’s side.

---

### Ⅲ. Describe what you expected to happen

I expect:

- The plugin or WASM logic to correctly encode the `modelId` only once using `PathEscape` (i.e., `:` → `%3A`);
- The actual request path and the path used for signature to match exactly;
- Ideally, this should follow the same behavior as boto3 without requiring users to manually encode `%` characters;
- Or, clear guidance on where and how encoding should be applied when using modelIds in path-based Bedrock APIs.

---

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. Configure AI Proxy provider as `bedrock` in Higress
2. Sign requests manually via WASM extension
3. Make this request:
```json
POST /model/anthropic.claude-3-5-sonnet-20240620-v1:0/converse
Host: bedrock-runtime.us-east-1.amazonaws.com
{
  "model": "anthropic.claude-3-5-sonnet-20240620-v1:0",
  "anthropic_version": "bedrock-2023-05-31",
  "messages": [
    { "role": "user", "content": "Hello!" }
  ]
}
```

### Ⅴ. Anything else we need to know?
1. The same AK/SK and modelId work fine via Python boto3
2. This issue only happens when colon (:) is used in modelId
[...]

PR #2402, opened 2025-06-12
fix(ai-proxy): fix bedrock Sigv4 mismatch

HecarimV:
fix bedrock Sigv4 mismatch
ref: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html
###### Path Encoding (Canonical URI)
   - Encode all characters except unreserved characters (`A-Z`, `a-z`, `0-9`, `-`, `_`, `.`, `~`) and reserved forward slashes (`/`).  
   - Example: `/my path` → `/my%20path` (space encoded as `%20`, **not** `+`).  
   - Preserve forward slashes (`/`) without encoding
   - Always use uppercase for encoded characters (e.g., `%2F` instead of `%2f`). 

fixes: #2396


johnlanni (approved):
LGTM

PR #2469, opened 2025-06-19
fix too much logs when nacos is not avaiable

luoxiner:

fix too much logs when nacos is not avaiable

There will be too much logs when nacos is not avaiable for mcp server discovery.


johnlanni (approved):
LGTM

PR #2507, opened 2025-06-27
fix: Error compiling golang-filter.so on arm64 machine (#2494)

### LINKED ISSUE #2494: Error compiling golang-filter.so on arm64 machine
### Ⅰ. Issue Description

An exception occurred when compiling golang-filter using the `make build-golang-filter` command on a local arm64 machine.
Image

It seems that the Dockerfile does not specify the installation of the x86 cross-compilation toolchain.
After specifying it like arm, it can be compiled normally. Do you need to modify it like this?

Image

---
### Ⅰ. 问题描述
在本地 arm64 机器上使用 `make build-golang-filter` 命令编译 golang-filter 异常。

Image

看 Dockerfile 中像是没有指定安装 x86 交叉编译的工具链。
像 arm 那样指定后，可以正常编译出来。看是否需要这样修改下呢？

Image

hongzhouzi:
Fix error compiling golang-filter.so on arm64 machine.

If you do not specify a specific architecture package, the tools for the corresponding architecture package will be installed. This will cause the x86 compiler tool to be installed on the arm machine, causing an error.

fixes #2494

### DIFF
--- a/plugins/golang-filter/Dockerfile
+++ b/plugins/golang-filter/Dockerfile
@@ -17,7 +17,7 @@ RUN if [ "$GOARCH" = "arm64" ]; then \
     else \
         echo "Installing AMD64 toolchain" && \
         apt-get update && \
-        apt-get install -y gcc binutils; \
+        apt-get install -y gcc-x86-64-linux-gnu binutils-x86-64-linux-gnu; \
     fi
 
 WORKDIR /workspace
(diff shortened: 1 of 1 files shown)


johnlanni:
The issue is with compiling the amd64 target on an arm64 machine, and this adjustment should fix it, right?
However, we need to verify that there are no issues when compiling the amd64 target on an amd64 machine using this method. The stability needs to be verified. This PR will only be merged after the verification is complete.

cr7258 (approved):
Verified on both arm64 and amd64 machines.

Jing-ze:
test on macos arm64 and linux amd64, it works fine.

johnlanni (approved):
LGTM

PR #2608, opened 2025-07-16
feat(ai-proxy): add consumerBinding

daixijun:

添加 Consumer 与 apiToken 的绑定关系，以实现同一个 Provider 有多个账户（多套 apiToken）情况下，可以指定某个 consumer 允许调用哪些账户下的模型，默认如果 consumer 没有配置，则是轮询所有的 apiToken

配置方法如下

```yaml
providers:
  - id: openai
    type: openai
    apiTokens:
      - "sk-abcd"
      - "sk-efgh"
    consumerBinding:
      consumer1:
        - sk-abcd
      consumer2:
        - sk-efgh
      consumer3:
        - sk-abcd
        - sk-efgh
```


CH3CHO:
我并不认为这是个正确的解法。consumer 和 provider 应该都是原子化的。如果需要在 provider 内部进一步细分，那么应该把它拆分成不同的 provider。

---
I don't think this is the correct solution. Consumer and provider should be atomized. If you need to further subdivided inside the provider, you should split it into different providers.

daixijun:
> 我并不认为这是个正确的解法。consumer 和 provider 应该都是原子化的。如果需要在 provider 内部进一步细分，那么应该把它拆分成不同的 provider。
> 
> I don't think this is the correct solution. Consumer and provider should be atomized. If you need to further subdivided inside the provider, you should split it into different providers.

我现在的做法也是拆成两个 provider 来配置， 但有个问题， 因为模型名都是一样的， 在路由上就没法做区分了
我不得不将其中一套 provider 下的模型名称全部加上特定的前缀配合 modelMapping 来实现

CH3CHO:
> 我现在的做法也是拆成两个 provider 来配置， 但有个问题， 因为模型名都是一样的， 在路由上就没法做区分了
> 我不得不将其中一套 provider 下的模型名称全部加上特定的前缀配合 modelMapping 来实现

可以考虑加一些使用 x-mse-consumer 这类认证头进行匹配的路由规则来实现。

---
> My current approach is to split it into two providers to configure it, but there is a problem, because the model names are the same, so there is no way to distinguish them in the route.
> I had to add a specific prefix to modelMapping to implement the model name under one of the sets of providers

You can consider adding some routing rules that use authentication headers like x-mse-consumer to match.

daixijun:
就是说还创建两套provider配置，通过x-mse-consumer头来匹配使用哪个provider么 

---
That is to say, two sets of provider configurations are created, and the x-mse-consumer header is used to match which provider is used

CH3CHO:
> 就是说还创建两套provider配置，通过x-mse-consumer头来匹配使用哪个provider么

是的

---
> That is to say, two sets of provider configurations are created, and should I use the x-mse-consumer header to match which provider to use

Yes

PR #2710, opened 2025-08-04
feat: Support adding a proxy server in between when forwarding requests to upstream

### LINKED ISSUE #2664: Support adding a proxy server in between when forwarding requests to upstream
## Why do you need it?

Due to the network limitations, such as Intranet/Internet barrier, AI technology sactions to China, we may need a proxy server to gain the access to the actual upstream server. But Higress doesn't support it out-of-box at the moment.

## How could it be?

TBD

## Other related information

- [Question: Can envoy be configured to use a second http/https proxy?](https://github.com/envoyproxy/envoy/issues/21175)

CH3CHO:

Support adding a proxy server in between when forwarding requests to upstream.

fixes #2664 

TBC


pkg/ingress/config/ingress_config.go
```diff
@@ -669,11 +681,18 @@ func (m *IngressConfig) convertEnvoyFilter(convertOptions *common.ConvertOptions
...
+			envoyFilters = append(envoyFilters, *ef)
+		}
+	}
+
 	// TODO Support other envoy filters
 
 	IngressLog.Infof("Found %d number of envoyFilters", len(envoyFilters))
 	m.mutex.Lock()
 	m.cachedEnvoyFilters = envoyFilters
+	// TODO (by CH3CHO): Make sure pilot can get the latest envoy filters.
```
johnlanni: What is the reason for this TODO?

johnlanni: What is the reason for this TODO?

PR #2723, opened 2025-08-06
fix: Fix the malfunction of _match_service_ rules in C++ Wasm plugins

CH3CHO:

Fix the malfunction of _match_service_ rules in C++ Wasm plugins.

The original implementation uses an incorrect property name when trying to get the current cluster name. Change it to the correct one and `_match_service_` rules work.

https://github.com/alibaba/higress/blob/main/plugins/wasm-cpp/common/route_rule_matcher.h#L166
image

https://github.com/alibaba/higress/blob/main/plugins/wasm-cpp/common/route_rule_matcher.h#L246
image


johnlanni (approved):
LGTM

PR #2743, opened 2025-08-11
fix(ip-restriction): fix bug of set ip_source_type

Jing-ze:
fix bug of set ip_source_type to origin-source

unit test will add later with wasm unit test framework


johnlanni (approved):
LGTM

PR #2867, opened 2025-09-04
feat(gzip): add gzip configuration support and update default settings

### LINKED ISSUE #2862: The default parameters of higress helm need to support enabling gzip to avoid generating excessive traffic.
在测试higress-gateway性能时，qps达到了14000多，CPU使用在8C左右，但是网络流量为什么这个大啊，higress-gateway的出站流量有100MB左右，但是apisix的只有10KB左右，到底是什么导致了higress-gateway产生这么大的出站流量啊，图如下所示：

Image

---
When testing the performance of higress-gateway, qps reached more than 14,000, and the CPU was used at around 8C. However, why is the network traffic so large? The outbound traffic of higress-gateway is about 100MB, but the apisix is ​​only about 10KB. What exactly caused higress-gateway to generate such a large outbound traffic? The figure is as follows:

Image

Aias00:
add gzip configuration support and update default settings

fixes #2862

### DIFF
--- a/helm/core/templates/configmap.yaml
+++ b/helm/core/templates/configmap.yaml
@@ -112,6 +112,9 @@ data:
     {{- if hasKey .Values "downstream" }}
     {{- $_ := set $newData "downstream" .Values.downstream }}
     {{- end }}
+    {{- if hasKey .Values "gzip" }}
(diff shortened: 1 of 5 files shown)


johnlanni (approved):
LGTM

PR #2943, opened 2025-09-23
release note supports system prompt

Guo-Chenxu:

support self-define system prompt while generating release notes

the content between `## system prompt` and the next `##` is system prompt, like this:

```
## system prompt

here is your self-define system prompt

## ...
```


johnlanni (approved):
LGTM

PR #3074, opened 2025-10-30
fix(log-request-response): enhance response body logging by checking Content-Encoding

Jing-ze:
Add content-encoding check in log-request-response plugin to skip logging compressed request/response bodies, avoiding garbled text in access logs.

No.

N/A - This PR only adds defensive checks. Existing tests should still pass.

1. Configure the plugin to log request/response bodies
2. Send a request with Content-Encoding: gzip header
3. Verify that the compressed body is not logged in access logs
4. Verify that uncompressed bodies are still logged normally


johnlanni (approved):
LGTM

PR #3168, opened 2025-11-26
Fix OpenAI capability rewrite dropping query string

wydream:

对于带正则的路径，将查询字符串从原始路径中剥离，避免干扰正则匹配，在匹配完成后将查询参数拼回

1. strip query parameters before running capability regex matching so IDs resolve correctly
2. append the preserved query string back onto the rewritten path, avoiding regressions for endpoints like videos/.../content

---

For paths with regular expressions, the query string is stripped from the original path to avoid interfering with the regular matching, and the query parameters are put back together after the matching is completed.

1. strip query parameters before running capability regex matching so IDs resolve correctly
2. append the preserved query string back onto the rewritten path, avoiding regressions for endpoints like videos/.../content


johnlanni (approved):
LGTM

PR #3355, opened 2026-01-18
feat: Implement CRD version validation at server startup, providing warnings and update instructions for outdated CRDs

### LINKED ISSUE #2278: Add version check to Higress CRDs
## Why do you need it?

Some key functions in Higress rely on its CRDs. When upgrading an existed Higress instance using Helm, CRDs won't be updated, unless their new definitions are applied manually. However, some features won't work properly without the correct version of CRDs, such as Nacos 3 MCP Server integration.

## How could it be?

Add a CRD version check into Higress Controller or Higress Console. Notify user in the log or on the web page about any version discrepancy.

## Other related information

N/A

Aias00:
Implement CRD version validation at server startup, providing warnings and update instructions for outdated CRDs
fixes: #2278

- [ ] **For new standalone features** (e.g., new wasm plugin or golang-filter plugin):
  - [ ] I have created a `design/` directory in the plugin folder
  - [ ] I have added the design document to the `design/` directory
  - [ ] I have included the AI Coding summary below

- [ ] **For regular updates/changes** (not new plugins):
  - [ ] I have provided the prompts/instructions I gave to the AI Coding tool below
  - [ ] I have included the AI Coding summary below


pkg/kube/crd_version.go
```diff
@@ -0,0 +1,202 @@
...
+//  2. Add the API definition to api/extensions/ or api/networking/
+//  3. Add an entry here with the expected version and required fields
+//  4. Update tests to verify the CRD
+//
+// CRD Information Sources:
+//   - Name: From CRD metadata.name in helm/core/crds/customresourcedefinitions.gen.yaml
+//   - ExpectedVersion: From CRD spec.versions[].name (the storage version)
+//   - RequiredFields: From CRD spec.versions[].schema.openAPIV3Schema.properties
+//   - Description: From API protobuf comments and CRD usage in code
+var RequiredCRDs = []CRDVersionInfo{
```
CH3CHO: Is it possible to detect these CRDs at runtime?

Aias00: Installed CRDs are already detected at runtime in `CheckCRDVersions()` via the apiextensions API `List()` call. The explicit `RequiredCRDs` slice is the expected compatibility contract: without it we cannot detect a CRD that is missing entirely, and we also cannot infer which version/fields are feature-critical for this binary from cluster state alone. I clarified that boundary in the comment block above `RequiredCRDs`.

CH3CHO: However, this may impose a maintainence issue. Everytime the CRDs are updated, we need to update `pkg/kube/crd_version.go` as well. Would it be better to use reflection to get all the fields of CRD structs and compare with the corresponding CRD resources? And we can support configuring an "optional" field list for each CRD which allow some fields to bypass the check.

Aias00: Yes, I think that direction makes sense. The important part is to auto-derive the expected CRD contract instead of manually maintaining it in `pkg/kube/crd_version.go`.

  I would prefer to implement that from the generated CRD manifest/schema (`customresourcedefinitions.gen.yaml`) rather than from plain Go reflection, because the manifest is the actual CRD source of truth and
  preserves schema semantics such as the storage version and OpenAPI schema structure.

  So conceptually it is still the same automation idea you suggested:
  - derive the expected fields automatically
  - compare with the live CRD resources
  - optionally support a small ignore/optional field list where needed

  I just think the manifest/schema-driven approach is a more accurate implementation of that idea than reflecting over Go structs directly.

CH3CHO: I'm OK with this approach. Could you update the implementation accordingly?

Aias00: Updated this direction in code. The expected CRD contract is now derived automatically from the generated CRD manifest (`api/kubernetes/customresourcedefinitions.gen.yaml`) instead of being hand-maintained in `pkg/kube/crd_version.go`. At runtime we still discover the live CRDs from the cluster, but the expected CRD name, storage version, and shipped storage schema now come from the generated manifest, with the comparison performed against the live storage schema.

pkg/kube/crd_version.go
```diff
@@ -0,0 +1,273 @@
...
+
+// CheckCRDVersions checks if all required CRDs exist with correct versions
+// Returns a list of warning messages if any issues are found
+func CheckCRDVersions(config *rest.Config) []string {
+	requiredCRDs, err := loadExpectedCRDContracts()
+	if err != nil {
+		return []string{fmt.Sprintf("Failed to load generated CRD contracts: %v", err)}
+	}
+
+	apiExtClientset, err := apiExtensionsClient.NewForConfig(config)
```
CH3CHO: 为啥这个变量名字里有个 `set`，而且 `s` 还没有大写？

pkg/kube/crd_version.go
```diff
@@ -0,0 +1,273 @@
...
+		if path == ignored || strings.HasPrefix(path, ignored+".") {
+			return true
+		}
+	}
+	return false
+}
+
+// fieldExistsInSchema checks if a field path exists in the schema
+// Field path format: "spec.fieldName" or "spec.nested.fieldName"
+func fieldExistsInSchema(schema *apiExtensionsV1.JSONSchemaProps, fieldPath string) bool {
```
CH3CHO: 这个函数每次都解析一遍这个 schema，是否可以改进呢？

pkg/kube/crd_version.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	paths := map[string]struct{}{}
+	collectSchemaPathsRecursive(&specSchema, "spec", paths)
+
+	collected := make([]string, 0, len(paths))
+	for path := range paths {
+		collected = append(collected, path)
+	}
+	sort.Strings(collected)
```
CH3CHO: 这里为什么要排序呢？

pkg/kube/crd_version.go
```diff
@@ -0,0 +1,273 @@
...
+		if !fieldExistsInSchema(liveSchema, field) {
+			missing = append(missing, field)
+		}
+	}
+
+	return missing
+}
+
+// checkRequiredFields is kept as a small helper for focused schema-path tests.
+func checkRequiredFields(schema *apiExtensionsV1.JSONSchemaProps, requiredFields []string) []string {
```
CH3CHO: 这个函数好像没有实际被用到？

pkg/kube/crd_version.go
Aias00: 已经修改了，现在改成先分别把 shipped schema 和 live schema 转成字段路径集合，再直接做差集比较，不会再对 live schema 按字段逐个重复解析。

另外，排序也收敛到了最终 missing fields 输出时才做，只用于保证 warning 和测试结果稳定

pkg/kube/crd_version.go
```diff
@@ -0,0 +1,227 @@
...
+		})
+	}
+
+	return requiredCRDs, nil
+}
+
+func checkCRDVersionsWithClient(client apiExtensionsV1Client.CustomResourceDefinitionInterface, requiredCRDs []CRDVersionInfo, optionalFieldPaths map[string][]string) []string {
+	warnings := []string{}
+
+	crdList, err := client.List(context.TODO(), metaV1.ListOptions{})
```
CH3CHO: `CheckCRDVersions` is called synchronously from `Server.Start` before the server starts its listeners, but this request uses `context.TODO()` without a deadline or cancellation tied to the server's `stop` channel.

If the API server accepts the connection but stalls while serving the CRD list—for example during an etcd or API server degradation—startup can remain blocked here indefinitely and cannot respond to the stop signal.

Could we pass a context into `CheckCRDVersions` and apply a bounded timeout, preferably also cancelling it when `stop` is closed? Since this is a diagnostic warning check, it should not be able to block server startup indefinitely.

CH3CHO (commented):
I found two medium-severity issues worth addressing or clarifying: startup cancellation for the CRD API request, and the field-contract maintenance model after the earlier manifest-derived implementation was narrowed. The current CI checks are green.

pkg/kube/crd_version.go
```diff
@@ -0,0 +1,227 @@
...
+)
+
+// CRDVersionInfo contains the expected CRD contract derived from the shipped manifest.
+type CRDVersionInfo struct {
+	Name            string
+	ExpectedVersion string
+	RequiredFields  []string
+}
+
+var criticalCRDFieldPaths = map[string][]string{
```
CH3CHO, line 52: The earlier resolved review thread concluded that the expected CRD schema would be derived from `customresourcedefinitions.gen.yaml`, and the follow-up stated that the shipped storage schema was being compared with the live schema.

In the current HEAD, however, `CRDContract` only derives the CRD name and storage version from the manifest, while field validation is again driven by this manually maintained `criticalCRDFieldPaths` map. Consequently, adding a feature-critical field to the generated manifest does not automatically add it to startup validation unless this separate map is also updated.

Was this partial revert intentional? If the reduced warning surface is desired, could we at least add a test that requires every shipped `spec` field to be either included in `criticalCRDFieldPaths` or explicitly excluded through `optionalCRDFieldPaths`? Otherwise, the maintenance gap discussed in the resolved thread remains.

pkg/kube/crd_version.go
Aias00: Addressed in 1312f5a5.

`CheckCRDVersions` now takes a `context.Context`, and `Server.Start` constructs it via `context.WithTimeout(30s)` that is also cancelled when the `stop` channel closes (a goroutine bridges `stop` → `cancel`). `checkCRDVersionsWithClient` fail-fasts on `ctx.Err()` before issuing the list request, so a stalled API server cannot block startup indefinitely and a stop signal is honoured promptly. The real REST client also honours `ctx` for the in-flight CRD list HTTP call.

A cancelled/expired context surfaces as a single `CRD version check skipped: ...` warning rather than failing startup, since this is a diagnostic check.

Covered by `TestCheckCRDVersionsWithClient_CancelledContext`, which pre-cancels the context and asserts the list reactor is never invoked and a "skipped" warning is returned.

pkg/kube/crd_version.go
Aias00: Addressed in 1312f5a5.

The reduced warning surface is intentional and kept (only `criticalCRDFieldPaths` is validated at runtime), but the maintenance gap is now closed by a guard test rather than relying on manual discipline:

- `optionalCRDFieldPaths` is populated with the remaining shipped `spec.*` fields for each CRD.
- `TestCriticalAndOptionalCRDFieldPaths_ClassifyAllShippedSpecFields` loads the shipped storage schemas (via the new `LoadCRDStorageSchemas()` in `api/kubernetes/crd_contract.go`) and requires every shipped `spec` field to be classified in `critical ∪ optional`, and vice-versa (no stale entries, no duplicates). Adding a new field to `customresourcedefinitions.gen.yaml` therefore makes the test fail until it is explicitly classified — so the critical set cannot silently drift from the shipped schema.

The maintenance contract is documented in the comment block above `criticalCRDFieldPaths`.

johnlanni (approved):
## Review: feat: Implement CRD version validation at server startup

绑定 head: `5f4e12e5aa310acbff9162d3d10d4684abfb30f4`（base: `55dac5ed`）
关联 Issue: #2278（Helm 升级不会更新 CRD，需要在启动时检测并提示版本差异）

### 审查范围（三点 diff，5 个新文件，+1285 行）
- `api/kubernetes/crd_contract.go` / `_test.go`：embed 内嵌 `customresourcedefinitions.gen.yaml`，解析出各 CRD 的 storage version 契约与 schema（sync.Once 缓存）。
- `pkg/kube/crd_version.go` / `_test.go`：启动时 List 集群 CRD，校验 storage version 与关键 spec 字段路径；critical/optional 字段分类由维护测试强制覆盖全部 shipped spec 字段，防止漂移。
- `pkg/bootstrap/server.go`：`Server.Start` 中以 30s timeout + stop 通道联动取消的 context 调用检查，仅输出 Warn 日志，不阻断启动。

### 重点核查结论
1. **不会阻断控制面启动**：所有错误路径（契约加载失败、client 创建失败、List 失败、ctx 取消/超时、CRD 缺失、无 storage version、schema 缺失、字段缺失）均降级为 warning，`Start` 继续执行。✔
2. **超时/停止路径**：检查前有 `ctx.Err()` fail-fast；`defer cancel` + goroutine 桥接 `stop→cancel`，最坏只延迟启动 30s，且收到停止信号立即取消；goroutine 在 ctx 完成后退出，无泄漏。✔ 上一轮 CH3CHO 提出的 context.TODO() 无界问题已在 `1312f5a` 解决并有 `TestCheckCRDVersionsWithClient_CancelledContext` 回归。
3. **版本比较健壮性**：以 shipped manifest 的 storage version 名为契约（wasmplugins=v1alpha1，http2rpcs/mcpbridges=v1），比较为精确字符串匹配并附带当前可用版本列表，语义清晰；字段检查基于 schema 路径集合，支持数组 Items 与 x-kubernetes-preserve-unknown-fields 截断。✔
4. **权限/离线降级**：`helm/core/templates/controller-clusterrole.yaml:70-73` 已有 `apiextensions.k8s.io/customresourcedefinitions` 的 get/list/watch；无权限或 API server 不可达时仅产生一条 warning。✔
5. **无 nil 风险**：`NewServer` 的 initFuncList 中 `initKubeClient` 无条件先行且失败即返回错误，`Start` 中 `s.kubeClient.RESTConfig()` 不会 panic。✔
6. **升级指引准确性**：提示的两条路径 `api/kubernetes/customresourcedefinitions.gen.yaml` 与 `helm/core/crds/customresourcedefinitions.gen.yaml` 在仓库中确实存在且内容一致（diff 校验通过）。✔
7. **并发/资源**：embed+sync.Once 只读缓存安全；每次启动新建一个轻量 apiextensions client，无泄漏。✔

### 测试与命令证据
- `go test ./api/kubernetes/... -count=1` → **ok**（本地执行，验证 3 条 CRD 契约与 storage version）。
- `gofmt -l`（5 个变更文件）→ 无输出；`go vet ./api/kubernetes/...` → 通过。
- `pkg/kube` 测试本地无法编译：`external/` 目录由 prebuild 从 git submodule 生成且被 gitignore，本只读环境禁止初始化/拉取 submodule（已按约束记录此限制，未改变 checkout）。该包测试已由 CI `coverage-test`（SUCCESS，head `5f4e12e`）覆盖，含 storage version 不匹配、List 失败、字段缺失、optional 旁路、ctx 取消、字段分类守护共 800+ 行用例。
- CI 状态：12 项 check（build、lint、coverage-test、gateway/higress-conformance-test、DCO、CLA、fossa 等）在 exact head 上全部 SUCCESS。

### 非阻塞建议（P3，不阻塞合入）
1. `pkg/bootstrap/server.go`：ctx 在 List 期间超时时日志为 "Failed to list CRDs: context deadline exceeded"，可考虑统一为 "CRD version check skipped" 语义，便于运维区分"环境问题跳过"与"真实失败"。
2. 升级指引中 `helm/core/crds/...` 是源码树路径，使用发布版 Helm chart 的用户可能没有本地副本，可补充 helm 源安装方式的示例（如 `helm show crds`）。
3. 日志中的 emoji（⚠️/✅）与仓库既有日志风格略有差异，属风格偏好。

### 结论
功能与 Issue #2278 诉求一致，错误路径全部安全降级，无 P0/P1 问题，CI 全绿，**APPROVE**。

PR #3424, opened 2026-02-01
feat(ai-statistics): support token details and builtin keys for reasoning_tokens/cached_tokens

johnlanni:
## What type of PR is this?

/kind feature

## What this PR does / why we need it:

This PR enhances the ai-statistics plugin to support recording detailed token usage information, specifically:

1. **New builtin attribute keys**:
   - `reasoning_tokens` - Extract reasoning token count from `output_token_details.reasoning_tokens` (for o1 and other reasoning models)
   - `cached_tokens` - Extract cached token count from `input_token_details.cached_tokens` (for prompt caching)
   - `input_token_details` - Record complete input token details as JSON
   - `output_token_details` - Record complete output token details as JSON

2. **Use cases**:
   - **Cost optimization**: Track cache hit rates to optimize prompt caching strategy
   - **Performance analysis**: Monitor reasoning token usage for reasoning models
   - **Usage statistics**: Fine-grained statistics of various token types

3. **Configuration example**:
```yaml
attributes:
  - key: reasoning_tokens
    apply_to_log: true
  - key: cached_tokens
    apply_to_log: true
  - key: input_token_details
    apply_to_log: true
  - key: output_token_details
    apply_to_log: true
```

## Which issue(s) this PR fixes:

Fixes #

## Does this PR introduce a user-facing change?

```release-note
ai-statistics plugin now supports recording token details including reasoning_tokens and cached_tokens for OpenAI Chat Completions
```

### DIFF
--- a/plugins/wasm-go/extensions/ai-statistics/main.go
+++ b/plugins/wasm-go/extensions/ai-statistics/main.go
@@ -969,9 +969,7 @@ func debugLogAiLog(ctx wrapper.HttpContext) {
 
 	// Get all user attributes as a map
 	userAttrs := make(map[string]interface{})
-	// Get all user attributes as a map
-	userAttrs := make(map[string]interface{})
-	
+
 	// Try to reconstruct from GetUserAttribute (note: this is best-effort)
 	// The actual attributes are stored internally, we log what we know
 	if question := ctx.GetUserAttribute("question"); question != nil {


johnlanni:
Fixed the compilation error. Changed from `wrapper.MarshalStr(v)` to `json.Marshal(v)` to properly convert `map[string]int64` to JSON string.

Sorry for not testing the build before submitting. The fix has been pushed.

johnlanni:
Fixed test error. Removed GetFilterState call as TestHost doesn't support this method. The test now only verifies that actions return correctly, matching the pattern in other test cases like TestBuiltinAttributesDefaultSource.

johnlanni:
Added token details to debug log output.

The `debugLogAiLog` function now includes:
- `reasoning_tokens`
- `cached_tokens`
- `input_token_details`
- `output_token_details`

These fields will be logged in debug mode when processing responses.

To verify the functionality manually, you can check the debug logs in the Gateway logs when making a request with token details in the response (e.g., OpenAI with prompt caching or reasoning models). The log will show:

```
[ai_log] attributes to be written: {"reasoning_tokens":25,"cached_tokens":80,"input_token_details":"{\"cached_tokens\":80}","output_token_details":"{\"reasoning_tokens\":25}",...}
```

The code flow:
1. `onHttpResponseBody` / `onHttpStreamingBody` extracts usage via `tokenusage.GetTokenUsage()`
2. Token details (input/output) are stored in context
3. `setAttributeBySource` is called for response body/streaming body
4. For builtin keys without value_source, `getBuiltinAttributeFallback` extracts details from context
5. `debugLogAiLog` logs all attributes including token details
6. `ctx.SetUserAttribute` stores the values for ai_log output

johnlanni:
Optimized debug logging to only run in debug/trace mode.

Changes:
- Added LogLevel constants (Trace, Debug, Info, Warn, Error, Critical)
- Added `shouldLogDebug()` helper function to check current log level via `proxywasm.CallForeignFunction("get_log_level", nil)`
- Modified `debugLogAiLog` to early return when not in debug/trace mode

Benefits:
- Avoids expensive JSON marshaling and string operations in production
- Reduces performance overhead when debug logging is not needed
- Only logs detailed attributes when explicitly enabled

The check is done at the beginning of `debugLogAiLog`, so all three call sites (onRequestBody, onHttpResponseBody, onHttpStreamingBody) automatically respect this optimization.

johnlanni:
Added length check for log level value to prevent panic.

Fix:
- Added `len(value) < 4` check before calling `binary.LittleEndian.Uint32(value)`
- This prevents potential panic if get_log_level returns a value with insufficient length

The change ensures:
```go
if len(value) < 4 {
    // Invalid log level value length
    return false
}
envoyLogLevel := binary.LittleEndian.Uint32(value[:4])
```

johnlanni:
Fixed duplicate variable declaration causing compilation error.

Problem:
- `debugLogAiLog` function had duplicate line: `userAttrs := make(map[string]interface{})`
- This caused 'variable redeclared' compilation error

Fix:
- Removed the duplicate declaration
- Now the function only declares `userAttrs` once at the beginning

This was introduced when adding the log level check to debugLogAiLog function.

PR #3576, opened 2026-03-09
fix(wasm): resolve stale route_name in wasm context after reroute

### LINKED ISSUE #3571: Wasm match_rule doesn't work after reroute triggered in a previous plugin in 2.2.0

- [X] I have searched the [issues](https://github.com/alibaba/higress/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description

Configure two WasmPlugins, A and B in Higress, executing in order in the filter chain.

A updates request header, pointing the request to another route. And B has a match_rule configured with the new route, but the match_rule isn't used when processing the request.

### Ⅱ. Describe what happened

Plugin B doesn't use the config in the match_rule associated with the new route to process the request.

Plugin B shall use the config in the match_rule associated with the new route to process the request.

1. Install Higress using the following command:
	```bash
	helm install higress -n higress-system higress.io/higress --create-namespace --render-subchart-notes --set global.local=true --set global.o11y.enabled=false
	```
2. Apply test configurations: [bug.yaml](https://github.com/user-attachments/files/25821203/bug.yaml)
3. Execute `curl http://localhost:18080/test` and check the `os.hostname` field in the response, whose value is expected to be `host-app`.
4. Execute `curl http://localhost:18080/test?userId=1`. The response is expected to be `{"hello":"world"}` which is replaced by `custom-response` plugin. But actually, the request reaches `foo-app` pod according to the `os.hostname` field in the response.

> It is recommended to provided Higress runtime logs and configurations for us to investigate your issue, especially for controller and gateway components.
> 
> Please checkout following documents on how to obtain these data.
> - https://higress.cn/docs/latest/ops/how-tos/view-logs/
> - https://higress.cn/docs/latest/ops/how-tos/view-configs/

Configurations of Higress 2.2.0:

- Pilot config: [pilot.json](https://github.com/user-attachments/files/25821201/pilot.json)
- Gateway config: [gateway.json](https://github.com/user-attachments/files/25821202/gateway.json)

The same config works as expected after downgrading Higress 2.1.11 with the following command:

```bash
helm upgrade higress -n higress-system higress.io/higress --create-namespace --render-subchart-notes --set global.local=true --set global.o11y.enabled=false --version 2.1.11
```

### Ⅵ. Environment:

- Higress version: 2.2.0
- OS: Windows 11 + Rancher Desktop 1.22.0 + K8s v1.28.15+k3s1
- Others: N/A

Jing-ze:

After upgrading Envoy from 1.27 to 1.36 (Higress 2.2.0), the `ROUTE_NAME` property resolution in the wasm context returns stale route names after `clearRouteCache()` is triggered by a preceding plugin (e.g., transformer with reroute). This causes subsequent plugins' `matchRule` evaluation to fail, as they see the old route name instead of the re-evaluated one.

**Root Cause:** In Envoy 1.36, `StreamInfoImpl::getRouteName()` delegates to `route_->routeName()`, but `clearRouteCache()` only clears `cached_route_` in ActiveStream — it does NOT update `streamInfo().route_`. The original code checked `info->getRouteName()` first (which reads the stale `streamInfo().route_`), so the second path via `filter_callbacks->route()` (which correctly triggers route re-evaluation) was never reached.

**Fix:** Swapped the resolution order in the `ROUTE_NAME` case of `findValue()` in `context.cc`:
1. Check `filter_callbacks->route()` **first** — this triggers `refreshCachedRoute()` when the route cache has been cleared, returning the correctly re-evaluated route.
2. Fall back to `info->getRouteName()` only when `filter_callbacks` is unavailable (e.g., during access log phase).

**Changes:**
- Updated `envoy/envoy` submodule to include the fix commit (`43287ff203`)
- Added e2e test case `WasmPluginsRerouteMatchRule` to reproduce and verify the fix

fixes #3571

An e2e conformance test has been added (`go-wasm-reroute-match-rule`). It sets up:
- A **transformer** plugin (priority 400) with matchRule on a default ingress, mapping query param `userId` to header `x-user-id` (triggering reroute)
- A **custom-response** plugin (priority 200) with matchRule on a target ingress (matched by `x-user-id: 1` header)
- Verifies that after reroute, the custom-response plugin correctly activates on the new route

run the e2e test:
```bash
make higress-wasmplugin-test PLUGIN_TYPE=GO TEST_SHORTNAME=WasmPluginsRerouteMatchRule

### DIFF
--- a/test/e2e/conformance/tests/go-wasm-reroute-match-rule.go
+++ b/test/e2e/conformance/tests/go-wasm-reroute-match-rule.go
@@ -1,0 +1,101 @@
+// Copyright (c) 2022 Alibaba Group Holding Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"testing"
+
+	"github.com/alibaba/higress/v2/test/e2e/conformance/utils/http"
+	"github.com/alibaba/higress/v2/test/e2e/conformance/utils/suite"
+)
+
+func init() {
+	Register(WasmPluginsRerouteMatchRule)
+}
+
+// WasmPluginsRerouteMatchRule tests that wasm plugin matchRules work correctly
+// after a reroute triggered by a previous plugin (issue #3571).
+//
+// Scenario:
+//   - Ingress "reroute-match-default": host reroute-match.com, path /, no header constraint
+//   - Ingress "reroute-match-target": host reroute-match.com, path /, header x-user-id: 1
+//   - Plugin A (transformer, priority 400): matchRule for "reroute-match-default",
+//     maps query param "userId" to header "x-user-id", causing reroute
+//   - Plugin B (custom-response, priority 200): matchRule for "reroute-match-target",
+//     returns custom response {"hello":"world"}
+//
+// When sending GET /?userId=1, plugin A adds x-user-id:1 header and triggers reroute.
+// Plugin B should see the NEW route "reroute-match-target" and return the custom response.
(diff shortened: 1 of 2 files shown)


Jing-ze:
Need to update the envoy binary.

CH3CHO:
--- FAIL: TestHigressConformanceTests/WasmPluginsRerouteMatchRule (300.03s)

image

johnlanni (approved):
LGTM

PR #3606, opened 2026-03-17
Feat/fix vertex chunk parsing instability

hongshi1:
## Summary

This PR hardens Vertex streaming response parsing in `ai-proxy`.

Previously, Vertex streaming handling parsed each incoming chunk directly by splitting on newlines and unmarshalling line-by-line. That was fragile when an upstream SSE `data:` event was split across multiple network chunks. It could also drop the final payload because the last chunk returned `[DONE]` immediately.

## Changes

- add buffered SSE data-line extraction helpers for streaming parsing
- make Vertex streaming response handling support split/incomplete chunks safely
- keep the final payload conversion on the last chunk, then append `data: [DONE]`
- add provider-level regression tests for:
  - split chunk reconstruction
  - final chunk payload preservation before `[DONE]`

## Verification

```bash
go test ./provider -run 'TestVertexStreamingResponseBody|TestVertexStreamingResponseBodyBuffersSplitChunks|TestVertexStreamingResponseBodyKeepsFinalPayloadBeforeDone'
go test -run TestVertex ./...


EndlessSeeker:
Hi, it looks like the CLA has not been signed yet. Could you please sign it so the PR can proceed? Thanks!

PR #3738, opened 2026-04-20
feat(ai-security-guard): add fallback JSON paths for response content extraction

JianweiWang:
## Summary

Add configurable fallback JSON paths (`responseContentFallbackJsonPaths` and `responseStreamContentFallbackJsonPaths`) to the `ai-security-guard` plugin, enabling response content extraction from non-OpenAI formats such as Anthropic Claude.

When the primary extraction path yields empty content, the plugin tries fallback paths in order. Paths equal to the primary path are automatically skipped. Setting either field to an empty array disables fallback explicitly.

### Key implementation details

- Default fallback paths cover both OpenAI and Claude response formats
- Effective fallback paths are pre-computed in the response header phase and cached in request context to avoid per-chunk recalculation
- Handles bare JSON, SSE data payloads, and multi-line data events
- Default variables are unexported; exposed via functions returning copies

### Changed files

- `plugins/wasm-go/extensions/ai-security-guard/config/config.go` — config struct & fallback path logic
- `plugins/wasm-go/extensions/ai-security-guard/lvwang/common/text/openai.go` — response content extraction with fallback
- `plugins/wasm-go/extensions/ai-security-guard/lvwang/common/text/openai_test.go` — unit tests for fallback extraction (new)
- `plugins/wasm-go/extensions/ai-security-guard/main_test.go` — integration tests
- `plugins/wasm-go/extensions/ai-security-guard/README.md` / `README_EN.md` — documentation updates


rinfx (approved):
LGTM

PR #3760, opened 2026-04-27
fix: preserve int64 precision in MCP tool argument parsing

Thomas-Eliot:
## Problem

`json.Unmarshal` decodes all JSON numbers into `float64`, which only has ~15-16 digits of decimal precision (2^53). This silently corrupts int64 values larger than 2^53.

For example: `1234567890123456789` → `1234567890123456768` (diff: 21)

This affects MCP tool arguments of type `integer` when the value exceeds 15-16 digits (e.g. 19-digit order IDs, snowflake IDs).

## Fix

Use `json.NewDecoder` with `UseNumber()` instead of `json.Unmarshal`, so numbers are parsed as `json.Number` (string) and then converted to `int64`/`float64` without intermediate precision loss.

## Changes

- Use `json.NewDecoder` + `UseNumber()` instead of `json.Unmarshal`
- Add `json.Number` case to `integer` and `number` type switches
- Handle `json.Number` in default branch (convert to string)

## Verification

```
Before: 1234567890123456789 → 1234567890123456768  ❌
After:  1234567890123456789 → 1234567890123456789  ✅
```


EndlessSeeker:
Good fix. One suggestion: instead of converting json.Number → int64/float64, consider keeping json.Number as-is in the arguments map. It's a string internally (no precision loss for any size) and marshals as a raw number (no quotes). Would simplify the code and avoid the int64 upper bound limit too.

Thomas-Eliot:
Thanks for the suggestion! I considered keeping `json.Number` as-is, but decided to keep the current approach for a couple of reasons:

1. **Type validation**: Converting to `int64`/`float64` validates that the value actually matches the declared schema type. For example, if a field is declared as `"integer"`, passing `"3.14"` will be caught. Keeping `json.Number` would silently accept any numeric string regardless of the declared type.

2. **Clear downstream contract**: With native Go types in the arguments map, downstream consumers (`convertArgToString`, `sjson.SetBytes`, body serialization) can work with `int64`/`float64` directly. Keeping `json.Number` would shift complexity to every consumption point — they'd all need to handle `json.Number` as a possible type.

The current approach converts correctly without precision loss, which is the core goal of this PR. The added `json.Number` cases in the type switch are the minimal change needed to fix the issue.

PR #3779, opened 2026-04-30
fix: fix non-json logs with log_as_json enabled

CH3CHO:

Fix the bug that even starting with `--log_as_json` argument, there are still some logs of controller written in plain text format.

Before:
image

image

After:

image

image

### DIFF
--- a/cmd/higress/main.go
+++ b/cmd/higress/main.go
@@ -18,7 +18,7 @@ import (
 	"fmt"
 	"os"
 
-	"istio.io/pkg/log"
+	"istio.io/istio/pkg/log"
 
 	"github.com/alibaba/higress/v2/pkg/cmd"
 )
--- a/pkg/cert/certmgr.go
+++ b/pkg/cert/certmgr.go
@@ -17,14 +17,12 @@ package cert
 import (
 	"context"
 	"fmt"
-	"os"
 	"reflect"
 	"sync"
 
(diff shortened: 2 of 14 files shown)


EndlessSeeker:
LGTM, two things to confirm:
1. zap.L() initialization order — After switching to zap.L() in certmgr.go, this relies on the istio log package having called zap.ReplaceGlobals before InitCertMgr runs. If not, certmagic logs will be silently dropped (no-op logger).
2. Log level change — The old logger was hardcoded to DebugLevel. Now it follows the global config. If the global level is Info, certmagic debug logs will be filtered out. Is this intentional?

CH3CHO:
> 1. zap.L() initialization order — After switching to zap.L() in certmgr.go, this relies on the istio log package having called zap.ReplaceGlobals before InitCertMgr runs. If not, certmagic logs will be silently dropped (no-op logger).

Based on the debug result, `zap.ReplaceGlobals` is called by `istio.io/istio/pkg/log` package before `InitCertMgr` runs.

image

image

> 2. Log level change — The old logger was hardcoded to DebugLevel. Now it follows the global config. If the global level is Info, certmagic debug logs will be filtered out. Is this intentional?

I think as an online service, we shall never use `debug` by default. Changing it back to `info` and allowing user change it via command line args should be an improvement.

EndlessSeeker (approved):

PR #3792, opened 2026-05-07
fix(ai-data-masking): support multi-word keywords with spaces in deny_words

JianweiWang:
## Summary

`DenyWord::check()` uses Jieba tokenizer which splits text by spaces, making it impossible to match keywords containing spaces (e.g. "hello world").

This PR fixes the issue by:
- Adding `phrase_words` field to `DenyWord` for space-containing keywords
- Using substring matching with boundary checking for phrase keywords
- Adding `allow_end_boundary` param for streaming mode support
- Threading `allow_end_boundary` through AiDataMasking call chain
- Updating streaming window size to cover longest phrase keyword
- Adding bug condition exploration and preservation property tests
- Updating README/README_EN to reflect multi-word keyword support
- Fixing `.prebuild` to skip gcc-multilib on arm64

## Test plan

- [x] Added property tests for bug condition exploration and preservation
- [ ] Verify multi-word keyword matching works in both streaming and non-streaming modes
- [ ] Verify boundary checking correctly identifies word boundaries
- [ ] Test with various space-containing keywords in deny_words configuration


johnlanni:
please fix the ci

PR #3788, opened 2026-05-05
fix(ai-proxy): preserve Bedrock Claude reasoning blocks

Betula-L:
## Ⅰ. Describe what this PR did

This PR fixes Bedrock Claude reasoning conversion in the `ai-proxy` protocol bridge.

Previously, Bedrock Claude `reasoningContent` could be appended to normal chat text as `<think>...</think>`. That made hidden reasoning user-visible and lost Anthropic Messages block structure. This change keeps reasoning in structured protocol fields and preserves native Anthropic block semantics when `/v1/messages` conversion is active.

Main changes:

- Emits Bedrock reasoning as structured OpenAI-compatible reasoning fields instead of injecting `<think>` tags into visible `content`.
- Converts Bedrock Claude reasoning back to native Anthropic `/v1/messages` blocks, including `thinking`, `redacted_thinking`, `thinking_delta`, `signature_delta`, and block stop events.
- Preserves Bedrock streaming `contentBlockIndex` and `contentBlockStop` in the same converted SSE chunk, avoiding cross-event side-channel queues that can reorder or duplicate Anthropic stream blocks.
- Suppresses orphan Anthropic `content_block_stop` events when Bedrock sends a stop for a block that was not opened by the bridge.
- Preserves ordered Anthropic follow-up content blocks for Bedrock Converse requests, including `thinking`, `redacted_thinking`, `text`, `tool_use`, and `tool_result`.
- Preserves exact Claude `thinking.budget_tokens` when routing Claude-native requests to Bedrock.
- Preserves Claude `output_config.format` and maps it to Bedrock Converse `outputConfig.textFormat`.
- Maps Claude adaptive `output_config.effort` to Bedrock Converse `additionalModelRequestFields.thinking.effort` for Bedrock-supported effort values (`low`, `medium`, or `high`).
- Downgrades forced tool choices to `auto` while thinking is enabled, matching Anthropic/Bedrock tool-use restrictions for thinking.
- Keeps empty `tool_result` text and empty `thinking` plus `signature` valid on Bedrock JSON by preserving required union fields.
- Preserves direct Claude provider thinking signatures, redacted thinking, and bridged tool results when `/v1/messages` conversion routes through Claude.
- Strips Claude bridge-only fields before forwarding converted requests to non-Bedrock and non-Claude upstream providers.

## Ⅱ. Does this pull request fix one issue?

No linked issue.

## Ⅲ. Why don't you add test cases (unit test/integration test)?

Regression tests were added for:

[...]

PR #3799, opened 2026-05-08
fix(ai-proxy): preserve empty Claude tool inputs

Betula-L:
## Summary

This is a follow-up to #3788. That change preserves Claude native content blocks through an internal `claude_content_blocks` bridge before converting them to Bedrock Converse blocks.

Claude Messages tool use blocks can carry an explicit empty input object for no-argument tools:

```json
{"type":"tool_use","id":"toolu_1","name":"list_items","input":{}}
```

When such a block is preserved through the internal bridge, `encoding/json` currently omits the empty map because the field is tagged with `omitempty`. After the internal JSON round-trip, the Bedrock request builder receives a nil input and serializes it as `"input": null`, which Bedrock rejects for `toolUse.input`.

This change keeps the explicit-empty-object distinction by using a pointer map for Claude native `tool_use.input`, matching the existing `claudeTextGenContent.Input` shape. The Bedrock request builder also normalizes missing tool input to `{}` as a defensive fallback so it never emits `toolUse.input:null`.

## Tests

```bash
go test ./provider -run 'TestClaudeToOpenAIConverter_ConvertClaudeRequestToOpenAI/preserve_empty_tool_use_input|TestBedrockRequestPreservesClaudeNoArgToolUseInput' -count=1
go test ./provider -run 'TestBedrockRequest|TestClaudeToOpenAIConverter_ConvertClaudeRequestToOpenAI|TestClaudeProvider' -count=1
go test ./provider -count=1
```

### DIFF
--- a/plugins/wasm-go/extensions/ai-proxy/provider/bedrock.go
+++ b/plugins/wasm-go/extensions/ai-proxy/provider/bedrock.go
@@ -1506,7 +1506,7 @@ func claudeContentBlocksToBedrockContents(blocks []claudeChatMessageContent) []b
 			}
 		case "tool_use":
 			result = append(result, bedrockMessageContent{ToolUse: &toolUseBlock{
-				Input:     block.Input,
+				Input:     claudeToolUseInput(block.Input),
 				Name:      block.Name,
 				ToolUseId: block.Id,
 			}})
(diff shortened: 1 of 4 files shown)


Betula-L:
@wydream   Sorry for introducing this bug. Could you please take a look at this PR when you have a chance? It fixes the regression and adds a focused test to cover the empty tool-use input case. Thanks.

wydream:
> @wydream Sorry for introducing this bug. Could you please take a look at this PR when you have a chance? It fixes the regression and adds a focused test to cover the empty tool-use input case. Thanks.

No worries, thanks for the quick fix. I reviewed the PR and verified the focused regression test plus the ai-proxy test suite. The change looks good to me.

wydream (approved):
LGTM

PR #3807, opened 2026-05-10
fix(key-auth): fix credential cache key

huchunnuan:
                                                                                                                        
  Fix performance issue and multiple bugs in the key_auth WASM plugin:                                                  
                                                                                                                        
  1. **Optimize consumer credential matching performance** — The original code iterated through all consumers           
  sequentially, calling `extractCredential` (header/query lookup) for each consumer. When there are many consumers, this
   causes significant performance degradation. Optimized with a fast/slow path strategy:
     - **Fast path**: When consumers use global keys (no per-consumer `keys`/`in_header`/`in_query` overrides), use O(1)
   `credential_to_name` map lookup instead of iterating all consumers                                                   
     - **Slow path**: Only fall back to per-consumer iteration for consumers with per-consumer key configuration, with 
  credential extraction results cached to avoid duplicate header/query lookups                                          
     - Extracted `checkAllowSet` and `hasPerConsumerKeyConfig` helpers to reduce code duplication
                                                                                                                        
  2. **Fix credentialCache cache key bug** — In the slow path, the cache key only used the header name (e.g., 
  `"apikey"`), without distinguishing the extraction source (`in_header`/`in_query`). When two consumers shared the same
   key name but had different extraction settings, the cache would return the wrong credential. Fixed by composing the 
  cache key as `"key:h"`, `"key:q"`, or `"key:hq"`.                                                                     
                                                            
  3. **Mask API key in debug logs** — `LOG_DEBUG` statements were printing raw credential values, risking key leakage in
   production logs. Added `maskCredential()` helper that shows only the first 3 characters (e.g., `"abc***"`).          
   
  4. **Fix realm prefix in `deniedUnauthorizedConsumer`** — Was using `"Basic realm="` (copy-paste from basic_auth      
  plugin), changed to `"Key realm="` to match `deniedInvalidCredentials`.                                               
   
  5. **Add missing `LOG_WARN` for realm parse failure** — The realm field parsing returned `false` without logging,     
  making it hard to diagnose configuration errors.                                                                      
   
  6. **Add `const` qualifier to `extractCredential`** — The method doesn't modify any member state.                     
                                                                                                                        
  No specific issue linked. Found during code review.       
                                                                                                                        
  The credentialCache cache key fix needs a test case where two consumers share the same key name but use different 
  extraction sources (one from header, one from query). Existing tests don't cover this scenario —                      
  `ConsumerMultiCredentials` uses different key names per consumer. Will add a dedicated test case in a follow-up PR.
                                                                                                                        
  The other changes (log masking, realm prefix, LOG_WARN, const) are non-functional or logging-only and don't require 
  new tests.                                                                                                            
   
  1. Configure key_auth with two consumers sharing the same key name but different `in_header`/`in_query` settings — 
  verify correct consumer is matched                                                                                    
  2. Trigger invalid credential and check debug logs show masked values (`abc***` not full key)
  3. Trigger unauthorized consumer rejection and verify `WWW-Authenticate: Key realm=...` header                        
  4. Configure invalid `realm` field and verify LOG_WARN appears                                                        
  5. Benchmark: configure many consumers with global keys, verify fast path O(1) lookup is used instead of per-consumer 
  iteration                                                                                                             
                                                                                                                        
  - The `const` on `extractCredential` depends on `getRequestHeader()` from proxy-wasm SDK being const-qualified. If the
   SDK version used doesn't support this, the const can be removed without functional impact.                           
  - Mutation analysis showed the credentialCache fix lacks test coverage — the existing `ConsumerMultiCredentials` test 
  uses different key names per consumer, so reverting the cache key change wouldn't be caught by tests.                 
  - The fast/slow path optimization extracts `perConsumerKeyConsumers` into a vector of pointers. The `perConsumerNames`
   set is built for O(1) membership check in the fast path — this trades a small set construction cost for avoiding     
  repeated linear scans.                                    
                                                                                                                        
  - [ ] **For new standalone features** (e.g., new wasm plugin or golang-filter plugin):
    - [ ] I have created a `design/` directory in the plugin folder                                                     
    - [ ] I have added the design document to the `design/` directory                                                   
    - [ ] I have included the AI Coding summary below                                                                   
                                                                                                                        
  1. "分析下 higress/plugins/wasm-cpp/extensions/key_auth/plugin.cc 还有需要优化的吗" — Asked for code analysis of 
  plugin.cc                                                                                                             
  2. Compared analysis with another AI tool's findings, identified overlapping and unique findings
  3. "开始修复" — Instructed to fix identified issues                                                                   
  4. "再检查下所有改动是否有问题" — Asked for final review of all changes                                               
  5. "是否可以变异一下看看是否有问题" — Asked for mutation testing analysis                                             
  6. "提交push代码" — Asked to commit and push                                                                          
                                                                                                                        
  ### AI Coding Summary                                                                                                 
                                                                                                                        
  **Key decisions:**                                        
  - Introduced fast/slow path strategy to avoid O(n) per-consumer iteration when consumers use global key configuration 
  — critical for deployments with many consumers                                                                        
  - Used `absl::StrCat` for cache key composition instead of string concatenation for consistency with existing codebase
   style                                                                                                                
  - Chose 3-character prefix for credential masking as a balance between debugging utility and security
  - Used `perConsumerNames` unordered_set for O(1) consumer name lookup instead of repeated linear scans                
                                                                                                                        
  **Limitations identified via mutation testing:**                                                                      
  - Credential masking (pure logging) is not testable with current test infrastructure                                  
  - The credentialCache cache key fix needs a dedicated test case with shared key names across consumers                
  - Realm prefix change would not be caught by existing tests (no assertion on WWW-Authenticate header value)

### DIFF
--- a/.agents/skills/issue-spec-apply/SKILL.md
+++ b/.agents/skills/issue-spec-apply/SKILL.md
@@ -1,0 +1,35 @@
+---
+name: issue-spec-apply
+description: Implement PROCESS comments for an issue-spec change and keep PR traceability synchronized.
+license: MIT
+compatibility: Requires issue-spec CLI.
+metadata:
+  author: issue-spec
+  version: "1.0"
+  generatedBy: "issue-spec"
+---
+
+# Issue Spec Apply
+
+Use when the user asks for /issue-spec:apply, issue-spec apply, or implementing PROCESS/TASK scopes from an issue-spec change.
+
+## Steps
+
+1. Read proposal/design/implement issue context and list typed comments with issue-spec comment list --json.
+2. Confirm issue-spec auth status --json includes the expected GitHub backend. Local gh-authenticated sessions can use the native gh backend; keep ISSUE_SPEC_TOKEN="$(gh auth token)" only as an older-version or forced-rest compatibility path.
+3. Create or update PROCESS comments with owner agent, scope, dependencies, write ownership, and status.
+4. Split non-trivial work into independent worker PROCESS nodes when file/module ownership does not overlap; execute independent workers in parallel when available.
+5. Add dedicated review PROCESS nodes for non-trivial changes. Review PROCESS nodes should own review scopes such as CLI/API behavior, workflow docs, tests, compatibility, or security-sensitive surfaces.
+6. Link each PROCESS to its TASK comments with issue-spec link.
+7. Implement the code changes for one PROCESS scope at a time, or integrate completed worker outputs by dependency order.
+8. Link every worker and review PROCESS to the PR with issue-spec pr link-process.
+9. Add PR rationale comments on key changed lines with issue-spec pr rationale, each linked to a SPEC comment.
+10. Mark PROCESS comments done only after implementation/review work and focused verification evidence exist.
+
+## Coordinator DAG Execution
+
+1. Build the ready set from PROCESS nodes whose dependencies are done.
+2. Keep immediate blocking work local when the next step depends on it.
+3. Spawn or assign independent worker agents only when their write ownership is disjoint.
+4. Spawn or assign independent review agents only when their review scopes are disjoint.
+5. Integrate completed outputs by dependency order and update PROCESS evidence before marking done.
--- a/.agents/skills/issue-spec-archive/SKILL.md
+++ b/.agents/skills/issue-spec-archive/SKILL.md
@@ -1,0 +1,24 @@
+---
+name: issue-spec-archive
+description: Create the post-merge durable spec archive PR for an issue-spec change.
+license: MIT
+compatibility: Requires issue-spec CLI.
+metadata:
+  author: issue-spec
+  version: "1.0"
+  generatedBy: "issue-spec"
+---
+
+# Issue Spec Archive
+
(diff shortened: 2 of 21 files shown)


CH3CHO:
请 @EndlessSeeker @johnlanni review 一下

---
Please @EndlessSeeker @johnlanni review it

johnlanni:
Can you provide a specific example that demonstrates the issue?

huchunnuan:
> Can you provide a specific example that demonstrates the issue?

In my scenario, the model request failure rate approaches 30% when the number of consumers exceeds 200, primarily due to calling extractCredential for each consumer

EndlessSeeker:
LGTM. One request: please add a test case for the cache key fix — two consumers sharing the same key name but with different in_header/in_query settings.

huchunnuan:
> LGTM. One request: please add a test case for the cache key fix — two consumers sharing the same key name but with different in_header/in_query settings.

have already added test case in plugin_test.cc

EndlessSeeker:
> > LGTM. One request: please add a test case for the cache key fix — two consumers sharing the same key name but with different in_header/in_query settings.
> 
> have already added test case in plugin_test.cc

plz sign the CLA

PR #3838, opened 2026-05-15
feat(jwt-auth): support remote JWKS

Betula-L:
## Summary
- Add `remote_jwks` support to wasm-go `jwt-auth` consumers as an alternative to inline `jwks`.
- Fetch, validate, cache, and refresh remote JWKS through a Higress service reference (`service_name`, `service_host`, `service_port`, `path`).
- Validate remote JWKS service configuration, including bare `service_host` plus separate `service_port`.
- Cache parsed JWKS and apply each consumer's `jwks_cache_duration` at read time.
- Serve the existing cached JWKS while a TTL refresh for the same remote service reference is in flight.
- Bound remote JWKS fetch timeout, cache duration, and response size.
- Guard stale fetch callbacks so an old response cannot clear newer fetch/cache state.
- Fail closed on fetch failures, invalid JWKS, throttled cold-cache refreshes, and unresolved key IDs. Tokens without `kid` are accepted only when the fetched JWKS contains exactly one key.
- Respect route/domain `allow` lists before triggering remote JWKS fetches.
- Apply `keep_token: false` and `claims_to_headers` side effects only after the selected authorized consumer verifies successfully.
- Document remote JWKS fields in the existing JWT plugin docs, explicitly marking them as wasm-go-only.

## Notes
- Remote JWKS responses larger than 64 KiB are rejected before calling `GetHttpCallResponseBody`.
- Cold-cache in-flight fetches and recent failures are denied locally to avoid request stampedes.
- `keep_token: false` removal is deferred until authentication succeeds; failed candidates no longer remove tokens before later consumers are checked.
- Concurrent requests during a cold remote JWKS fetch are fail-closed instead of coalesced because proxy-wasm HTTP call callbacks are bound to one stream context.

## Tests
- `go test ./...` in `plugins/wasm-go/extensions/jwt-auth`
- `git diff --check`

### DIFF
--- a/plugins/wasm-go/extensions/jwt-auth/config/config.go
+++ b/plugins/wasm-go/extensions/jwt-auth/config/config.go
@@ -14,6 +14,8 @@
 
 package config
 
+import "github.com/go-jose/go-jose/v3"
+
 var (
 	// DefaultClaimToHeaderOverride 是 claim_to_override 中 override 字段的默认值
 	DefaultClaimToHeaderOverride = true
@@ -23,6 +25,12 @@ var (
 
 	// DefaultKeepToken 是 KeepToken 的默认值
 	DefaultKeepToken = true
+
+	// DefaultJWKsCacheDuration is the default remote JWKS cache duration in seconds.
+	DefaultJWKsCacheDuration = int64(600)
+
+	// DefaultJWKsFetchTimeout is the default remote JWKS fetch timeout in milliseconds.
+	DefaultJWKsFetchTimeout = int64(1500)
 
 	// DefaultFromHeader 是 from_header 的默认值
 	DefaultFromHeader = []FromHeader{{
@@ -37,6 +45,9 @@ var (
 	DefaultFromCookies = []string{}
 )
 
+// RemoteJWKsMinRefreshIntervalSeconds is the shared lower bound for remote JWKS cache TTL and retry backoff.
+const RemoteJWKsMinRefreshIntervalSeconds = int64(30)
+
 // JWTAuthConfig defines the struct of the global config of higress wasm plugin jwt-auth.
 // https://higress.io/zh-cn/docs/plugins/jwt-auth
 type JWTAuthConfig struct {
@@ -44,6 +55,9 @@ type JWTAuthConfig struct {
 	//
 	// Consumers 配置服务的调用者，用于对请求进行认证
 	Consumers []*Consumer `json:"consumers"`
+
+	// RuleSet records whether at least one domain or route rule is configured.
+	RuleSet bool `json:"-"`
 
 	// 全局配置
 	//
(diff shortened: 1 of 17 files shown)


CH3CHO:
Please update README.md files accordingly.

Betula-L:
@CH3CHO Updated.

Betula-L:
@CH3CHO How can I make progress on this feature?

CH3CHO:
> @CH3CHO How can I make progress on this feature?

I'm too busy to spare some time to look into this PR recently. But I will try to look into it during the weekend. If possible, could @EndlessSeeker @johnlanni review this PR?

plugins/wasm-go/extensions/jwt-auth/config/config.go
```diff
@@ -68,6 +74,17 @@ type Consumer struct {
 	// https://www.rfc-editor.org/rfc/rfc7517
 	JWKs string `json:"jwks"`
 
+	// JWKsURI specifies the HTTPS remote JWKS URL used to refresh signing keys.
+	// The gateway must be able to reach the host through an FQDN outbound cluster.
+	JWKsURI string `json:"jwks_uri"`
```
CH3CHO: I don't think letting user to configure a raw URI is a good idea. It results that we need to generating service entries and other related configuration in the controller.

Since we have already supported configuring different services using `McpBridge`, I would suggest using the same way as other plugins are doing right now, using `service_name`, `service_host` and `service_port` in the plugin config as the reference to a service configured or discovered from `McpBridge`

Betula-L: @CH3CHO Updated.

plugins/wasm-go/extensions/jwt-auth/handler/jwks_cache.go
```diff
@@ -0,0 +1,317 @@
...
+		if ok && remoteJWKsFetchInFlight(consumer, now) {
+			return cached.keys, nil
+		}
+		if !remoteJWKsFetchAllowed(consumer, now) {
+			return nil, errRemoteJWKsRefreshThrottled
+		}
+		return nil, errRemoteJWKsCacheMiss
+	}
+
+	return parseJWKs(raw)
```
CH3CHO, line 118: It would be better to have the parsed local JWKs cached as well.

CH3CHO (commented):
Generally LGTM.

CH3CHO (approved):

Betula-L:
I split the remote JWKS fetch and retry state out from the JWT verifier so the two concerns no longer share mutable state.

CH3CHO:
Hi @Betula-L, a follow-up thing related to this PR.

Although the `jwt-auth` plugin referred by Higress Console is the C++ version, the remote JWKS support you added allows user to use the Golang version in specific scenarios. However, the README files you modified are not in the `wasm-go` folder, but in the `wasm-cpp` folder, resulting a mismatch between source code and documation.

Would you like to take some time to migration documentation changes to the right files?

PR #3851, opened 2026-05-20
fix(ai-statistics): skip empty streaming model values

JianweiWang:
## What changed

- Treat missing JSON paths, explicit JSON `null`, and empty string `""` as empty values for `first` and `replace` extraction from `response_streaming_body`.
- Preserve valid non-string values such as `0` and `false`.
- Document the behavior difference in the ai-statistics README files.
- Add host-level streaming tests that verify the final `ai_log` output for empty/null chunks and missing model paths.

## Why

Some streaming providers can emit early chunks with an empty `model` field before later chunks carry the actual model name. Previously, `first` could lock onto `""`, and `replace` could be cleared by a trailing empty value.

## Validation

- `go test -run TestOnHttpStreamingBody ./...` in `plugins/wasm-go/extensions/ai-statistics`
- `go test ./...` in `plugins/wasm-go/extensions/ai-statistics`
- `git diff --check`

### DIFF
--- a/plugins/wasm-go/extensions/ai-statistics/README.md
+++ b/plugins/wasm-go/extensions/ai-statistics/README.md
@@ -62,6 +62,8 @@ Attribute 配置说明:
 - `replace`：多个 chunk 中取最后一个有效 chunk 的值
 - `append`：拼接多个有效 chunk 中的值，可用于获取回答内容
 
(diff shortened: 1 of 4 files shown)


johnlanni (approved):
LGTM

PR #3860, opened 2026-05-22
fix(ai-proxy): natively route Anthropic Messages to vertex :rawPredict

Jing-ze:
## Background

Anthropic [officially supports Claude Code on Google Vertex AI](https://code.claude.com/docs/en/google-vertex-ai) (alongside Bedrock and Microsoft Foundry). Claude Code talks to Vertex AI's Anthropic-native partner endpoint:

```
POST https://{LOCATION}-aiplatform.googleapis.com/v1/projects/{PROJECT}/locations/{LOCATION}/publishers/anthropic/models/{MODEL}:rawPredict
POST .../publishers/anthropic/models/{MODEL}:streamRawPredict
```

This endpoint accepts the standard Anthropic [Messages API](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/partner-models/claude/use-claude) body verbatim — required fields are `anthropic_version: vertex-2023-10-16`, `messages`, `max_tokens`; the model is taken from the URL, not the body.

Users want to put Higress in front of Vertex so Claude Code (and any other Anthropic SDK client) can be routed through ai-proxy.

## Problem

The vertex provider did **not** register `ApiNameAnthropicMessages` in its `DefaultCapabilities`. As a result, every `/v1/messages` request fell through the generic main.go path that triggers `ConvertClaudeRequestToOpenAI`, then either:

- (mode A, \`vertexOpenAICompatible: true\`) was forwarded to vertex's OpenAI-compatible Gemini endpoint, where vertex internally re-translated back to Anthropic for Claude models — a lossy Anthropic→OpenAI→Anthropic round-trip; or
- (mode B, default) was re-converted Claude→Anthropic by \`buildClaudeTextGenRequest\` and sent to \`:rawPredict\`, but with the same lossy intermediate OpenAI representation.

The double conversion silently dropped Anthropic-specific fields:

| Field | Why it matters | Symptom when dropped |
|---|---|---|
| `tools[i].type` (e.g. \`web_search_20250305\`) | Server-tool discriminator carries only \`type\`, no \`name\` | Upstream rejected with \`tools.0.custom.name: String should have at least 1 character\` |
| `tools[i].cache_control` | Prompt-caching breakpoints on tool definitions | Silent prompt-cache miss → cost / latency regression |
| `thinking` block | Claude extended thinking | Silently disabled |

The most visible failure was the misleading `tools.0.custom.name` 400 hitting Claude Code users whenever a server tool was in the request (Claude Code enables `web_search` by default).

## Fix

Register `ApiNameAnthropicMessages` for the vertex provider and forward `/v1/messages` straight to the Anthropic-native partner endpoint. All the helpers (`vertexPathAnthropicTemplate`, `vertexAnthropicVersion`, `getAhthropicRequestPath`) were already in place; this PR wires them up properly.

**Request path** (`onAnthropicMessagesRequestBody`):
- Apply `modelMapping` to translate client model names → vertex fully-qualified names (e.g. `claude-sonnet-4` → `claude-sonnet-4@20250514`)
- Construct `publishers/anthropic/models/{model}:rawPredict` / `:streamRawPredict`
- Strip `model` from the body (vertex requires it in the URL only)
- Inject `anthropic_version: vertex-2023-10-16`
- **No protocol conversion** — body forwarded verbatim, so all Anthropic-specific fields survive

**Response path**: vertex's `:rawPredict` and `:streamRawPredict` already emit standard Anthropic JSON / SSE, so `TransformResponseBody` and `OnStreamingResponseBody` return the body byte-for-byte when `apiName == ApiNameAnthropicMessages` (the same passthrough pattern used by the bedrock provider).

**Config interaction**: `vertexOpenAICompatible: true` now only affects `/v1/chat/completions`. `/v1/messages` always takes the native path, because vertex's OpenAI-compatible endpoint is designed for Gemini and adds no value for Claude requests (it would just cost a round-trip).

## Behavior matrix after fix

| Client API | `vertexOpenAICompatible` | Routing |
|---|---|---|
| `/v1/messages` | `false` / unset | vertex Anthropic native (`:rawPredict`), body passthrough |
| `/v1/messages` | `true` | **same** — config doesn't divert Anthropic API |
| `/v1/chat/completions` | `false` | unchanged (existing Claude → Anthropic transform via `:rawPredict` for Claude models, vertex native chat for Gemini) |
| `/v1/chat/completions` | `true` | unchanged (vertex OpenAI-compatible endpoint) |

## Out of scope

- vertex platform-level limits on Anthropic server tools (vertex doesn't support `bash`/`text_editor`/`code_execution`; `web_search` requires vertex's own non-versioned format + org policy + Claude 4.0+). After this fix, errors come from vertex with accurate diagnostics instead of being masked by the misleading `custom.name` synthesized by ai-proxy. Users hitting `web_search` issues should run `/config` in Claude Code to disable it, or enable vertex's web search integration explicitly.
- The OpenAI-compatible chat-completions path's residual lossy conversion for Claude models (separate concern; OpenAI clients on Claude are expected to lose Anthropic-only fields).

## Migration / config

Clients pointing `/v1/messages` at a Higress with this fix will start hitting vertex's Anthropic-native endpoint automatically. Users must configure \`modelMapping\` so Claude Code's short model names are translated to vertex's fully-qualified ones:

\`\`\`yaml
provider:
  type: vertex
  vertexProjectId: my-project
  vertexRegion: us-east5
  vertexAuthKey: |
    { ... service account json ... }
  vertexAuthServiceName: vertex-auth
  modelMapping:
    claude-sonnet-4:     claude-sonnet-4@20250514
    claude-sonnet-4-5:   claude-sonnet-4-5@20250929
    claude-opus-4:       claude-opus-4@20250514
    claude-opus-4-1:     claude-opus-4-1@20250805
    claude-haiku-4-5:    claude-haiku-4-5@20250930
\`\`\`

## Test plan

- [x] New unit tests in `provider/vertex_test.go` under `TestVertexAnthropicPassthrough_*`:
  - builtin tool `type` preserved (regression test for the `tools.0.custom.name` bug)
  - stream vs non-stream path construction (`:rawPredict` vs `:streamRawPredict`)
  - model mapping applied / unconfigured pass-through
  - custom tool `cache_control` + `thinking` block survive
  - `vertexOpenAICompatible: true` doesn't divert `/v1/messages`
  - `TransformRequestBodyHeaders` dispatches `ApiNameAnthropicMessages` to the new handler
  - non-stream and SSE response bodies returned byte-for-byte
- [x] Full `go test ./provider/` passes — no regression in existing vertex / claude tests
- [x] `go build ./...` clean in the ai-proxy module


rinfx (approved):
LGTM

PR #3878, opened 2026-05-25
test(traffic-tag): cover operator matrix, weight CDF boundaries, and parser reject paths

Jing-ze:
## What changed

Adds `plugins/wasm-go/extensions/traffic-tag/main_extra_test.go` to lift unit-test coverage of the traffic-tag wasm plugin from **65.8% → 93.0%**.

Single new file, no edits to production code or existing tests — mirrors the stacking style introduced in d9255536 for hmac-auth-apisix.

### Coverage delta

| Function | Before | After |
|---|---|---|
| `validate` | 62.5% | **100.0%** |
| `parseWeightConfig` | 69.6% | **100.0%** |
| `compileRegex` | 85.7% | **100.0%** |
| `setDefaultTag` | 66.7% | **100.0%** |
| `getFullRequestURL` | 0% | **100.0%** |
| `getQueryParameter` | 0% | 85.7% |
| `parseCookie` | partial | **100.0%** |
| `onWeightRequestHeaders` | 85.7% | **100.0%** |
| `matchCondition` | 50.7% | 85.9% |
| `getConditionValue` | 50.0% | 81.2% |
| `addTagHeader` | 50.0% | 75.0% |
| **total** | **65.8%** | **93.0%** |

## Why

Traffic-tag is on the WASM → native golang-filter migration list (P2, simple SubFilter batch). The migration depends on a wasm behavioral baseline to assert equivalence — at 65.8% several core branches were zero/half covered (parameter-type path entirely 0%, weight CDF gap-no-match 0%, percentage operator branch 0%, all reject paths absent). 93% gives the migration something to port and diff against.

## What's covered

- **Parser reject paths**: empty fields, invalid operator/logic/type, percentage out-of-range / multi-value / non-integer, weight negative / >100 / cumulative sum >100, invalid regex, missing header name.
- **Operator state machine**: `equal` / `not_equal` / `prefix` / `regex` / `in` / `not_in` / `percentage` × `and` / `or` logic, with both match and non-match branches.
- **conditionType dispatch**: header / cookie / parameter, including `getFullRequestURL` + `getQueryParameter` walks via `:path=/x?...`.
- **Weight CDF**: table-driven `onWeightRequestHeaders` with `randomNum=0/29/30/79/80/99/100/130/180` covering the gap-no-match branch (cumulative<100 with bucket in the gap) and the mod=0 wrap case.
- **Default-tag fallback**: applied when all conditions miss, skipped when key or value is empty (early returns).
- **Idempotency**: `addTagHeader` skips when the target header already exists on the inbound request.

## What's not covered, and why

- `main()` — wasm entrypoint, no test surface.
- `validate`'s default invalid-type / invalid-operator branches — runtime unreachable; `validate` itself rejects at parse time before those defaults can fire.
- `proxywasm.AddHttpRequestHeader` error path — framework does not expose a failure hook.
- `url.Parse` error path inside `getQueryParameter` — practically unreachable for any path Envoy hands the filter.

## Validation

```bash
cd plugins/wasm-go/extensions/traffic-tag
go test -race -timeout 180s ./...                           # PASS in ~1.3s
go test -coverprofile=/tmp/tt.out ./...                     # coverage: 93.0%
go tool cover -func=/tmp/tt.out | tail -1                   # total 93.0%
go vet ./...                                                # clean
```

### DIFF
--- a/plugins/wasm-go/extensions/traffic-tag/main_extra_test.go
+++ b/plugins/wasm-go/extensions/traffic-tag/main_extra_test.go
@@ -1,0 +1,964 @@
+// Copyright (c) 2025 Alibaba Group Holding Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+
+package main
+
+import (
+	"crypto/sha256"
+	"encoding/binary"
+	"encoding/json"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/higress-group/proxy-wasm-go-sdk/proxywasm/types"
+	"github.com/higress-group/wasm-go/pkg/test"
+	"github.com/stretchr/testify/require"
+)
+
+// === helpers ============================================================
+
+// mustConfig marshals m to JSON and fails the test on error.
+func mustConfig(t *testing.T, m map[string]interface{}) json.RawMessage {
+	t.Helper()
+	b, err := json.Marshal(m)
+	require.NoError(t, err)
+	return b
+}
+
+// findHeader returns the value of the first matching header (case-insensitive).
+func findHeader(headers [][2]string, key string) (string, bool) {
+	k := strings.ToLower(key)
+	for _, h := range headers {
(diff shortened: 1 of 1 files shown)


johnlanni (approved):
LGTM

PR #3895, opened 2026-05-29
feat(ai-security-guard): add Embedding API content detection support

JianweiWang:

Add **Embedding API** content detection support to the `ai-security-guard` plugin (under the `MultiModalGuard` action).

Previously the multi-modal guard handled `text_generation`, `image_generation` and `mcp` API types. This PR introduces a new `embedding` API type so that requests/responses of OpenAI-compatible Embedding endpoints can be moderated by the security service.

Key changes:
- **config**: add `ApiEmbedding = "embedding"` API type and a new `responseErrorContentJsonPath` config item (used to extract error content such as `error.message` from non-200 embedding responses).
- **handler** (`multi_modal_guard`): route `embedding` apiType to a new embedding handler for request body / response headers / response body; explicitly skip streaming moderation for embedding (embedding has no streaming responses) with a warning log.
- **embedding handler** (`lvwang/multi_modal_guard/embedding/openai.go`): new handler that extracts content from the request (`input`, supporting both string and string-array inputs) and the response (`data`), and moderates them via the security service.
- **main**: for the `embedding` apiType, do **not** skip response body checking on non-200 responses, so that error content (`error.message`) can still be inspected.

No.

Unit tests added in `main_test.go`:
- `TestEmbeddingConfig` — config parsing incl. backward compatibility when `responseErrorContentJsonPath` is omitted.
- `TestEmbeddingRequest` — request detection (string & array `input`).
- `TestEmbeddingResponse` — response detection.
- `TestEmbeddingStreamingIgnored` — streaming responses are passed through (no moderation).
- `TestEmbeddingNon200Response` — error content checked on non-200 responses.

```bash
cd plugins/wasm-go/extensions/ai-security-guard
go build ./...
go test ./...
```

All packages pass locally.

- Embedding does not support streaming; `responseStreamContentJsonPath` is intentionally ignored for `embedding` apiType.
- Net change is additive — no existing API-type behavior is modified.
- This branch is rebased onto the latest `main` (single commit).

### DIFF
--- a/plugins/wasm-go/extensions/ai-security-guard/config/config.go
+++ b/plugins/wasm-go/extensions/ai-security-guard/config/config.go
@@ -118,6 +118,7 @@ const (
 	ApiTextGeneration  = "text_generation"
 	ApiImageGeneration = "image_generation"
 	ApiMCP             = "mcp"
+	ApiEmbedding       = "embedding"
 )
 
 // provider types
@@ -206,6 +207,7 @@ type AISecurityConfig struct {
 	ResponseStreamContentJsonPath          string
 	ResponseContentFallbackJsonPaths       []string
 	ResponseStreamContentFallbackJsonPaths []string
+	ResponseErrorContentJsonPath           string
 	DenyCode                               int64
 	DenyMessage                            string
 	ProtocolOriginal                       bool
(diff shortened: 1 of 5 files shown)


johnlanni (approved):
LGTM

PR #3924, opened 2026-06-05
fix(helm): move imagePullPolicy from PodSpec to container level in gateway pod template

geekspeng:
## Summary

- Move `imagePullPolicy` from PodSpec level (`spec.template.spec.imagePullPolicy`) to container level (`spec.template.spec.containers[].imagePullPolicy`) in `helm/core/templates/_pod.tpl`

## Problem

When installing Higress with `--set global.imagePullPolicy="IfNotPresent"`, Helm renders `imagePullPolicy` at the PodSpec level, which is not a valid Kubernetes API field. This causes the installation to fail with:

```
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Deployment.spec.template.spec): unknown field "imagePullPolicy" in io.k8s.api.core.v1.PodSpec
```

## Root Cause

In `helm/core/templates/_pod.tpl`, the `imagePullPolicy` block was placed directly under `spec:` (PodSpec level) instead of inside the container definition where it belongs.

**Before (incorrect):**
```yaml
spec:
  imagePullPolicy: IfNotPresent    # Wrong: PodSpec level
  containers:
    - name: higress-gateway
      image: ...
```

**After (correct):**
```yaml
spec:
  containers:
    - name: higress-gateway
      image: ...
      imagePullPolicy: IfNotPresent  # Correct: container level
```

## Test Plan

- [x] Verified the fix by comparing with other templates (`controller-deployment.yaml`, `plugin-server-deployment.yaml`, `redis/statefulset.yaml`) which already correctly place `imagePullPolicy` at the container level
- [x] Run `helm template` with `--set global.imagePullPolicy=IfNotPresent` to verify the rendered manifest is valid
- [x] Run `helm install --dry-run` to validate against the Kubernetes API


CH3CHO (approved):
LGTM. 不过 imagePullPolicy 这个设置似乎没有覆盖所有的 Pod。能否再提交一个 PR 进行修复呢？

PR #3928, opened 2026-06-05
fix(oidc): fix Set-Cookie header corruption during session refresh

Jing-ze:
## Summary

- Fix session refresh failure caused by multiple `Set-Cookie` headers being comma-joined into a single header value
- Bump `oauth2-proxy` to include higress-group/oauth2-proxy#10
- Update `onHttpResponseHeaders` to emit each `Set-Cookie` as a separate response header

## Root Cause

When a session cookie exceeds 4KB, it is split into multiple shards (`_oauth2_proxy_0`, `_oauth2_proxy_1`, ...), each written as a separate `Set-Cookie` header. During session refresh, the `Proxy()` function joined these with `strings.Join(cookies, ",")` into a single string, and the WASM plugin wrote it as one `Set-Cookie` header.

`Set-Cookie` is the only HTTP header that **cannot** be comma-folded ([RFC 6265](https://datatracker.ietf.org/doc/html/rfc6265#section-3)), because the `Expires` attribute contains commas (e.g., `Thu, 05 Jun 2026 09:10:07 GMT`). Browsers silently discard the malformed cookie, so the refreshed session is never saved. The next request carries the stale cookie (whose refresh token has already been consumed by the IDP), resulting in `No valid authentication in request` and an unexpected SSO redirect.

The login path (`redirectToLocation`) already handled this correctly by emitting each `Set-Cookie` as a separate header. This fix aligns the refresh path to the same behavior.

## Changes

**oauth2-proxy** (higress-group/oauth2-proxy#10, already merged):
- Store cookies as `[]string` instead of comma-joined `string` in WASM context

**oidc plugin** (this PR):
- `main.go`: type-assert context value as `[]string` and call `AddHttpResponseHeader` once per cookie
- `go.mod/go.sum`: bump oauth2-proxy to `396b01ca2ca4`


johnlanni (approved):
LGTM

PR #3671, opened 2026-04-01
feat: support claude custom url

ThxCode-Chen:
## Ⅰ. Describe what this PR did
Synchronizes the claudeCustomUrl configuration support for the Claude (Anthropic) provider in plugins/wasm-go/extensions/ai-proxy with the implementation used in the main Higress tree.

provider/provider.go: Adds the claudeCustomUrl field on ProviderConfig (with metadata comments) and parses it from JSON in FromJson.

provider/claude.go: When claudeCustomUrl is set, parses host and path (supports plain origin, path prefix, or a direct /v1/messages URL), adjusts capabilities accordingly, stores custom domain for outbound requests, and keeps the default behavior when the field is empty. TransformRequestHeaders uses the custom host when configured; otherwise it still uses resolveDomain("", claudeDomain) so providerDomain and other domain overrides keep working.

test/claude.go: Adds WASM integration-style tests for config load and for :authority being rewritten to the custom host.

## Ⅱ. Does this pull request fix one issue?

No. This is a parity / sync change; there is no linked issue.

## Ⅲ. Why don't you add test cases (unit test/integration test)? 

Integration coverage was added in test/claude.go (plugin config + request headers). No extra unit tests were added in provider/claude_test.go because the main Higress reference implementation does not add separate CreateProvider / claudeCustomUrl unit tests there either; behavior is covered at the integration layer.

## Ⅳ. Describe how to verify it

From plugins/wasm-go/extensions/ai-proxy:
go test ./provider/... ./test/... -count=1
Manually: configure a Claude provider with claudeCustomUrl (e.g. https://custom.example.com/anthropic) and confirm upstream requests use the expected Host and path for chat/messages traffic.

## Ⅴ. Special notes for reviews

Interaction with providerDomain: Custom URL host wins when claudeCustomUrl is set; when it is unset, existing resolveDomain behavior is unchanged.

Capabilities: Logic matches the main repo: a direct messages endpoint only overrides message-related paths; otherwise the configured path acts as a prefix for default capability paths.

## Ⅵ. AI Coding Tool Usage Checklist (if applicable)

**Please check all applicable items:**

- [√ ] **For new standalone features** (e.g., new wasm plugin or golang-filter plugin):
  - [ ] I have created a `design/` directory in the plugin folder
  - [ ] I have added the design document to the `design/` directory
  - [ ] I have included the AI Coding summary below

- [ ] **For regular updates/changes** (not new plugins):
[...]

PR #3676, opened 2026-04-02
feat: support claude custom url

ThxCode-Chen:
## Ⅰ. Describe what this PR did
Synchronizes the claudeCustomUrl configuration support for the Claude (Anthropic) provider in plugins/wasm-go/extensions/ai-proxy with the implementation used in the main Higress tree.

provider/provider.go: Adds the claudeCustomUrl field on ProviderConfig (with metadata comments) and parses it from JSON in FromJson.

provider/claude.go: When claudeCustomUrl is set, parses host and path (supports plain origin, path prefix, or a direct /v1/messages URL), adjusts capabilities accordingly, stores custom domain for outbound requests, and keeps the default behavior when the field is empty. TransformRequestHeaders uses the custom host when configured; otherwise it still uses resolveDomain("", claudeDomain) so providerDomain and other domain overrides keep working.

test/claude.go: Adds WASM integration-style tests for config load and for :authority being rewritten to the custom host.

## Ⅱ. Does this pull request fix one issue?

No. This is a parity / sync change; there is no linked issue.

## Ⅲ. Why don't you add test cases (unit test/integration test)? 

Integration coverage was added in test/claude.go (plugin config + request headers). No extra unit tests were added in provider/claude_test.go because the main Higress reference implementation does not add separate CreateProvider / claudeCustomUrl unit tests there either; behavior is covered at the integration layer.

## Ⅳ. Describe how to verify it

From plugins/wasm-go/extensions/ai-proxy:
go test ./provider/... ./test/... -count=1
Manually: configure a Claude provider with claudeCustomUrl (e.g. https://custom.example.com/anthropic) and confirm upstream requests use the expected Host and path for chat/messages traffic.

## Ⅴ. Special notes for reviews

Interaction with providerDomain: Custom URL host wins when claudeCustomUrl is set; when it is unset, existing resolveDomain behavior is unchanged.

Capabilities: Logic matches the main repo: a direct messages endpoint only overrides message-related paths; otherwise the configured path acts as a prefix for default capability paths.

## Ⅵ. AI Coding Tool Usage Checklist (if applicable)

**Please check all applicable items:**

- [√ ] **For new standalone features** (e.g., new wasm plugin or golang-filter plugin):
  - [ ] I have created a `design/` directory in the plugin folder
  - [ ] I have added the design document to the `design/` directory
  - [ ] I have included the AI Coding summary below

- [ ] **For regular updates/changes** (not new plugins):
[...]

PR #3935, opened 2026-06-08
fix(ai-proxy): use custom domain as Host header in Claude provider

### LINKED ISSUE #3934: ai-proxy: Claude provider hardcodes Host header to api.anthropic.com, ignoring claudeCustomUrl
## Description

When configuring a Claude provider with a custom backend URL (e.g., Huawei Cloud ModelArts, AWS Bedrock proxy, or any Anthropic-compatible gateway), the `Host` header is unconditionally overwritten to `api.anthropic.com`. This causes requests to be rejected by the actual upstream with 404, because the upstream gateway/WAF validates the Host header and rejects unknown hosts.

## Root Cause

In `plugins/wasm-go/extensions/ai-proxy/provider/claude.go`, the `TransformRequestHeaders` method hardcodes the domain:

```go
func (c *claudeProvider) TransformRequestHeaders(ctx wrapper.HttpContext, apiName ApiName, headers http.Header) {
    util.OverwriteRequestPathHeaderByCapability(headers, string(apiName), c.config.capabilities)
    util.OverwriteRequestHostHeader(headers, claudeDomain)  // always sets Host to api.anthropic.com
    // ...
}
```

Unlike the OpenAI provider (`openai.go`), which checks for a custom domain and uses it when configured:

```go
// openai.go - correct behavior
func (m *openaiProvider) TransformRequestHeaders(...) {
    if m.customDomain != "" {
        util.OverwriteRequestHostHeader(headers, m.customDomain)  // uses custom domain
    } else {
        util.OverwriteRequestHostHeader(headers, defaultOpenaiDomain)
    }
}
```

The Claude provider's `CreateProvider` method also does **not** parse `claudeCustomUrl` to extract a custom domain -- there is no `customDomain` field on the `claudeProvider` struct at all.

As a result, even though `providerDomain` and `providerBasePath` are set in the WasmPlugin config, the Claude provider ignores them for the Host header.

## Reproduction

1. Configure a Claude provider pointing to a non-Anthropic upstream:

```yaml
type: claude
apiTokens:
  - ${secret.ai-key.example}
claudeCustomUrl: https://api.modelarts-maas.com/anthropic/v1
protocol: openai
```

2. Send a chat completion request through the AI proxy.
3. Observe: the upstream receives `Host: api.anthropic.com` instead of `Host: api.modelarts-maas.com`.
4. The upstream gateway returns 404 because it does not recognize the Host.

## Expected Behavior

When `claudeCustomUrl` is configured, the Host header should reflect the domain from that URL (e.g., `api.modelarts-maas.com`). The behavior should mirror what the OpenAI provider already does with `openaiCustomUrl`.

## Existing Fix

PR [#3671](https://github.com/higress-group/higress/pull/3671) ("feat: support claude custom url") by @ThxCode-Chen implements exactly this fix and includes integration tests. It was closed without review or merge. The patch:

- Adds `claudeCustomUrl` parsing in `CreateProvider` (extracts host + path from the URL)
- Adds `customDomain` field to `claudeProvider` struct
- Updates `TransformRequestHeaders` to use `customDomain` when set
- Adds WASM integration tests for custom URL Host header rewriting

## Impact

This bug makes it impossible to use the Claude provider type with any third-party Anthropic-compatible API gateway (Huawei Cloud ModelArts, AWS Bedrock with proxy, LiteLLM, etc.). Users are forced to either:
- Use the OpenAI provider type (losing Claude-specific protocol features), or
- Deploy an EnvoyFilter as a workaround to rewrite the Host header

## Environment

- Higress chart version: 2.2.2
- Plugin: ai-proxy (wasm-go)

Aaron199:
## Description

When `claudeCustomUrl` is configured in a Claude provider, the `TransformRequestHeaders` method unconditionally overwrites the `Host` header to `api.anthropic.com`. This causes requests to third-party Anthropic-compatible gateways (Huawei Cloud ModelArts, AWS Bedrock proxies, LiteLLM, etc.) to be rejected with 404, because the upstream gateway/WAF validates the Host header and rejects unknown hosts.

### `claude.go`

1. **Added `"path"` import** for path joining when computing custom capability paths.
2. **`CreateProvider`** — When `claudeCustomUrl` is set, parses the URL to extract:
   - `customDomain` — the host portion, stored on the provider struct
   - `customPath` — the path portion, used to adjust API capabilities
   - Supports plain origin, path prefix, and direct `/v1/messages` endpoint URLs
3. **`claudeProvider` struct** — Added `customDomain string` field.
4. **`TransformRequestHeaders`** — Uses `customDomain` when configured instead of hardcoded `claudeDomain`. Falls back to `resolveDomain("", claudeDomain)` when no custom URL is set.

### `provider.go`

1. **`ProviderConfig`** — Added `claudeCustomUrl` field with JSON/YAML tags.
2. **`FromJson`** — Parses `claudeCustomUrl` from JSON config.

## Verification

- The behavior when `claudeCustomUrl` is empty is unchanged (uses `resolveDomain("", claudeDomain)`).
- The behavior mirrors the OpenAI provider (`openai.go`), which already does the same for `openaiCustomUrl`.

## Related

Fixes #3934
Ref: #3671 (unmerged prior attempt by @ThxCode-Chen)

### DIFF
--- a/plugins/wasm-go/extensions/ai-proxy/provider/claude.go
+++ b/plugins/wasm-go/extensions/ai-proxy/provider/claude.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"path"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


ThxCode-Chen:
Thank you for your recognition of my code, but the community follows a more common approach: https://github.com/higress-group/higress/pull/3686

I've tested this and haven’t encountered the host blocking issue you mentioned. Could you provide a specific example?

Also, the original PR has been moved here for discussion: https://github.com/higress-group/higress/pull/3676
The original PR has been reviewed by CH3CHO.

PR #3943, opened 2026-06-09
feat: Enhance SSL passthrough support

### LINKED ISSUE #1596: higress支持ssl-passthrough么
想通过kubeconfig实现在外部访问集群；如何实现类似SSL透传特性（nginx.ingress.kubernetes.io/ssl-passthrough: "true"）解决证书认证问题。

zijiren233:
fixes: #1596

This pull request introduces support for SSL passthrough in Ingress resources, allowing certain hosts to be handled using TLS passthrough mode based on annotations. The changes include new annotation parsing logic, updates to the conversion pipeline to honor passthrough ownership, and comprehensive tests to verify the correct behavior. The implementation ensures that only the first root path Ingress for a host can claim SSL passthrough ownership, which is critical for correct gateway and virtual service generation.

Key changes include:

**SSL Passthrough Annotation Support:**
* Added a new annotation parser (`ssl_passthrough.go`) to recognize and parse the `ssl-passthrough` annotation, with a corresponding config struct and logic to set passthrough mode when enabled.
* Registered the new annotation handler in the annotation manager and added a helper method `IsSSLPassthrough` to check if passthrough is enabled on an Ingress. [[1]](diffhunk://#diff-ddb75c5114f0e92c04e27acc0f35c66d3eac13a0d3cf8a10bfe87e59b149dcf5R60-R61) [[2]](diffhunk://#diff-ddb75c5114f0e92c04e27acc0f35c66d3eac13a0d3cf8a10bfe87e59b149dcf5R120-R123) [[3]](diffhunk://#diff-ddb75c5114f0e92c04e27acc0f35c66d3eac13a0d3cf8a10bfe87e59b149dcf5R166)

**Ingress Conversion Pipeline Updates:**
* Updated the gateway and virtual service conversion logic to call `preparePassthroughTLSHostOwners`, which determines which hosts are claimed for passthrough by the first root path Ingress with passthrough enabled. [[1]](diffhunk://#diff-0edc009a5759e00729c3aba40c397870430d0173388d20bef28681f0b0665aa8R443) [[2]](diffhunk://#diff-0edc009a5759e00729c3aba40c397870430d0173388d20bef28681f0b0665aa8R510-R511) [[3]](diffhunk://#diff-0edc009a5759e00729c3aba40c397870430d0173388d20bef28681f0b0665aa8R629-R716)
* Modified the virtual service conversion to use new naming and cluster ID logic, ensuring correct association for passthrough and non-passthrough hosts. [[1]](diffhunk://#diff-0edc009a5759e00729c3aba40c397870430d0173388d20bef28681f0b0665aa8L573-R579) [[2]](diffhunk://#diff-0edc009a5759e00729c3aba40c397870430d0173388d20bef28681f0b0665aa8L588-R610) [[3]](diffhunk://#diff-0edc009a5759e00729c3aba40c397870430d0173388d20bef28681f0b0665aa8R629-R716)
* Added helper functions for extracting root path hosts and for creating passthrough servers. [[1]](diffhunk://#diff-0edc009a5759e00729c3aba40c397870430d0173388d20bef28681f0b0665aa8R629-R716) [[2]](diffhunk://#diff-db0bb2eb8fe267c708bbb870132c07a825cfaeada314b39c174766bef94ebd69R83-R96)

**Testing Improvements:**
* Added comprehensive unit tests for passthrough annotation parsing and the new passthrough ownership logic, including edge cases for multiple Ingresses and fallback scenarios. [[1]](diffhunk://#diff-d6cdd0d4b68843dc3624939b817af46ef3275b6e29cab7dfc6876c3da9da42bbR1-R112) [[2]](diffhunk://#diff-6ecb1b3bc77fc20474e6d94f443d966d84ade708b3b6a10aa4b89902f1df2f2cR113-R440)

**Dependency and Import Updates:**
* Imported the necessary Kubernetes networking API packages and constants for TLS passthrough handling. [[1]](diffhunk://#diff-0edc009a5759e00729c3aba40c397870430d0173388d20bef28681f0b0665aa8R49-R50) [[2]](diffhunk://#diff-6ecb1b3bc77fc20474e6d94f443d966d84ade708b3b6a10aa4b89902f1df2f2cR26) [[3]](diffhunk://#diff-db0bb2eb8fe267c708bbb870132c07a825cfaeada314b39c174766bef94ebd69R18) [[4]](diffhunk://#diff-db0bb2eb8fe267c708bbb870132c07a825cfaeada314b39c174766bef94ebd69R27)

These changes collectively enable robust, annotation-driven SSL passthrough support in the Ingress controller, with clear precedence and test coverage for correct behavior.

### DIFF
--- a/pkg/ingress/config/ingress_config.go
+++ b/pkg/ingress/config/ingress_config.go
@@ -440,6 +440,7 @@ func (m *IngressConfig) convertGateways(configs []common.WrapperConfig) []config
 	if err != nil {
 		IngressLog.Errorf("Get higress https configmap err %v", err)
 	}
+	m.prepareDuplicateTLSHosts(&convertOptions, configs, httpsCredentialConfig)
 	for idx := range configs {
 		cfg := configs[idx]
 		clusterId := common.GetClusterId(cfg.Config.Annotations)
@@ -506,7 +507,11 @@ func (m *IngressConfig) convertVirtualService(configs []common.WrapperConfig) []
 		}
 	}
 
-	m.prepareDuplicateTLSHosts(&convertOptions, configs)
+	httpsCredentialConfig, err := m.httpsConfigMgr.GetConfigFromConfigmap()
+	if err != nil {
+		IngressLog.Errorf("Get higress https configmap err %v", err)
+	}
+	m.prepareDuplicateTLSHosts(&convertOptions, configs, httpsCredentialConfig)
 
 	// convert http route
 	for idx := range configs {
@@ -635,19 +640,62 @@ func virtualServiceNameAndClusterID(cleanHost string, wrapperVS *common.WrapperV
 	return common.CreateConvertedName(constants.IstioIngressGatewayName, cfg.Namespace, cfg.Name, cleanHost), common.GetClusterId(cfg.Annotations)
 }
 
-func (m *IngressConfig) prepareDuplicateTLSHosts(convertOptions *common.ConvertOptions, configs []common.WrapperConfig) {
-	httpsCredentialConfig, err := m.httpsConfigMgr.GetConfigFromConfigmap()
-	if err != nil {
-		IngressLog.Errorf("Get higress https configmap err %v", err)
+func (m *IngressConfig) prepareDuplicateTLSHosts(convertOptions *common.ConvertOptions, configs []common.WrapperConfig, httpsCredentialConfig *cert.Config) {
+	if convertOptions.SSLPassthroughTLSHosts == nil {
+		convertOptions.SSLPassthroughTLSHosts = map[string]*config.Config{}
 	}
+
+	rootPathOwners := map[string]*config.Config{}
+	for idx := range configs {
+		cfg := configs[idx]
+		if cfg.AnnotationsConfig.IsCanary() {
+			continue
+		}
+
+		for _, host := range ingressRootPathHosts(cfg.Config.Spec) {
+			_, exist := rootPathOwners[host]
+			if !exist {
+				rootPathOwners[host] = cfg.Config
+				if cfg.AnnotationsConfig.IsSSLPassthrough() {
+					convertOptions.SSLPassthroughTLSHosts[host] = cfg.Config
+				}
+			}
+		}
+	}
+
 	tlsHostOwners := map[string]*config.Config{}
 	for idx := range configs {
 		cfg := configs[idx]
 		if cfg.AnnotationsConfig.IsCanary() {
 			continue
 		}
+
+		if cfg.AnnotationsConfig.IsSSLPassthrough() {
+			hosts := ingressRuleHosts(cfg.Config.Spec)
+			for _, host := range hosts {
+				if !common.IsSSLPassthroughTLSHostOwner(convertOptions, cfg.Config, host) {
+					common.AddDuplicateTLSHost(convertOptions, cfg.Config, host)
+					continue
+				}
+				owner, exist := tlsHostOwners[host]
+				if exist && !common.SameConfig(owner, cfg.Config) {
+					common.AddDuplicateTLSHost(convertOptions, cfg.Config, host)
+					continue
+				}
+				if !exist {
+					tlsHostOwners[host] = cfg.Config
+				}
+			}
+			continue
+		}
+
 		hosts := ingressTLSHosts(cfg, httpsCredentialConfig)
 		for _, host := range hosts {
+			if _, exist := convertOptions.SSLPassthroughTLSHosts[host]; exist {
+				common.AddDuplicateTLSHost(convertOptions, cfg.Config, host)
+				continue
+			}
+
 			owner, exist := tlsHostOwners[host]
 			if exist && !common.SameConfig(owner, cfg.Config) {
 				common.AddDuplicateTLSHost(convertOptions, cfg.Config, host)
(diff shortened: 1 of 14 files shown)


johnlanni:
I think there is a blocker in the current SSL passthrough ownership logic.

`preparePassthroughTLSHostOwners` precomputes the passthrough owner only from the first root path for each host. This can let a later SSL passthrough Ingress preempt an earlier normal TLS Ingress that already owns `host:443` with non-root paths.

One reproducible case:

1. Ingress A has `example.com`, a TLS secret, and path `/api` -> `api:8443`.
2. Ingress B, processed later, has `nginx.ingress.kubernetes.io/ssl-passthrough: "true"` and path `/` -> `root:443` for the same host.

Because `preparePassthroughTLSHostOwners` only considers root paths, Ingress B becomes the passthrough owner. Then the normal TLS branch in `ConvertGateway` sees a different passthrough owner and skips creating the `SIMPLE` HTTPS server for Ingress A. The final Gateway uses a `PASSTHROUGH` server for `example.com:443`, while the `/api` HTTPRoute/VirtualService from Ingress A is still generated. That route will no longer be reachable over HTTPS because TLS is now passed through instead of terminated.

Relevant code paths:

- `pkg/ingress/config/ingress_config.go:639-661`
- `pkg/ingress/kube/ingressv1/controller.go:496-501`
- `pkg/ingress/kube/ingress/controller.go:516-521`
- `pkg/ingress/kube/ingressv1/controller.go:431-443`
- `pkg/ingress/kube/ingress/controller.go:451-463`

I think the conflict resolution should still preserve the existing conversion order / first owner of `host:443`. A later SSL passthrough Ingress should not preempt an earlier normal TLS Ingress. One possible fix is to avoid using `PassthroughTLSHostOwners` to make the normal TLS branch skip itself; let the existing `wrapperGateway.IsHTTPS()` / `preDomainBuilder` conflict handling decide who already owns the HTTPS listener. Alternatively, the precomputed owner map needs to represent the first Ingress that occupies `host:443`, not just the first root-path passthrough candidate.

zijiren233:
@johnlanni Thanks for the detailed analysis. I think this behavior is aligned with ingress-nginx.

The relevant ingress-nginx flow is:

1. ssl-passthrough is marked at server/host level when any Ingress for the host has the ssl-passthrough annotation: https://github.com/kubernetes/ingress-nginx/blob/dbb11b92ddb77bee9c35e462479129354c84f939/internal/ingress/controller/controller.go#L1434-L1436

2. Locations are then merged by path. The first real location for "/" replaces the default backend; an earlier non-root path such as /api only creates or owns that /api location: https://github.com/kubernetes/ingress-nginx/blob/dbb11b92ddb77bee9c35e462479129354c84f939/internal/ingress/controller/controller.go#L811-L845

3. When building ssl-passthrough backends, ingress-nginx only uses the root location. Non-root locations are ignored for ssl-passthrough: https://github.com/kubernetes/ingress-nginx/blob/dbb11b92ddb77bee9c35e462479129354c84f939/internal/ingress/controller/controller.go#L666-L680

So in the example, Ingress A owns the /api HTTP location, while Ingress B owns the root location with ssl-passthrough enabled. ingress-nginx would use the root location as the passthrough backend for example.com:443. That means the /api route remains valid for HTTP/plain routing, while HTTPS for the host is handled as passthrough.

zijiren233:
However, I believe this is indeed a multi-tenancy issue, and I’m uncertain whether security should be compromised solely for compatibility with NGINX Ingress behavior.

johnlanni:
Thanks, that makes sense for the example I gave.

One follow-up: if the goal is to match ingress-nginx, I think there is another edge case worth clarifying. In ingress-nginx, `SSLPassthrough` is enabled at the server/host level when any Ingress for that host has the annotation, while the passthrough backend is selected from the first root `/` location. That root location does not necessarily belong to the annotated Ingress.

In the current implementation, `preparePassthroughTLSHostOwners` only records an owner when the first root-path Ingress itself has `ssl-passthrough` enabled. So a normal root-path Ingress followed by an ssl-passthrough Ingress for the same host would not create a passthrough owner, whereas ingress-nginx would still enable passthrough for the host and use the existing root location as the backend.

Is this stricter Higress behavior intentional, or should the model separate "host has ssl-passthrough enabled" from "first root-path owner"?

zijiren233:
@johnlanni I agree with this distinction.

The current implementation merges two ingress-nginx concepts into a single owner map, which makes this edge case behave differently from ingress-nginx. ingress-nginx treats them separately:

1. The host/server has ssl-passthrough enabled if any Ingress for the host has the annotation.
2. The passthrough backend is taken from the first root "/" location for that host.

So for this case:

- Ingress A: example.com, path "/" -> root:443, no ssl-passthrough annotation
- Ingress B: example.com, ssl-passthrough annotation, path "/foo" -> foo:443

ingress-nginx enables ssl-passthrough for example.com and uses Ingress A's "/" location as the passthrough backend.

I think Higress should model those two pieces separately for nginx-ingress compatibility.

Does this direction look right to you? If yes, I’ll update the implementation and add regression coverage for both cases.

zijiren233:
@johnlanni I updated the implementation to match the ingress-nginx model more closely.

johnlanni:
Thanks for the update. I rechecked the latest implementation, and the separation now looks right to me:

- any annotated Ingress enables SSL passthrough at the host level
- the first root `/` path owner is used as the passthrough backend

That matches the ingress-nginx model we discussed, and the added regression coverage covers the edge case. I don't see any further blockers from my side.

johnlanni (approved):
LGTM

PR #3956, opened 2026-06-11
fix(ai-proxy): clean client Authorization header for api-key providers

### LINKED ISSUE #3955: ai-proxy: Claude standard mode forwards client Authorization header together with x-api-key
## Problem

When using `ai-proxy` with Claude in standard mode, Higress sets the upstream provider credential in `x-api-key`, but it does not remove the original client `Authorization` header.

If a client calls Higress with:

```http
Authorization: Bearer <gateway-consumer-api-key>
```

the upstream Claude-compatible provider receives:

```http
Authorization: Bearer <gateway-consumer-api-key>
x-api-key: <provider-token>
```

Some upstream providers validate `Authorization` when it is present, even if `x-api-key` is also provided. This causes the upstream to reject the request with `invalid api key`, because the `Authorization` value is a Higress consumer credential, not an upstream provider token.

## Evidence

I reproduced this locally with Higress all-in-one by configuring:

- a Claude provider in standard mode,
- a route protected by key-auth,
- a local echo upstream.

The echo upstream received both headers:

```http
authorization: Bearer <gateway-consumer-api-key>
x-api-key: <provider-token>
anthropic-version: 2023-06-01
x-mse-consumer: <consumer-name>
```

I also verified the upstream provider behavior directly:

| Case | x-api-key | Authorization | Result |
| --- | --- | --- | --- |
| A | provider token | none | 200 |
| B | provider token | Bearer provider token | 200 |
| C | provider token | Bearer gateway consumer key | 401 invalid api key |

## Root Cause

In `plugins/wasm-go/extensions/ai-proxy/provider/claude.go`, Claude Code mode rewrites `Authorization`, but standard mode only sets `x-api-key`:

```go
if c.config.claudeCodeMode {
    headers.Set("authorization", "Bearer "+token)
    headers.Del("x-api-key")
} else {
    headers.Set("x-api-key", c.config.GetApiTokenInUse(ctx))
}
```

Since `ai-proxy` transforms the existing request header set, the original client `Authorization` remains unless it is explicitly removed.

Azure provider has a similar risk because it sets `api-key` without removing `Authorization`.

## Expected Behavior

For providers that authenticate upstream requests with non-`Authorization` headers, Higress should not forward the client-to-gateway `Authorization` header by default.

For Claude standard mode, upstream should receive:

```http
x-api-key: <provider-token>
```

and should not receive the client's gateway `Authorization` header.

## Proposed Fix

In Claude standard mode:

```go
headers.Set("x-api-key", c.config.GetApiTokenInUse(ctx))
headers.Del(util.HeaderAuthorization)
```

In Azure API key mode:

```go
headers.Set("api-key", m.config.GetApiTokenInUse(ctx))
headers.Del(util.HeaderAuthorization)
```

Claude Code mode should remain unchanged because it intentionally uses OAuth Bearer tokens.

XinhhD:
## Summary

Clean client `Authorization` headers for AI providers that authenticate upstream requests with non-`Authorization` API key headers.

This PR updates:

- Claude standard mode: remove client `Authorization` after setting upstream `x-api-key`.
- Azure provider: remove client `Authorization` after setting upstream `api-key`.

Claude Code mode is unchanged because it intentionally uses `Authorization: Bearer <oauth-token>`.

## Why

`ai-proxy` transforms the original downstream request headers. If a provider sets a non-`Authorization` credential header without deleting the original `Authorization`, the upstream can receive both:

```http
Authorization: Bearer <gateway-consumer-api-key>
x-api-key: <provider-token>
```

Some upstream providers validate `Authorization` when present and reject the request because the value is a Higress consumer credential, not an upstream provider token.

## Changes

- In Claude standard mode, delete `Authorization` after setting `x-api-key`.
- In Azure provider, delete `Authorization` after setting `api-key`.
- Keep provider modes that intentionally use `Authorization` unchanged.

## Testing

Manual local reproduction was done with:

- Higress all-in-one.
- A temporary Claude provider in standard mode.
- A key-auth protected AI route.
- A local echo upstream.

Before the fix, upstream received both:

```http
authorization: Bearer <gateway-consumer-api-key>
x-api-key: <provider-token>
```

After applying the patch, expected upstream headers:

```http
x-api-key: <provider-token>
```

and no client `Authorization`.

Recommended command:

```bash
go test ./...
```

Closes #3955


wydream:
@XinhhD 

Thanks for the contribution.
A direct regression test would be useful here: send a Claude standard-mode / Azure request with Authorization: Bearer gateway-token in the incoming headers, then assert that the transformed upstream headers no longer contain Authorization. This would lock in the intended behavior and prevent this header leak from coming back later.

XinhhD:
@wydream Thank you for the suggestion! I have added the requested regression tests for Azure and Claude standard mode in both Go and Wasm unit test suites, verifying that any client-provided `Authorization` header is successfully cleaned up before being sent to the upstream. The tests pass locally.

wydream (approved):
LGTM

PR #3964, opened 2026-06-12
fix: merge InferencePool route configs during HTTPRoute merge

EndlessSeeker:
### Background

When multiple `HTTPRoute` resources with `InferencePool` backends are attached to the same Gateway and hostname, Higress merges them into one generated `VirtualService`. Before this change, `mergeHTTPRoutes()` only merged the HTTP route list and parent annotations, but did not merge `config.Config.Extra`.

`ConfigExtraPerRouteRuleInferencePoolConfigs` is stored in `Extra` and later used to generate per-route GAIE/EPP `ext_proc` configuration. If only the first route's `Extra` survives the merge, later routes can lose their EPP binding and fall back to normal Envoy load balancing.

This mirrors the upstream Istio fix:

- Upstream PR: https://github.com/istio/istio/pull/58393
- Upstream issue: https://github.com/istio/istio/issues/58392

### Changes

- Merge `ConfigExtraPerRouteRuleInferencePoolConfigs` maps while merging HTTPRoute-derived VirtualServices.
- Preserve first-wins behavior for non-InferencePool `Extra` keys.
- Deep copy the base InferencePool config map before merging, matching upstream's race-safety fix for shallow `Config.DeepCopy()` behavior.
- Keep Higress' current `krt.Collection[*config.Config]` type chain instead of adopting the later upstream `krt.Collection[config.Config]` migration.
- Add a Higress-adapted golden regression case based on the upstream Istio test scenario:
  - two independent `HTTPRoute`s
  - same Gateway and hostname
  - different `InferencePool` backends
  - one merged generated `VirtualService`
- Add a focused unit test that directly verifies merged `Extra` retains both per-route InferencePool/EPP configs.

### Tests

```bash
go test ./pkg/ingress/kube/gateway/istio -run '^(TestConvertResources/http|TestMergeHTTPRoutesMergesInferencePoolExtra)$' -count=1
go test ./pkg/ingress/kube/gateway/istio -count=1
git diff --check -- pkg/ingress/kube/gateway/istio/route_collections.go pkg/ingress/kube/gateway/istio/route_collections_test.go pkg/ingress/kube/gateway/istio/conversion_test.go pkg/ingress/kube/gateway/istio/testdata/http.yaml pkg/ingress/kube/gateway/istio/testdata/http.yaml.golden pkg/ingress/kube/gateway/istio/testdata/http.status.yaml.golden
```


johnlanni (approved):
LGTM

PR #3973, opened 2026-06-15
fix(ai-proxy): preserve Vertex thoughtSignature in OpenAI tool calls

### LINKED ISSUE #3974: Vertex Gemini tool calls fail because thoughtSignature is dropped in OpenAI-compatible conversion
# Vertex Gemini tool calls fail because thoughtSignature is dropped in OpenAI-compatible conversion

## Summary

When using the Vertex provider through OpenAI-compatible `/v1/chat/completions`, Gemini 3.x models with thinking and tools can fail on the second tool-calling turn because `ai-proxy` drops Vertex `thoughtSignature` values.

Vertex/Gemini requires the original `thoughtSignature` to be returned with the corresponding `functionCall` in the next request. Higress currently converts only `functionCall.name` and `functionCall.args` to OpenAI-compatible `tool_calls`, so clients have no signature to send back.

## Affected Versions

Checked against:

- Higress latest release `v2.2.2`
- Higress `main` commit `2897c1e83a8254a46824b95d195933eedf909a27`

Both versions appear affected.

## Error

The upstream Vertex response is HTTP 400:

```text
Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:Skill`, position 4. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details.
```

## Trigger Conditions

This happens when all of the following are true:

- Provider is Vertex/Gemini.
- Model uses Gemini 3.x behavior requiring thought signatures for function calling.
- Request uses tools/function calling.
- Thinking/reasoning is enabled, for example `reasoning_effort=high`.
- The conversation continues after a model-generated function call with a tool result.

## Current Behavior

In `provider/vertex.go`:

- Vertex response `functionCall` is converted to OpenAI-compatible `tool_calls`.
- Only `name` and `arguments` are preserved.
- `thoughtSignature` is not represented in `tool_calls`.
- On the next turn, Higress rebuilds Vertex `functionCall` from `tool_calls`, but can only restore `name` and `args`.
- Vertex rejects the follow-up request because `functionCall.thoughtSignature` is missing.

## Expected Behavior

Higress should preserve Gemini thought signatures across the OpenAI-compatible boundary:

- Vertex response `part.thoughtSignature` should be exposed as `tool_calls[].thought_signature`.
- A subsequent OpenAI-compatible request containing `tool_calls[].thought_signature` should restore it to Vertex `functionCall.thoughtSignature`.

## Why This Is Needed

Google documents thought signatures as required conversation state for Gemini thinking + function calling. Clients using Higress OpenAI-compatible APIs cannot preserve this state unless Higress exposes it in the converted response.

Reference:

https://ai.google.dev/gemini-api/docs/thought-signatures

## Suggested Fix

Add a `ThoughtSignature` field to the internal `toolCall` model and map it in both directions:

- Vertex response -> OpenAI-compatible response:
  - `vertexPart.ThoughtSignature` -> `toolCall.ThoughtSignature`
- OpenAI-compatible request -> Vertex request:
  - `toolCall.ThoughtSignature` -> `vertexPart.ThoughtSignature`

The attached patch implements this behavior and adds tests for both directions.

XinhhD:
# Preserve Vertex Gemini thoughtSignature in OpenAI-compatible tool calls

Fixes #3974

## What

This PR preserves Gemini `thoughtSignature` values when converting Vertex function calls to and from OpenAI-compatible `tool_calls`.

The conversion now maps:

- Vertex response `part.thoughtSignature` -> OpenAI-compatible `tool_calls[].thought_signature`
- OpenAI-compatible request `tool_calls[].thought_signature` -> Vertex request `functionCall.thoughtSignature`

## Why

Gemini 3.x thinking + function calling requires the original thought signature to be returned with the corresponding function call in later turns. Without this value, Vertex rejects follow-up requests with:

```text
Function call is missing a thought_signature in functionCall parts.
```

Before this change, Higress only preserved `functionCall.name` and `functionCall.args` through the OpenAI-compatible layer, so clients had no way to send the required signature back.

## Changes

- Add `ThoughtSignature` to the internal OpenAI-compatible `toolCall` struct as `json:"thought_signature,omitempty"`.
- Add `ThoughtSignature` to Vertex `vertexPart` as `json:"thoughtSignature,omitempty"`.
- Preserve the field when building non-streaming OpenAI-compatible responses.
- Preserve the field when building streaming OpenAI-compatible response chunks.
- Restore the field when building Vertex requests from OpenAI-compatible history.
- Add tests for both directions.

## Tests

```bash
go test ./provider
go test ./config ./util ./provider
go test . -run 'TestVertex|Test_bodyDebugFields|Test_getApiName|Test_normalizeOpenAiRequestBody|Test_convertResponseBodyToClaude_glue'
```

Docker build validation:

```bash
cd plugins/wasm-go
GOPROXY=https://goproxy.cn,direct \
PLUGIN_NAME=ai-proxy \
IMG=ai-proxy:local-thought-signature \
BUILDER=higress-registry.cn-hangzhou.cr.aliyuncs.com/plugins/wasm-go-builder:go1.24.4-oras1.0.0 \
make build-image
```

## Compatibility

The new `thought_signature` field is optional. Existing OpenAI-compatible clients that ignore unknown fields continue to work. Clients that preserve unknown tool call fields can now support Gemini thought signatures correctly.


wydream (approved):
LGTM

PR #3985, opened 2026-06-17
fix(ai-proxy): include tool call type in Claude stream deltas

### LINKED ISSUE #3984: Claude streaming tool call argument deltas emit empty tool call type
# Summary

When converting Claude native streaming responses to OpenAI-compatible chat completion chunks, `content_block_delta` events with `input_json_delta` emit `delta.tool_calls[]` entries without `type:"function"`. Because the shared `toolCall` model serializes `Type string `json:"type"` without `omitempty`, the streamed chunk contains `"type":""`.

Strict OpenAI-compatible clients validate every streamed `tool_calls[]` item and reject the empty type.

# Current behavior

A Claude tool-call argument delta can be converted to a chunk like:

```json
{
  "choices": [
    {
      "delta": {
        "tool_calls": [
          {
            "index": 0,
            "type": "",
            "function": { "arguments": "..." }
          }
        ]
      }
    }
  ]
}
```

# Expected behavior

Every streamed OpenAI-compatible tool call delta should include the function tool type:

```json
{
  "index": 0,
  "type": "function",
  "function": { "arguments": "..." }
}
```

# Suggested fix

Set `Type: "function"` in the Claude `input_json_delta` conversion branch in `provider/claude.go`. Bedrock already does this for its `ToolUse` delta path; a regression test can guard both providers.

XinhhD:
# Summary

Fix Claude-native streaming to OpenAI-compatible conversion so tool-call argument delta chunks include `type: "function"`.

Claude `content_block_delta` events with `input_json_delta` previously emitted a `tool_calls[]` element with only `index` and `function.arguments`. Since `toolCall.Type` serializes without `omitempty`, strict OpenAI-compatible clients received `"type":""` and rejected the stream chunk.

Fixes #3984

# Changes

- Set `Type: "function"` for Claude `input_json_delta` stream conversion.
- Add a Claude regression test for streamed tool-call argument deltas.
- Add a Bedrock regression test confirming its existing ToolUse delta path keeps `type: "function"`.

# Tests

```bash
go test ./provider -run TestClaudeProvider_StreamToolCallArgumentDeltaIncludesFunctionType -count=1
go test ./provider -run TestBedrockStreamToolCallArgumentDeltaIncludesFunctionType -count=1
go test ./provider -count=1
```

### DIFF
--- a/plugins/wasm-go/extensions/ai-proxy/provider/bedrock_thinking_test.go
+++ b/plugins/wasm-go/extensions/ai-proxy/provider/bedrock_thinking_test.go
@@ -115,6 +115,28 @@ func TestBedrockStreamPreservesClaudeNativeIndexesAndStops(t *testing.T) {
 	assert.Equal(t, 2, *events[0].Payload.Index)
 }
 
+func TestBedrockStreamToolCallArgumentDeltaIncludesFunctionType(t *testing.T) {
+	provider := &bedrockProvider{}
+	ctx := newMockMultipartHttpContext()
+
+	chunk, err := provider.convertEventFromBedrockToOpenAI(ctx, ConverseStreamEvent{
+		ContentBlockIndex: 0,
+		Delta: &converseStreamEventContentBlockDelta{
+			ToolUse: &toolUseBlockDelta{Input: `{"path":"/tmp/example"}`},
+		},
+	})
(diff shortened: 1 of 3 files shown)


wydream (approved):
LGTM

PR #3989, opened 2026-06-17
feat(ai-proxy): vLLM passthrough for Anthropic Messages and newer OpenAI endpoints

Rand01ph:

The vLLM provider only exposed the OpenAI-style endpoints (chat/completions,
completions, models, embeddings) plus rerank. Recent vLLM versions also natively
serve the Anthropic Messages API and the newer OpenAI endpoints. This exposes them
as **passthrough** capabilities (forwarded without protocol translation):

- `POST /v1/messages` (Anthropic Messages)
- `POST /v1/messages/count_tokens` (Anthropic token counting)
- `POST /v1/responses` (OpenAI Responses API)
- `POST /v1/audio/transcriptions`
- `POST /v1/audio/translations`

Changes:
- `provider/vllm.go` — register the five endpoints in `DefaultCapabilities` and
  resolve them in `GetApiName`. `count_tokens` is matched before `/v1/messages`
  because the former path contains the latter as a substring.
- `provider/provider.go` — add the `ApiNameAnthropicCountTokens` api name and the
  `/v1/messages/count_tokens` path, and include it in `needToProcessRequestBody`.
- `main.go` — register `/v1/messages/count_tokens` in the framework
  `pathSuffixToApiName` table so it is recognized for providers using the OpenAI
  (non-`original`) protocol, which is how vLLM providers are configured.

No linked issue.

Unit tests are included:
- `provider/vllm_test.go` — `DefaultCapabilities` and `GetApiName` (including the
  count_tokens-before-messages ordering, plus a regression guard for the existing
  endpoints).
- `main_test.go` — a `Test_getApiName` case for `/v1/messages/count_tokens`.

1. `cd plugins/wasm-go/extensions/ai-proxy && go test ./...` — passes.
2. Configure a vLLM provider (protocol `openai/v1`) pointing at a vLLM server that
   serves these endpoints, then call `/v1/messages`, `/v1/messages/count_tokens`,
   `/v1/responses`, `/v1/audio/transcriptions` and `/v1/audio/translations` through
   the gateway and confirm they reach the upstream and return native responses.

Passthrough only — the plugin does not translate protocols for these endpoints; it
routes the request to the configured vLLM upstream and streams the native response
back, matching vLLM's behaviour of serving these APIs directly.

Tool: Claude Code (Claude Opus). Paraphrased instructions:
- "Update the Higress vLLM provider plugin — it currently only supports the OpenAI
  endpoints, but recent vLLM versions also serve the Anthropic protocol endpoints
  and the newer OpenAI endpoints. Add passthrough support for Anthropic
  `/v1/messages` and `/v1/messages/count_tokens`, OpenAI `/v1/responses`, and the
  `/v1/audio` transcription/translation endpoints."
- Follow-up: "Add `/v1/messages/count_tokens` support" and make sure it routes
  correctly for vLLM providers (which use the `openai/v1` protocol, not `original`).

### AI Coding Summary

- **Key decisions:** implemented the endpoints as passthrough (vLLM serves them
  natively, no translation needed); resolve `count_tokens` before `/v1/messages`
  because the path is a superstring; register `count_tokens` in the framework
  `pathSuffixToApiName` table (not only the provider's `GetApiName`) because for an
  `openai/v1`-protocol provider the framework uses the path table and only consults
  the provider's `GetApiName` for `original`-protocol providers.
- **Major changes:** see section I.
- **Limitations:** passthrough only; no request/response body transformation for
  these endpoints.


wydream (changes requested):
@Rand01ph 

Thanks for the PR. The overall direction looks reasonable to me: exposing these newer vLLM-native endpoints as passthrough capabilities keeps the implementation simple and avoids unnecessary protocol translation.

I noticed two areas worth addressing or clarifying:

1. `vllmCustomUrl` direct endpoint handling does not include the newly added endpoints.

   `isVllmDirectPath` currently only treats paths ending with `/completions` or `/rerank` as direct endpoint URLs. With this PR, if a user configures `vllmCustomUrl` as a direct endpoint such as:

   - `http://host/v1/responses`
   - `http://host/v1/messages`
   - `http://host/v1/audio/transcriptions`

   the code will treat that path as a base path and may generate incorrect upstream paths such as `/v1/responses/responses` or `/v1/messages/messages`.

   If direct endpoint URLs are intended to be supported for vLLM, please extend `isVllmDirectPath` for the newly added endpoints and add a small unit test for the path mapping behavior.

2. The new tests cover the main routing/capability mapping, but not a few important edge cases.

   Current tests cover `DefaultCapabilities`, `GetApiName`, and global recognition of `/v1/messages/count_tokens`, which is good. It would be useful to also cover:

   - `vllmCustomUrl` path rewriting for one of the newly added endpoints
   - audio transcription/translation passthrough behavior, especially multipart requests
   - behavior when non-vLLM providers receive `/v1/messages/count_tokens`

   These do not look like major blockers, but adding coverage here would make the passthrough behavior more robust and easier to maintain.

Rand01ph:
Thanks for the careful review — both points addressed in 75c8a27.

**1. Direct endpoint handling.** `isVllmDirectPath` now recognizes the newly added endpoints (`/responses`, `/messages`, `/messages/count_tokens`, `/audio/transcriptions`, `/audio/translations`), so a `vllmCustomUrl` pointing directly at one of them is forwarded as-is instead of being treated as a base path and double-appended (e.g. no more `/v1/responses/responses`).

**2. Tests.** Added:
- `TestVllm_isVllmDirectPath` — direct-path detection for the new endpoints and base paths.
- `TestVllmProviderInitializer_CreateProvider_customUrl` — `vllmCustomUrl` path rewriting: a base path gets the per-API suffix appended, a direct endpoint is forwarded unchanged.
- `TestVllm_passthroughBodyAndSupport` — audio transcription/translation are **not** in `needToProcessRequestBody`, so multipart bodies pass through untouched (JSON processing applies only to messages/count_tokens/responses); and a provider that doesn't declare the `count_tokens` capability rejects the now-globally-recognized path via `isSupportedAPI`, so non-vLLM providers are unaffected.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

wydream (approved):
LGTM

PR #3990, opened 2026-06-17
fix(ai-proxy): add ids for Vertex tool calls

### LINKED ISSUE #3929: ai-proxy: Vertex standard path loses Gemini tool replay metadata; replay succeeds only with vertexOpenAICompatible
We are using Higress `/v1/chat/completions` with Vertex + Gemini `gemini-3.5-flash`.

With the standard Vertex provider:

```
type: vertex
protocol: openai
the first tool-call response does not contain:
```

choices[0].message.tool_calls[0].id
choices[0].message.tool_calls[0].extra_content.google.thought_signature
Because of that, second-round tool replay fails with upstream 400 missing thought_signature.

After switching the same model to:

```
type: vertex
protocol: openai
vertexOpenAICompatible: true
```

Higress returns both fields, and the same replay flow succeeds.

What we observed
Standard Vertex path
First-round response through Higress:

```
{
  "choices": [
    {
      "message": {
        "role": "assistant",
        "tool_calls": [
          {
            "type": "function",
            "function": {
              "name": "get_weather",
              "arguments": "{\"city\":\"Tokyo\"}"
            }
          }
        ]
      }
    }
  ]
}
```
Replay then fails with:

```
{
  "error": {
    "code": 400,
    "message": "Function call is missing a thought_signature in functionCall parts..."
  }
}
```

With vertexOpenAICompatible: true

First-round response through Higress:

```
{
  "choices": [
    {
      "message": {
        "role": "assistant",
        "tool_calls": [
          {
            "id": "function-call-...",
            "type": "function",
            "function": {
              "name": "get_weather",
              "arguments": "{\"city\":\"Tokyo\"}"
            },
            "extra_content": {
              "google": {
                "thought_signature": "..."
              }
            }
          }
        ]
      }
    }
  ]
}
```

Second-round replay succeeds with 200 OK.

Expected behavior
Either:
The standard type: vertex path should preserve enough Gemini replay metadata for multi-turn tool calling, or
Higress docs/UI should clearly state that Gemini multi-turn tool replay requires vertexOpenAICompatible: true
Additional gap
In our deployment, the UI did not expose vertexOpenAICompatible, so we had to patch YAML directly.

Questions

> Is this a bug in the standard Vertex translation path, or a known limitation?
> If it is a limitation, is vertexOpenAICompatible: true the officially recommended path for Gemini multi-turn tool replay?
> Can the UI expose this option directly?

XinhhD:
# Summary

Generate OpenAI-compatible tool call IDs when the standard Vertex provider converts native Gemini `functionCall` parts into `tool_calls[]`.

Vertex/Gemini native function calls do not expose OpenAI `tool_calls[].id`, but strict OpenAI-compatible clients expect each tool call to have a string ID. Without it, clients can reject streamed tool-call chunks with errors such as `Expected id to be a string`.

Fixes #3929

# Changes

- Generate `call_<uuid>` IDs for non-streaming Vertex function-call responses.
- Keep generated stream tool-call IDs stable per tool-call index within the request context.
- Add regression tests for non-streaming IDs and streaming stable IDs.

# Tests

```bash
go test ./provider -run 'TestVertexProvider(PreservesFunctionCallThoughtSignature|StreamToolCallIncludesStableID)' -count=1\ngo test ./provider -count=1\ngo test ./config ./util ./provider -count=1\ngit diff --check\n```\n


wydream (approved):
LGTM

PR #3997, opened 2026-06-20
feat(tls): default to skip backend cert verification, with opt-in strict paths

CH3CHO:

This PR changes Higress's default behavior for upstream TLS client connections to **skip server certificate verification**, so that backends using self-signed certificates or internal CAs work out of the box without 502 handshake failures.

Three call sites that synthesize `ClientTLSSettings` are routed through a new shared helper, `util.DefaultSkipSimpleTLS(sni)`, in `pkg/ingress/kube/util/tls.go`:

| Path | File | Before | After |
|------|------|--------|-------|
| Direct registry HTTPS | `registry/direct/watcher.go::generateDestinationRule` | literal `InsecureSkipVerify=true` | `util.DefaultSkipSimpleTLS(w.getSni(se))` |
| Nacos MCP server HTTPS | `registry/nacos/mcpserver/watcher.go::generateDrForMcpServer` (HttpsProtocol branch) | literal `InsecureSkipVerify=true` | `util.DefaultSkipSimpleTLS("")` |
| Ingress annotation HTTPS | `pkg/ingress/kube/annotations/upstreamtls.go::processSimple` | unconditional `InsecureSkipVerify=true` | reads `config.UpstreamTLS.SSLVerify` and sets the flag only when verification is disabled, aligning with `processMTLS` |

Supporting changes:

- New helper `pkg/ingress/kube/util/tls.go` + its unit test (`tls_test.go`).
- New `registry/direct/watcher_test.go` covering HTTPS / HTTP service entries and SNI propagation.
- Two new test cases in `pkg/ingress/kube/annotations/upstreamtls_test.go` for `processSimple` with `SSLVerify=true` (strict) and `SSLVerify=false` (skip).
- Unified the protobuf wrapper import to `google.golang.org/protobuf/types/known/wrapperspb` in the touched files (the legacy `github.com/golang/protobuf/ptypes/wrappers` `BoolValue` is a deprecated type alias for the same underlying type).
- Adjusted `registry/nacos/watcher_test.go` weight / labels expectations to match the implementation (weight `< 0.5` rounds to 0, not 1; metadata without a `protocol` key is cleared). These are pre-existing test mismatches surfaced by the same code path, not new behavior.

The `BackendTLSPolicy` path (`pkg/ingress/kube/gateway/istio/backend_policies.go`) is **intentionally left untouched** (reverted to `main`): the unconditional `InsecureSkipVerify=true` there conflicts with Istio's `validateTLS`, which rejects the generated `DestinationRule` whenever a user also configures `SubjectAltNames` / `CaCertificates` / `CredentialName`. Users who need strict verification on that path configure `BackendTLSPolicy` explicitly; users who need self-signed cert compatibility use a different ingress surface.

No.

It addresses a general usability gap (self-signed backend certs breaking TLS) rather than a tracked issue, and the user-driven scope narrowed it to three call sites while leaving the Gateway API path alone.

Tests are included:

- `pkg/ingress/kube/util/tls_test.go` — `TestDefaultSkipSimpleTLS` covers both populated and empty SNI.
- `registry/direct/watcher_test.go` — `TestGenerateDestinationRule_HTTPS_SkipVerify`, `TestGenerateDestinationRule_HTTP_ReturnsNil`, `TestGenerateDestinationRule_MatchesHelper` exercise the helper integration.
- `pkg/ingress/kube/annotations/upstreamtls_test.go::TestApplyTrafficPolicy` — two new sub-cases verify the `SSLVerify=true` strict path and the `SSLVerify=false` skip path. The four pre-existing cases continue to pass.

From the repo root:

```bash
# Build
go build ./...

# Targeted tests (the changed packages)
go test -count=1 ./pkg/ingress/kube/util/...
go test -count=1 ./pkg/ingress/kube/annotations/...
go test -count=1 ./registry/direct/...
go test -count=1 ./registry/nacos/mcpserver/...
go test -count=1 ./registry/nacos/...

# Verify the Gateway API file is unchanged
git diff main HEAD -- pkg/ingress/kube/gateway/istio/backend_policies.go
# (should print nothing)
```

Expected behavior matrix:

| Scenario | Outcome |
|----------|---------|
| Direct registry points at a self-signed HTTPS upstream | TLS handshake succeeds (skip verify) |
| Nacos MCP server registered over HTTPS with a self-signed cert | TLS handshake succeeds (skip verify) |
| Ingress annotation `proxy-ssl-verify: on` + HTTPS upstream | Strict verification (no skip) — Envoy uses the configured CA / SAN |
| Ingress annotation without `proxy-ssl-verify: on` (or `off`) + HTTPS | TLS handshake succeeds (skip verify) |
| `BackendTLSPolicy` with `Validation.Hostname` / `SubjectAltNames` / `CACertificateRefs` | Strict verification; generated `DestinationRule` is accepted by Istio validateTLS |

N/A

### DIFF
--- a/pkg/ingress/kube/annotations/upstreamtls.go
+++ b/pkg/ingress/kube/annotations/upstreamtls.go
@@ -18,7 +18,7 @@ import (
 	"regexp"
 	"strings"
 
-	"github.com/golang/protobuf/ptypes/wrappers"
+	wrapperspb "google.golang.org/protobuf/types/known/wrapperspb"
 	networking "istio.io/api/networking/v1alpha3"
 	"istio.io/istio/pilot/pkg/model/credentials"
 
@@ -155,7 +155,7 @@ func processMTLS(config *Ingress) *networking.ClientTLSSettings {
 	if !config.UpstreamTLS.SSLVerify {
 		// This api InsecureSkipVerify hasn't been support yet.
 		// Until this pr https://github.com/istio/istio/pull/35357.
-		tls.InsecureSkipVerify = &wrappers.BoolValue{
+		tls.InsecureSkipVerify = &wrapperspb.BoolValue{
 			Value: false,
 		}
 	}
@@ -170,15 +170,15 @@ func processMTLS(config *Ingress) *networking.ClientTLSSettings {
 func processSimple(config *Ingress) *networking.ClientTLSSettings {
 	tls := &networking.ClientTLSSettings{
 		Mode: networking.ClientTLSSettings_SIMPLE,
-		InsecureSkipVerify: &wrappers.BoolValue{
-			Value: true,
-		},
 	}
-
+	// Default to skip verification to support self-signed/internal-CA backends.
+	// Respect explicit opt-in via proxy-ssl-verify: on (which sets SSLVerify=true).
+	if !config.UpstreamTLS.SSLVerify {
+		tls.InsecureSkipVerify = &wrapperspb.BoolValue{Value: true}
+	}
 	if config.UpstreamTLS.EnableSNI && config.UpstreamTLS.SNI != "" {
 		tls.Sni = config.UpstreamTLS.SNI
 	}
-
 	return tls
 }
 
--- a/pkg/ingress/kube/annotations/upstreamtls_test.go
+++ b/pkg/ingress/kube/annotations/upstreamtls_test.go
@@ -17,10 +17,10 @@ package annotations
 import (
 	"testing"
 
-	"github.com/golang/protobuf/ptypes/wrappers"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"google.golang.org/protobuf/testing/protocmp"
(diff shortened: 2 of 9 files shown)


EndlessSeeker:
Could you help confirm the intended behavior for the mTLS path here?

In `processMTLS`, when `SSLVerify` is false, the code still sets `InsecureSkipVerify` to `false`, which means certificate verification is not skipped. This seems opposite to the expected meaning of `proxy-ssl-verify: off`.

If skip verification is supported for mTLS now, should this be `InsecureSkipVerify: true`? If it is still not supported for mTLS, should we document that `proxy-ssl-verify` only takes effect for the SIMPLE TLS path?

EndlessSeeker:
One more concern: for the direct registry and Nacos MCP Server paths, this seems to change the default behavior from not explicitly skipping certificate verification to explicitly setting `InsecureSkipVerify=true`.

Could we add a user-facing switch for this instead of making it unconditional? For example, keep the old behavior by default, and only set `InsecureSkipVerify=true` when the user explicitly opts in. Otherwise users who rely on the previous certificate verification behavior do not have a way to restore it after upgrading.

PR #4009, opened 2026-06-24
feat(key-auth): consumer group metadata + X-Mse-Consumer-Group header injection

geekspeng:
## Summary

Adds `group` metadata to `key-auth` consumers and injects the `X-Mse-Consumer-Group` request header on successful authentication. This is **PR 1 of 2** that adds consumer-group quota support to the AI gateway (PR 2 is `feat/ai-quota-group-quota`).

- Add `Consumer.Group string` field with admin UI annotations; old YAML defaults to `""` (no behavior change).
- Reject configs where any group name duplicates a consumer name (`G ∩ N = ∌`) at startup. Error lists ALL conflict pairs so operators see every collision in one restart.
- Inject `X-Mse-Consumer-Group` only when `Consumer.Group != ""`. Header name uppercase, downstream reads lowercase.
- Replace `credential2Name map[string]string` with `credential2Consumer map[string]Consumer` so the auth path gets the full struct (incl. Group) in one lookup.

Group is intentionally **not** format-validated — same rule as `name`. Format validation was removed from the spec as over-defense.

## Why this PR ships first

ai-quota (PR 2) maps both `group` and `consumer` to the same Redis key namespace `chat_quota:<name>`. If a group name duplicates a consumer name, the group's shared-pool writes would clobber the consumer's private-pool key. key-auth's G∩N check is the single chokepoint that prevents this; ai-quota cannot ship safely until it lands.

## Test plan

- [x] `go test ./...` in `plugins/wasm-go/extensions/key-auth` passes (existing suite + 6 new tests)
- [x] `Consumer.Group` parses correctly; absent field defaults to `""`
- [x] `G ∩ N = ∌` startup check: single conflict, multiple conflicts, and shared-group (legal) all covered
- [x] `X-Mse-Consumer-Group` injected only when `Group != ""`; verified absent in both lowercase and uppercase form when group is empty
- [x] `credential2Consumer` rename verified across main.go + both test files (no stray `credential2Name` references in key-auth)

### DIFF
--- a/plugins/wasm-go/examples/key-auth/README.md
+++ b/plugins/wasm-go/examples/key-auth/README.md
@@ -35,6 +35,7 @@ description: Key 认证插件配置参考
 | `credential` | string   | `credential` 和 `credentials` 二选一 | -      | 配置该consumer的一个访问凭证 |
 | `credentials` | array of string | `credential` 和 `credentials` 二选一 | -      | 配置该consumer的多个访问凭证，不能与 `credential` 同时配置 |
 | `name`       | string   | 必填     | -      | 配置该consumer的名称     |
+| `group`      | string   | 选填     | -      | 配置该调用方所属的分组（ai-quota 等下游插件会按 group 共享额度池）。group 名不可与任何 consumer.name 重名。 |
 
 ### 鉴权配置（非必需）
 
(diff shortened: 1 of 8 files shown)


CH3CHO:
The active `key-auth` is the C++ version, not the Go one, whose functions are not fully aligned with  the former.

PR #4013, opened 2026-06-24
fix(oidc): fail closed when verifier is unavailable

EndlessSeeker:
## Summary
- return a 503 local response when the OIDC verifier is unavailable instead of forwarding the request upstream
- add regression tests for protected-route fail-closed behavior and allowlist bypass behavior
- align the OIDC module proxy-wasm-go-sdk version with the wasm-go test harness used by other plugins

Addresses GHSA-54c2-m23g-6wp4.

## Tests
- `GOOS=wasip1 GOARCH=wasm go build -buildmode=c-shared -o main.wasm ./`
- `go test -coverprofile=coverage-oidc.out -covermode=atomic -coverpkg=./... ./...`


Jing-ze:
因为OIDC插件有一个初始化的流程，如果用户配置错误，会导致oidc初始化失败，fail close就会导致用户的所有需要oidc鉴权请求失败，这个是符合预期的吗

---
Because the OIDC plug-in has an initialization process, if the user configuration is wrong, it will cause the oidc initialization to fail, and fail close will cause all the user's oidc authentication requests to fail. Is this as expected?

EndlessSeeker:
> 因为OIDC插件有一个初始化的流程，如果用户配置错误，会导致oidc初始化失败，fail close就会导致用户的所有需要oidc鉴权请求失败，这个是符合预期的吗
> 
> Because the OIDC plug-in has an initialization process, if the user configuration is wrong, it will cause the oidc initialization to fail, and fail close will cause all the user's oidc authentication requests to fail. Is this as expected?

如果初始化的时候用户配置是正确的，而且oidc服务是正常的，插件正常启动，但是后续运行中oidc服务出现了异常，这种情况是不是会导致全部请求被默认放行，从而无法起到安全认证的效果了

Jing-ze:
> > 因为OIDC插件有一个初始化的流程，如果用户配置错误，会导致oidc初始化失败，fail close就会导致用户的所有需要oidc鉴权请求失败，这个是符合预期的吗
> > Because the OIDC plug-in has an initialization process, if the user configuration is wrong, it will cause the oidc initialization to fail, and fail close will cause all the user's oidc authentication requests to fail. Is this as expected?
> 
> 如果初始化的时候用户配置是正确的，而且oidc服务是正常的，插件正常启动，但是后续运行中oidc服务出现了异常，这种情况是不是会导致全部请求被默认放行，从而无法起到安全认证的效果了

是的，有这个问题，这个修复看起来没问题

PR #4027, opened 2026-06-26
fix(ai-proxy): pass through Vertex OpenAI-compatible unicode escapes

storyicon:

This PR fixes Vertex OpenAI-compatible response handling in ai-proxy.

Vertex OpenAI-compatible APIs may return valid ASCII-safe JSON/SSE with non-ASCII text encoded as `\uXXXX`. The proxy should pass those responses through unchanged and let downstream JSON/SSE clients decode them.

The previous proxy-side decoder rewrote JSON escape sequences directly, which could corrupt surrogate-pair characters such as emoji into replacement characters.

Changes:
- Pass through Vertex OpenAI-compatible streaming and non-streaming responses unchanged.
- Remove the now-unused Unicode escape decoder helpers.
- Add regression coverage for escaped Chinese text and surrogate-pair emoji in JSON and SSE responses.

No linked issue.

Test cases are added/updated.

```bash
go test ./util -v
go test -run TestVertex -v
```

This fix is scoped to Vertex OpenAI-compatible response passthrough behavior. It does not change Vertex Express mode, Anthropic passthrough, or transformed Gemini response handling.


storyicon:
I think the gateway should not rewrite Unicode escape sequences in OpenAI-compatible model responses unless there is a very strong reason to do so.

For this path, the upstream response is already valid JSON/SSE. Escapes such as `\uXXXX` are part of the JSON representation and should be decoded by the downstream JSON/SSE client when it parses the response. If the gateway rewrites those bytes, it adds assumptions about JSON string escaping, SSE framing, chunk boundaries, and Unicode composition.

Those assumptions are easy to get wrong in edge cases, for example surrogate pairs, split streaming chunks, escaped content inside strings, or future response fields. Once the gateway mutates the model output before the client sees it, otherwise valid LLM output can be corrupted silently and cause unpredictable customer impact.

So unless there is a compelling reason that cannot be handled at the client/protocol layer, I believe the safer behavior for Vertex OpenAI-compatible responses is to pass through the upstream response unchanged.

Issue #4035, opened 2026-06-30
CORS plugin issue #1743 implementation coordination


johnlanni:
This issue tracks coordination for splitting and implementing CORS plugin fixes related to #1743.

All comments may be posted from the same GitHub account by different coding agents. Each agent comment must start with:

```text
Agent: <agent-name>
Scope: <owned task area>
Branch/PR: <branch or PR link, if any>
Status: planning | in-progress | blocked | ready-for-review | merged
```

Ownership plan:

- Agent Coordinator: behavior contract and task coordination
- Agent A: request flow and invalid CORS handling
- Agent B: config parsing, matching, wildcard behavior, and origin pattern matching
- Agent C: `Vary: Origin` and response header merging
- Agent D: docs, final regression tests, and `task.md` cleanup

Initial coordination rules:

- Do not change another agent's owned files/functions without commenting here first.
- Link every PR back to this issue and to #1743.
- Before opening a PR, comment with the intended behavior changes and test coverage.
- After opening a PR, comment with the PR link and remaining risks.
- After merge, update the relevant task status in `plugins/wasm-go/extensions/cors/task.md`.

Behavior contract is pending the Coordinator agent's first decision comment.

johnlanni:
Agent: Coordinator
Scope: behavior contract and task coordination
Branch/PR: none
Status: decided

Decision:
- Invalid actual CORS requests must continue upstream. They must not be rejected with a direct `403` by the CORS plugin. The plugin should omit CORS allow/expose/credential/max-age headers for these responses, and should strip upstream CORS policy headers so upstream cannot bypass the configured CORS policy.
- Direct preflight handling applies only when `IsCorsRequest && IsPreFlight`. Same-origin `OPTIONS` requests with preflight-like headers must continue upstream.
- Valid preflight requests should be handled directly with `204 No Content`, no body, detail `cors.preflight`, and the configured CORS response headers.
- Invalid preflight requests should be handled directly with `204 No Content`, no body, detail `cors.preflight.invalid`, and no `Access-Control-Allow-*`, `Access-Control-Expose-Headers`, `Access-Control-Allow-Credentials`, or `Access-Control-Max-Age` headers. This creates a failed CORS result without turning the CORS plugin into an HTTP authorization filter.
- `allow_methods: ["*"]` means any non-empty requested method is allowed. For preflight responses, echo the requested method instead of returning a fixed method list or literal `*`.
[...]

PR #4036, opened 2026-06-30
fix: align cors plugin behavior with cors semantics

johnlanni:
## Summary

Addresses part of #1743.
Coordination and worker verification are tracked in #4035.

This PR aligns the CORS plugin behavior with browser CORS semantics and the coordination contract from #4035:

- Invalid actual CORS requests now continue upstream instead of returning a direct 403. The plugin omits CORS allow headers and strips upstream CORS policy headers so browsers enforce the result.
- Direct local preflight handling is gated by both `IsCorsRequest` and `IsPreFlight`; same-origin `OPTIONS` requests with preflight-like headers continue upstream.
- Valid and invalid preflight requests are answered directly with `204 No Content`; invalid preflight responses omit CORS policy headers so browsers fail the preflight.
- Default method/header parsing now splits comma-separated defaults correctly and trims request header tokens.
- `allow_methods: ["*"]` echoes the requested preflight method instead of returning an incomplete fixed method list.
- `allow_headers: ["*"]` echoes normalized requested preflight headers and omits `Access-Control-Allow-Headers` when none were requested.
- Origin pattern regexes are anchored to avoid substring matches.
- `Vary: Origin` is merged when `Access-Control-Allow-Origin` is request-specific.
- README and README_EN document the new behavior, including the `expose_headers: ["*"]` plus credentials compatibility caveat.

## Agent Ownership

- Agent A: request/response flow, invalid actual request behavior, direct preflight handling, preflight status/detail/header behavior, and upstream CORS header stripping.
- Agent B: config parsing, method/header matching, wildcard behavior, default parsing, and origin pattern anchoring.
- Agent C: `Vary: Origin` merge behavior and focused response-header tests.
- Agent D: documentation updates, regression-test coverage review, and task tracking cleanup.
- Coordinator: integrated review, branch/commit/PR submission, and #4035 status update.

## Tests

Run from `plugins/wasm-go/extensions/cors`:

```bash
go test -count=1 ./...
```

Run from repository root:

```bash
git diff --check
```

Both passed before PR creation.

## Residual Risk

- Same-origin detection still depends on Higress/Envoy runtime scheme/host behavior under TLS termination, forwarded scheme/host, and host rewrite scenarios. This PR preserves the existing detection model and records that deployment-level verification as follow-up.
- `expose_headers: ["*"]` with `allow_credentials: true` remains accepted for compatibility. Documentation clarifies that browsers treat `*` as a literal header name for credentialed requests.


plugins/wasm-go/extensions/cors/main.go
```diff
@@ -94,19 +100,17 @@ func onHttpRequestHeaders(ctx wrapper.HttpContext, corsConfig config.CorsConfig,
...
-	if !httpCorsContext.IsValid {
-		headers := make([][2]string, 0)
-		headers = append(headers, [2]string{config.HeaderPluginTrace, "trace"})
-		proxywasm.SendHttpResponseWithDetail(http.StatusForbidden, "cors.forbidden", headers, []byte("Invalid CORS request"), -1)
-		return types.ActionPause
-	}
-
 	// Response directly when it is cors preflight request
-	if httpCorsContext.IsPreFlight {
+	if httpCorsContext.IsCorsRequest && httpCorsContext.IsPreFlight {
```
johnlanni, line 104: Agent: Agent A
Scope: request flow and invalid CORS handling

This gate intentionally requires both `IsCorsRequest` and `IsPreFlight`. `Process` can identify an OPTIONS request with preflight-like headers even when the Origin is same-origin, and that request should continue upstream instead of being swallowed by the CORS plugin.

plugins/wasm-go/extensions/cors/config/cors_config.go
```diff
@@ -80,6 +79,7 @@ func newOriginPatternFromString(declaredPattern string) OriginPattern {
 			patternValue += ":(" + strings.ReplaceAll(portList, ",", "|") + ")"
 		}
 	}
+	patternValue = "^" + patternValue + "$"
```
johnlanni, line 82: Agent: Agent B
Scope: config parsing, matching, wildcard behavior, and origin pattern matching

Anchoring the generated regex makes origin patterns match the full Origin value. Without `^`/`$`, `regexp.MatchString` can accept a malicious suffix such as `http://api.example.com.evil.com` for a pattern intended to cover only `*.example.com`.

plugins/wasm-go/extensions/cors/main.go
```diff
@@ -94,19 +100,17 @@ func onHttpRequestHeaders(ctx wrapper.HttpContext, corsConfig config.CorsConfig,
...
 		headers := make([][2]string, 0)
 		headers = append(headers, [2]string{config.HeaderPluginTrace, "trace"})
-		proxywasm.SendHttpResponseWithDetail(http.StatusOK, "cores.preflight", headers, nil, -1)
+		if !httpCorsContext.IsValid {
+			proxywasm.SendHttpResponseWithDetail(http.StatusNoContent, "cors.preflight.invalid", headers, nil, -1)
```
johnlanni, line 108: Agent: Agent A
Scope: request flow and invalid CORS handling

Invalid preflight now returns an OK-class `204` with no CORS policy headers. That makes the browser fail the preflight through normal CORS evaluation while avoiding the old gateway-level `403` behavior.

plugins/wasm-go/extensions/cors/config/cors_config.go
```diff
@@ -151,10 +151,10 @@ func (c *CorsConfig) FillDefaultValues() {
 		c.allowOrigins = []string{defaultMatchAll}
 	}
 	if len(c.allowHeaders) == 0 {
-		c.allowHeaders = []string{defaultAllowHeaders}
+		c.allowHeaders = splitCommaSeparatedValues(defaultAllowHeaders)
```
johnlanni, line 154: Agent: Agent B
Scope: config parsing, matching, wildcard behavior, and origin pattern matching

The default method and header constants are comma-separated strings, while matching and response serialization operate on individual tokens. Using the shared splitter here fixes the old method-splitting bug and prevents default headers from being stored as one unmatchable value.

plugins/wasm-go/extensions/cors/main.go
```diff
@@ -120,6 +124,7 @@ func onHttpResponseHeaders(ctx wrapper.HttpContext, corsConfig config.CorsConfig
 	// Remove upstream cors response headers if existed
 	proxywasm.RemoveHttpResponseHeader(config.HeaderAccessControlAllowOrigin)
 	proxywasm.RemoveHttpResponseHeader(config.HeaderAccessControlAllowMethods)
+	proxywasm.RemoveHttpResponseHeader(config.HeaderAccessControlAllowHeaders)
```
johnlanni, line 127: Agent: Agent A
Scope: request flow and invalid CORS handling

Because invalid actual CORS requests now continue upstream, the response path must strip every upstream CORS policy header before deciding whether to add plugin-managed headers. Including `Access-Control-Allow-Headers` closes the previous gap where upstream policy could conflict with or bypass the plugin config.

plugins/wasm-go/extensions/cors/config/cors_config.go
```diff
@@ -360,21 +360,29 @@ func (c *CorsConfig) checkHeaders(requestHeaders string) (string, bool) {
...
+	if len(allowHeaders) == 0 {
+		return "", false
 	}
 
-	// Return all request headers when allowHeaders contains *
-	if c.allowHeaders[0] == defaultMatchAll {
-		return requestHeaders, true
+	// Return normalized request headers when allowHeaders contains *.
+	if containsMatchAll(allowHeaders) {
+		return strings.Join(splitCommaSeparatedValues(requestHeaders), ","), true
```
johnlanni, line 370: Agent: Agent B
Scope: config parsing, matching, wildcard behavior, and origin pattern matching

For wildcard `allow_headers`, echoing the normalized request header list is intentional. It lets `Access-Control-Allow-Headers` reflect exactly what the preflight asked for after trimming/dropping empty tokens, and it naturally omits the response header when the preflight did not request any headers.

plugins/wasm-go/extensions/cors/config/cors_config.go
```diff
@@ -391,20 +399,46 @@ func (c *CorsConfig) checkMethods(requestMethod string) (string, bool) {
...
+		return "", false
+	}
 
 	// Find method existed in allowMethods configuration
-	for _, method := range c.allowMethods {
+	allowMethods := splitCommaSeparatedValues(strings.Join(c.allowMethods, ","))
+	for _, method := range allowMethods {
 		if method == defaultMatchAll {
-			return defaultAllAllowMethods, true
+			return requestMethod, true
```
johnlanni, line 411: Agent: Agent B
Scope: config parsing, matching, wildcard behavior, and origin pattern matching

Wildcard `allow_methods` echoes the requested method so extension methods such as `PROPFIND` are actually usable. The previous fixed list could validate the request as allowed while returning an allow-methods header that still caused the browser to fail the preflight.

plugins/wasm-go/extensions/cors/main.go
```diff
@@ -94,19 +100,17 @@ func onHttpRequestHeaders(ctx wrapper.HttpContext, corsConfig config.CorsConfig,
...
+	if httpCorsContext.IsCorsRequest && httpCorsContext.IsPreFlight {
 		headers := make([][2]string, 0)
 		headers = append(headers, [2]string{config.HeaderPluginTrace, "trace"})
-		proxywasm.SendHttpResponseWithDetail(http.StatusOK, "cores.preflight", headers, nil, -1)
+		if !httpCorsContext.IsValid {
+			proxywasm.SendHttpResponseWithDetail(http.StatusNoContent, "cors.preflight.invalid", headers, nil, -1)
+			return types.ActionPause
+		}
+		headers = append(headers, buildCorsHeaders(httpCorsContext)...)
+		headers = appendVaryOriginHeader(headers, httpCorsContext.AllowOrigin)
```
johnlanni, line 112: Agent: Agent C
Scope: Vary: Origin and response header merging

Valid preflight responses are sent directly from the request-header phase, so they will not pass through `onHttpResponseHeaders`. Appending `Vary: Origin` here keeps local preflight responses cache-correct when `Access-Control-Allow-Origin` echoes a specific request origin.

plugins/wasm-go/extensions/cors/main.go
```diff
@@ -140,24 +145,110 @@ func onHttpResponseHeaders(ctx wrapper.HttpContext, corsConfig config.CorsConfig
 	}
 
 	// Add Cors headers when it is cors and valid request
+	for _, header := range buildCorsHeaders(httpCorsContext) {
+		proxywasm.AddHttpResponseHeader(header[0], header[1])
+	}
+	addVaryOriginHeader(httpCorsContext.AllowOrigin)
```
johnlanni, line 151: Agent: Agent C
Scope: Vary: Origin and response header merging

`Vary` is merged after the plugin adds its CORS policy headers, so the helper sees the final response-header state. This also preserves upstream non-CORS cache directives because the cleanup above strips CORS policy headers only and intentionally leaves existing `Vary` values available to merge.

plugins/wasm-go/extensions/cors/main.go
```diff
@@ -140,24 +145,110 @@ func onHttpResponseHeaders(ctx wrapper.HttpContext, corsConfig config.CorsConfig
...
+	merged := make([]string, 0, len(values)+len(extraValues))
+	seen := make(map[string]struct{}, len(values)+len(extraValues))
+
+	for _, value := range append(values, extraValues...) {
+		for _, token := range strings.Split(value, ",") {
+			token = strings.TrimSpace(token)
+			if len(token) == 0 {
+				continue
+			}
+			normalizedToken := strings.ToLower(token)
```
johnlanni: Agent: Agent C
Scope: Vary: Origin and response header merging

The dedupe key is normalized to lowercase, while the emitted token keeps the first spelling seen from the response. That avoids duplicate `Origin` tokens such as `origin, Origin` without rewriting unrelated `Vary` values that may matter to caches or tests.

plugins/wasm-go/extensions/cors/README_EN.md
```diff
@@ -15,15 +15,17 @@ Plugin execution priority: `340`
...
 | max_age               | number           | Optional | 86400 seconds                                                                                                              | Maximum time for browsers to cache CORS results, in seconds. <br/>Within this time frame, browsers will reuse the previous inspection results.                                                                                                 |
 > Note
 > * allow_credentials is a very sensitive option, please enable it with caution. Once enabled, allow_credentials and allow_origins cannot both be *, if both are set, the allow_origins value of "*" takes effect.
 > * allow_origins and allow_origin_patterns can be set simultaneously. First, check if allow_origins matches, then check if allow_origin_patterns matches.
-> * Illegal CORS requests will return HTTP status code 403, with the response body content as "Invalid CORS request".
+> * For actual CORS requests, if the Origin or Method does not match the configuration, the plugin continues the request upstream, does not add CORS response headers, and removes upstream CORS policy response headers. Browsers then block the result according to CORS rules.
```
johnlanni, line 43: Agent: Agent D
Scope: docs, final regression tests, and task.md cleanup

This doc line is intentional because the plugin no longer treats invalid actual CORS requests as gateway authorization failures. The request still reaches upstream, but the plugin withholds generated CORS headers and strips upstream CORS policy headers so browsers enforce the CORS failure.

plugins/wasm-go/extensions/cors/main_vary_test.go
```diff
@@ -0,0 +1,190 @@
...
+		})
+	}
+}
+
+func TestIssue1743LiteralWildcardAllowOriginDoesNotAddVaryOrigin(t *testing.T) {
+	headers := appendVaryOriginHeader([][2]string{
+		{config.HeaderAccessControlAllowOrigin, "*"},
+	}, "*")
+
+	require.False(t, test.HasHeader(headers, headerVary))
```
johnlanni, line 96: Agent: Agent C
Scope: Vary: Origin and response header merging

This test locks the cache rule boundary: literal `Access-Control-Allow-Origin: *` is not request-origin-specific, so adding `Vary: Origin` would be unnecessary cache fragmentation. The helper should only add `Vary` when the allow-origin value depends on the incoming `Origin`.

plugins/wasm-go/extensions/cors/README_EN.md
```diff
@@ -15,15 +15,17 @@ Plugin execution priority: `340`
...
+> * For CORS preflight requests, valid requests are answered directly with `204 No Content` and the configured CORS response headers. Invalid requests are answered directly with `204 No Content`, but without `Access-Control-Allow-*`, `Access-Control-Expose-Headers`, `Access-Control-Allow-Credentials`, or `Access-Control-Max-Age`; browsers treat the preflight as failed.
```
johnlanni, line 44: Agent: Agent D
Scope: docs, final regression tests, and task.md cleanup

This captures the new preflight contract in user-facing terms: both valid and invalid preflight requests return an OK-class 204, but only valid preflights include the CORS policy headers. Invalid preflights fail in the browser because the allow headers are absent, not because the plugin returns 403.

plugins/wasm-go/extensions/cors/README_EN.md
```diff
@@ -15,15 +15,17 @@ Plugin execution priority: `340`
...
-| expose_headers        | array of string  | Optional | -                                                                                                                           | Allowed Headers for the responder to carry that are not part of CORS specifications during cross-origin access. * can be used to indicate any Header is allowed.                                                                                 |
+| allow_methods         | array of string  | Optional | GET, PUT, POST, DELETE, PATCH, OPTIONS                                                                                     | Allowed Methods for cross-origin access, for example: GET, POST, etc. `*` can be used to indicate all Methods are allowed; preflight responses echo the current `Access-Control-Request-Method`.                                               |
```
johnlanni, line 35: Agent: Agent D
Scope: docs, final regression tests, and task.md cleanup

The echo behavior is documented here because `*` is implemented as compatibility-safe preflight output rather than a literal `Access-Control-Allow-Methods: *`. Echoing the requested method avoids returning an incomplete fixed method list and works for extension methods such as PROPFIND.

plugins/wasm-go/extensions/cors/README_EN.md
```diff
@@ -15,15 +15,17 @@ Plugin execution priority: `340`
...
+| allow_headers         | array of string  | Optional | DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,<br/>If-Modified-Since,Cache-Control,Content-Type,Authorization | Allowed Headers for the requester to carry that are not part of CORS specifications during cross-origin access. `*` can be used to indicate any Header is allowed; preflight responses echo normalized `Access-Control-Request-Headers`, and omit `Access-Control-Allow-Headers` when no request headers were requested. |
+| expose_headers        | array of string  | Optional | -                                                                                                                           | Allowed Headers for the responder to expose during cross-origin access. `*` can be used, but for credentialed requests browsers treat `Access-Control-Expose-Headers: *` as a literal header name instead of exposing all headers.                |
```
johnlanni, line 37: Agent: Agent D
Scope: docs, final regression tests, and task.md cleanup

This warning is kept in the config table because the plugin intentionally preserves backward compatibility and accepts `expose_headers: ["*"]` with credentials. The important user-facing nuance is browser behavior: credentialed requests treat `*` as a literal exposed header name, so this setting does not expose every response header.

plugins/wasm-go/extensions/cors/main.go
```diff
@@ -140,24 +145,110 @@ func onHttpResponseHeaders(ctx wrapper.HttpContext, corsConfig config.CorsConfig
...
+	filteredHeaders := make([][2]string, 0, len(headers)+1)
+	for _, header := range headers {
+		if !strings.EqualFold(header[0], headerVary) {
+			filteredHeaders = append(filteredHeaders, header)
+		}
+	}
+	return append(filteredHeaders, [2]string{headerVary, mergedVary})
+}
+
+func addVaryOriginHeader(allowOrigin string) {
```
johnlanni, line 186: 这里envoy遇到多个vary头是会自动合并的，不需要插件逻辑做get合并处理

johnlanni: Agent: Agent C
Scope: Vary: Origin merge behavior and Vary tests

已按建议调整：`addVaryOriginHeader` 不再调用 `GetHttpResponseHeaders`、不再在插件侧解析/合并/去重多个 `Vary` 头，而是仅在 `Access-Control-Allow-Origin` 为请求特定 origin 时追加 `Vary: Origin`，由 Envoy 负责多个 `Vary` 头的合并。

同时更新了 focused Vary 测试：保留“特定 origin 会追加 Vary”和“wildcard origin 不追加 Vary”的覆盖，并把已有 Vary 的用例改为验证插件追加行为，而不是验证插件内合并行为。

Verification:
- `git diff --check`: passed
- `go test -count=1 ./...` under `plugins/wasm-go/extensions/cors`: passed

Commit: 2bb28474e5b2c33c436d86873706ee01e6f77c0a

plugins/wasm-go/extensions/cors/README.md
```diff
@@ -10,25 +10,44 @@ description: 跨域资源共享插件配置参考
...
-插件执行阶段：`授权阶段`
-插件执行优先级：`340`
+插件执行阶段：`认证阶段`
+插件执行优先级：`2000`
+
+## 版本说明
+
+### 2.0.1
+
+相比 `2.0.0`，本版本对齐浏览器 CORS 语义，并将插件执行优先级调整为 `2000`。
```
CH3CHO, line 20: 所以这里其实没有最大值是吗，界面上限制最大只能设置为 1000 的约束是否需要调整？

plugins/wasm-go/extensions/cors/README.md
```diff
@@ -10,25 +10,44 @@ description: 跨域资源共享插件配置参考
...
+
+具体变更和升级注意事项：
+
+* 实际 CORS 请求的 Origin 或 Method 不匹配时，插件不再直接返回 `403`，而是继续转发到后端；插件不会添加 CORS 允许响应头，并会移除后端返回的 CORS policy 响应头，由浏览器按 CORS 规则拦截。若已有监控或客户端逻辑依赖网关直接返回 `403`，升级后需要相应调整。
+* CORS 预检请求统一由插件直接返回 `204 No Content`。非法预检请求不会返回 `Access-Control-Allow-*`、`Access-Control-Expose-Headers`、`Access-Control-Allow-Credentials` 或 `Access-Control-Max-Age`，浏览器会判定预检失败；若已有逻辑依赖非法预检返回 `403`，升级后需要相应调整。
```
CH3CHO, line 25: > 若已有逻辑依赖非法预检返回 `403`，升级后需要相应调整。

按理说预检响应 JS 是拿不到的吧。

plugins/wasm-go/extensions/cors/cors.yaml
```diff
@@ -63,5 +63,5 @@ spec:
       configDisable: false
       ingress:
         - ingress-cors-httpbin
-  url: oci://docker.io/2456868764/cors:1.0.0
-  imagePullPolicy: Always
\ No newline at end of file
+  url: oci://docker.io/2456868764/cors:2.0.1
```
CH3CHO: 这个 URL 是否应该改为指向我们的官方仓库？

plugins/wasm-go/extensions/cors/config/cors_config.go
```diff
@@ -391,20 +399,46 @@ func (c *CorsConfig) checkMethods(requestMethod string) (string, bool) {
...
+func containsMatchAll(values []string) bool {
+	for _, value := range values {
+		if value == defaultMatchAll {
+			return true
+		}
+	}
+	return false
+}
+
 func (c *CorsConfig) isPreFlight(origin, method, controllerRequestMethod string) bool {
```
CH3CHO: `controllerRequestMethod` is a possible typo of `controlRequestMethod`. If so, a fix would be suggested.

CH3CHO:
Thanks for the detailed Agent A/B/C/D breakdown in the inline comments — the design rationale for the fail-open behavior is clear and aligned with the Fetch spec.

I want to flag one residual-risk scenario that I think deserves a stronger upgrade note, specifically for the `allow_credentials: true` case. The current README warning focuses on monitoring/SLO that depend on a gateway-side `403`, but there's a more subtle data-flow change worth calling out:

**Before (1.0.0):** invalid Origin → 403 from the gateway. Request never reaches upstream.

**After (2.0.1):** invalid Origin → `ActionContinue`. Request reaches upstream with `Cookie`/`Authorization` headers intact; only the *response* is sanitized.

Browser CORS is a **client-side read-side enforcement** mechanism. It does not prevent:

1. **State-changing side effects** on behalf of an authenticated user (CSRF-class). A `POST /api/transfer` from a non-allowed origin will still mutate upstream state even though JS can't read the response.
2. **Side-channel disclosure** via status code (200 vs 404), timing (password-check latency), or response-size (search-result count) — all observable to attacker JS through `fetch` failure metadata.
3. **Logging exposure**: the full request URL, upstream response code, and body size end up in Higress access logs / Envoy upstream logs / upstream application logs, accessible to operators.

The classic affected deployment pattern is: *upstream authenticates the session cookie and trusts the gateway to enforce the Origin allowlist.* That's a reasonable assumption with 1.0.0's fail-closed behavior, and it silently breaks after this upgrade.

Three suggestions, in order of effort:

**1. README hardening (low effort, high value)** — In the 2.0.1 upgrade notes, add an explicit callout:

> ⚠️ When `allow_credentials: true`, requests from disallowed origins **still reach upstream** with the user's credentials. Browser CORS only blocks JS from reading the response — it does not block request side effects, side-channel leakage, or upstream log exposure. Make sure your upstream service independently validates Origin / uses CSRF tokens, or stays on 1.0.0.

**2. Optional `strict_cors` config flag (medium effort)** — Lets operators who relied on the old fail-closed behavior keep it:

```yaml
strict_cors: true   # preserve 1.0.0 403 behavior for invalid origins
```

Default `false` keeps the new Fetch-spec-aligned behavior.

**3. Test coverage for the credentials case (low effort)** — Add a test that asserts: an invalid origin request with a `Cookie` header does reach the upstream mock, and the response CORS headers are stripped. This locks in the design contract so a future refactor doesn't accidentally drift back to fail-closed (which would silently break the browser-side CORS semantics the PR is trying to align with).

Happy to send a follow-up PR for any of these if useful. Otherwise, suggestion 1 alone would address my HIGH concern.

plugins/wasm-go/extensions/cors/README.md
johnlanni: 是的

plugins/wasm-go/extensions/cors/README.md
johnlanni: 嗯标准浏览器场景不会，还是写了这个变动保险一些

plugins/wasm-go/extensions/cors/cors.yaml
johnlanni:
```suggestion
  url: oci://higress-registry.cn-hangzhou.cr.aliyuncs.com/plugins/cors:2.0.1
```

plugins/wasm-go/extensions/cors/config/cors_config.go
johnlanni: Good catch — fixed in 45947ea6, renamed `controllerRequestMethod` to `controlRequestMethod` for consistency with the rest of the code.

PR #4123, opened 2026-07-10
test(e2e): add claude conformance cases for ai-proxy

### LINKED ISSUE #1719: Implement claude LLM mock server & e2e test case

123123213weqw:

Add e2e conformance cases for the **claude** provider of the `ai-proxy` wasm plugin, covering both the non-streaming and streaming request paths.

- **claude case 1 (non-streaming)**: sends an OpenAI-shaped chat completion request (`stream:false`) to host `api.anthropic.com`, and asserts the OpenAI-shaped response that `responseClaude2OpenAI` produces. The mock echoes the last user message text (`你好，你是谁？`) inside an Anthropic `content:[{type:text,text:...}]` block with `stop_reason:"end_turn"`. The plugin converts it to `choices[0].message.content` and maps `end_turn` to `finish_reason:"stop"`; the `model` field is the mapped model `claude-3-5-sonnet` (from `ctxKeyFinalRequestModel`). `id`/`created`/`usage` are ignored via `JsonBodyIgnoreFields`.
- **claude case 2 (streaming)**: sends the same request with `stream:true` and only asserts `StatusCode: 200` plus `Content-Type: text/event-stream`, because the mock's SSE sequence carries dynamic timestamps/usage that cannot be matched byte-for-byte.

Provider config: `type: claude`, `apiTokens: [fake_token]`, `modelMapping: {"gpt-3": claude-3-5-sonnet, "*": claude-3-haiku}`. A new Ingress routes `api.anthropic.com` to the shared `llm-mock-service`.

fixes #1719

This PR itself adds the e2e conformance tests for the claude provider. The mock server side that these cases depend on lives in the `higress-group/mock-server` repo (claude mock, already published in the latest `llm-mock` image).

Run the conformance suite:

```
go vet -C test/e2e/conformance ./tests/...
```

The only `go vet` finding is the pre-existing non-constant format string warning in `httproute-limit.go` (unrelated to this change). The claude cases compile cleanly and follow the same pattern as the existing openai/grok/dify cases.

- The expected non-streaming body was derived directly from `responseClaude2OpenAI` + `stopReasonClaude2OpenAI` in `plugins/wasm-go/extensions/ai-proxy/provider/claude.go`: `end_turn` -> `finishReasonStop` (`"stop"`), and `Model` is `ctx.GetStringContext(ctxKeyFinalRequestModel, "")` which holds the mapped model (`claude-3-5-sonnet` for a `gpt-3` request).
- The streaming case intentionally only checks status + content-type, matching the approach used by other streaming cases whose mock SSE contains dynamic fields.
- `go vet` was run with the `external/` submodule links populated; only the known `httproute-limit.go` warnings remain.

### DIFF
--- a/test/e2e/conformance/tests/go-wasm-ai-proxy.go
+++ b/test/e2e/conformance/tests/go-wasm-ai-proxy.go
@@ -1207,6 +1207,52 @@ data: [DONE]
 					},
 				},
 			},
+			{
+				Meta: http.AssertionMeta{
+					TestCaseName:  "claude case 1: non-streaming request",
+					CompareTarget: http.CompareTargetResponse,
+				},
+				Request: http.AssertionRequest{
+					ActualRequest: http.Request{
+						Host:        "api.anthropic.com",
+						Path:        "/v1/messages",
+						Method:      "POST",
+						ContentType: http.ContentTypeApplicationJson,
+						Body:        []byte(`{"model":"gpt-3","messages":[{"role":"user","content":"你好，你是谁？"}],"stream":false}`),
+					},
+				},
+				Response: http.AssertionResponse{
+					ExpectedResponse: http.Response{
+						StatusCode:           200,
+						ContentType:          http.ContentTypeApplicationJson,
+						JsonBodyIgnoreFields: []string{"id", "created", "usage"},
+						Body:                 []byte(`{"id":"x","choices":[{"index":0,"message":{"role":"assistant","content":"你好，你是谁？"},"finish_reason":"stop","logprobs":null}],"created":0,"model":"claude-3-5-sonnet","object":"chat.completion","usage":{}}`),
+					},
+				},
+			},
+			{
+				Meta: http.AssertionMeta{
+					TestCaseName:  "claude case 2: streaming request",
+					CompareTarget: http.CompareTargetResponse,
+				},
+				Request: http.AssertionRequest{
+					ActualRequest: http.Request{
+						Host:        "api.anthropic.com",
+						Path:        "/v1/messages",
+						Method:      "POST",
+						ContentType: http.ContentTypeApplicationJson,
+						Body:        []byte(`{"model":"gpt-3","messages":[{"role":"user","content":"你好，你是谁？"}],"stream":true}`),
+					},
+				},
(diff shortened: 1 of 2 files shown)


CH3CHO:
```
2026-07-11T02:49:28.1884103Z     --- FAIL: TestHigressConformanceTests/WasmPluginAiProxy (387.23s)
2026-07-11T02:49:28.1885753Z         --- FAIL: TestHigressConformanceTests/WasmPluginAiProxy/WasmPlugins_ai-proxy (387.10s)
```

The test doesn't pass.

123123213weqw:
Thanks for flagging the failing `higress-wasmplugin-test (GO)` check. The problem was the request path used by the two new claude cases.

**Root cause**

Both claude cases sent their requests to `/v1/messages`. In the ai-proxy plugin the global path mapping (`getApiName` in `main.go`) resolves `/v1/messages` to `ApiNameAnthropicMessages`. Because the claude provider natively advertises that API (`DefaultCapabilities` maps `ApiNameAnthropicMessages` -> `/v1/messages`), the plugin treats the request as **native-claude passthrough** (the branch at `main.go:260`): it skips the OpenAI->Claude request transformation and the Claude->OpenAI response transformation, leaving only model mapping.

The llm-mock-server answers the `/v1/messages` path with a Claude-format body. Since the response was passed through unconverted, the captured response did not match the expected OpenAI `chat.completion` assertion, so the case failed.

**Fix**

Changed both claude cases to send requests to `/v1/chat/completions`, the same path every other provider case uses. With this path the apiName resolves to `ApiNameChatCompletion`, so:
- `TransformRequestBody` converts the OpenAI request to Claude format (and the provider rewrites the upstream path to `/v1/messages`), and
- `TransformResponseBody` / `OnStreamingResponseBody` convert the Claude response back to OpenAI.

The mock still receives the request at `/v1/messages` (the provider rewrites the path via `OverwriteRequestPathHeaderByCapability`), returns Claude format, and the plugin hands back the expected OpenAI-format body. The ingress, `WasmPlugin` config and `modelMapping` are unchanged.

This is a one-line path change per case and should make the wasmplugin (GO) e2e run pass.

PR #4127, opened 2026-07-11
feat(helm): decouple gateway Promtail from the bundled o11y stack

### LINKED ISSUE #4126: feat(helm): decouple gateway Promtail access-log shipping from the bundled o11y stack + configurable Loki URL
### Feature request

The gateway access-log shipping added in #846 works, but its Promtail sidecar is **welded to the bundled observability stack** and its push target is **hardcoded**, so there is no way (via values) to ship the gateway access log to an **externally-managed Loki** without also deploying the bundled Loki/Grafana/Prometheus.

#### Current behavior

- `higress-core` gates the Promtail sidecar (and the `accessLogFile` → file rewrite) on `global.o11y.enabled`.
- `higress-console` gates the **bundled Loki + Grafana + Prometheus** on `merge (dict) .Values.o11y .Values.global.o11y`. Because Sprig `merge`/mergo treats `false` as *empty*, a local `higress-console.o11y.enabled: false` is silently overwritten by `global.o11y.enabled: true` — so **turning on the Promtail sidecar forces the whole bundled stack on**.
- The Promtail push URL is hardcoded in `helm/core/templates/promtail.yaml`: `http://higress-console-loki.<ns>:3100/loki/api/v1/push` (only the namespace is templated).

#### Motivation

Users who already run their own observability stack currently have no clean path:

- #1926 — users ask how to centrally collect access/error logs with their own tooling; the only answer today is "enable `global.o11y.enabled`", which brings the whole bundled stack.
- #2456 (`日志过大导致 loki OOM`) and #3662 (`开监控之后内存持续增加直到溢出反复重启`) — the bundled stack's resource footprint is exactly why some users want to point Promtail at their own, separately-operated Loki instead.

#### Proposal (backward-compatible)

Add two values under the existing `global.o11y.promtail` block in `helm/core`:

```yaml
global:
  o11y:
    enabled: false
    promtail:
      enabled: false   # NEW: inject the Promtail sidecar + write the access-log file,
                       #      WITHOUT deploying the bundled Loki/Grafana/Prometheus
      lokiUrl: ""      # NEW: Promtail push target; empty => current default
                       #      http://higress-console-loki.<ns>:3100/loki/api/v1/push
```

Wherever the sidecar / file path is gated, use `or o11y.enabled (dig "promtail" "enabled" false o11y)` (nil-safe). In `promtail.yaml` the client URL becomes `{{ $config.lokiUrl | default (printf "http://higress-console-loki.%s:3100/loki/api/v1/push" .Release.Namespace) }}`.

**Backward compatibility:** both new keys default to empty/`false`, so existing installs and the `global.o11y.enabled: true` bundled mode render **byte-identically** (verified with `helm template` diff).

**Scope (intentionally small):** only the Promtail path in `helm/core`; `higress-console` is not touched. Loki authentication (basic/bearer) and multi-tenancy (`X-Scope-OrgID`) are **out of scope for this change** (single-tenant, unauthenticated Loki) and can be a follow-up.

I have this implemented + verified with `helm template` (backward-compat control render and the new external-Loki render) and `helm lint`. PR to follow.

---

#846 引入的网关访问日志采集能力可用，但其 Promtail 边车**与内置可观测性三件套硬绑定**、推送地址**硬编码**，导致无法仅通过 values 把网关访问日志推送到**外部自建的 Loki** 而不连带部署内置 Loki/Grafana/Prometheus。

**现状**：`higress-core` 的 Promtail 边车与 `accessLogFile` 改写由 `global.o11y.enabled` 控制；`higress-console` 的内置三件套由 `merge (dict) .Values.o11y .Values.global.o11y` 控制，而 Sprig `merge`/mergo 把 `false` 当空值，导致本地 `higress-console.o11y.enabled: false` 被 `global.o11y.enabled: true` 覆盖——**一开边车就强制部署整套内置栈**；且推送地址硬编码为 `http://higress-console-loki.<ns>:3100/loki/api/v1/push`。

**动机**：#1926 用户想用自己的工具集中收日志；#2456、#3662 内置栈 OOM/内存问题，正是想改用自建 Loki 的理由。

**提案（向后兼容）**：在 `global.o11y.promtail` 下新增 `enabled`（在不部署内置栈的前提下注入边车+写文件）与 `lokiUrl`（推送地址，留空复现当前默认）。gate 改为 `or o11y.enabled (dig "promtail" "enabled" false o11y)`（nil 安全）。两个新值默认 empty/false，存量安装与 `global.o11y.enabled: true` 模式**字节级不变**。**范围**：仅 `helm/core`；Loki 鉴权与多租户（`X-Scope-OrgID`）不在本次范围内，留作后续。PR 随后提交。

weetime:

Decouples the gateway Promtail sidecar from the bundled observability stack and makes the Loki push URL configurable, so the gateway access log can be shipped to an **externally-managed Loki** without deploying the bundled Loki/Grafana/Prometheus.

- Adds `global.o11y.promtail.enabled` — inject the Promtail sidecar + write the access-log file to `/var/log/proxy/access.log`, **independent of** `global.o11y.enabled` (which additionally deploys the bundled stack).
- Adds `global.o11y.promtail.lokiUrl` — configurable push target (full URL); empty defaults to the current `higress-console-loki` service.
- Gates use `or o11y.enabled (dig "promtail" "enabled" false o11y)` (nil-safe).

Only `helm/core` is touched; `higress-console` is untouched.

fixes #4126

The chart has no unit-test harness and the change is values-gated template wiring. Verified with `helm template` diffs and `helm lint` (see below).

Backward compatibility — **byte-identical** to the pre-change render:

```
helm template t ./helm/core                                 # defaults → stdout access log, no sidecar
helm template t ./helm/core --set global.o11y.enabled=true  # bundled mode → unchanged
```

New behavior:

```
helm template t ./helm/core \
  --set global.o11y.promtail.enabled=true \
  --set global.o11y.promtail.lokiUrl=http://my-loki:3100/loki/api/v1/push
```
renders the `promtail` sidecar, `accessLogFile: /var/log/proxy/access.log`, and `- url: http://my-loki:3100/loki/api/v1/push`, with no bundled backends.

nil-safe regression guard: `helm template t ./helm/core --set global.o11y.promtail=null` still renders. `helm lint` clean.

Also verified end-to-end on a live cluster: gateway JSON access log → Promtail sidecar → externally-deployed Loki → queried in Grafana.

- Both new values default to empty/`false` → **zero behavioral change** for existing installs.
- Scope is intentionally minimal. Loki authentication (basic/bearer) and multi-tenancy (`X-Scope-OrgID`) are **out of scope** here (single-tenant, unauthenticated Loki) and can be a follow-up — documented in the `lokiUrl` values comment.

### DIFF
--- a/helm/core/templates/_pod.tpl
+++ b/helm/core/templates/_pod.tpl
@@ -203,11 +203,11 @@ template:
         - mountPath: /opt/plugins
           name: local-wasmplugins-volume
         {{- end }}
-        {{- if $o11y.enabled }}
+        {{- if or $o11y.enabled (dig "promtail" "enabled" false $o11y) }}
         - mountPath: /var/log/proxy
           name: log
         {{- end }}
-      {{- if $o11y.enabled }}
+      {{- if or $o11y.enabled (dig "promtail" "enabled" false $o11y) }}
         {{- $config := $o11y.promtail }}
       - name: promtail
         image: {{ $config.image.repository | default (printf "%s/higress/promtail" .Values.global.hub) }}:{{ $config.image.tag }}
@@ -295,7 +295,7 @@ template:
       emptyDir: {}
     - name: proxy-socket
       emptyDir: {}
-    {{- if $o11y.enabled }}
+    {{- if or $o11y.enabled (dig "promtail" "enabled" false $o11y) }}
     - name: log
       emptyDir: {}
     - name: tmp
--- a/helm/core/templates/configmap.yaml
+++ b/helm/core/templates/configmap.yaml
@@ -3,7 +3,7 @@
     # Refer to https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md#21-trust-domain
     trustDomain: "cluster.local"
     accessLogEncoding: TEXT
-    {{- if .Values.global.o11y.enabled }}
+    {{- if or .Values.global.o11y.enabled (dig "promtail" "enabled" false .Values.global.o11y) }}
     accessLogFile: "/var/log/proxy/access.log"
     {{- else }}
     accessLogFile: "/dev/stdout"
--- a/helm/core/templates/promtail.yaml
+++ b/helm/core/templates/promtail.yaml
@@ -1,5 +1,5 @@
 {{- $o11y := .Values.global.o11y  }}
-{{- if $o11y.enabled }}
+{{- if or $o11y.enabled (dig "promtail" "enabled" false $o11y) }}
 {{- $config := $o11y.promtail }}
 apiVersion: v1
 kind: ConfigMap
@@ -13,7 +13,7 @@ data:
       http_listen_port: {{ $config.port }}
 
     clients:
-    - url: http://higress-console-loki.{{ .Release.Namespace }}:3100/loki/api/v1/push
+    - url: {{ $config.lokiUrl | default (printf "http://higress-console-loki.%s:3100/loki/api/v1/push" .Release.Namespace) }}
 
     positions:
       filename: /tmp/promtail-positions.yaml
--- a/helm/core/values.yaml
+++ b/helm/core/values.yaml
@@ -378,6 +378,17 @@ global:
   o11y:
     enabled: false
     promtail:
+      # -- Inject the Promtail sidecar and write the gateway access log to a file,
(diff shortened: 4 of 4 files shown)


CH3CHO:
I would prefer the o11y suite is a whole. If you have Promtail installed externally, would it be better to have Prometheus and Grafana installed externally as well?

PR #4135, opened 2026-07-13
test: add Gateway API v1.4 conformance CI

EndlessSeeker:
## Summary

- run the upstream Gateway API v1.4.0 Conformance Suite as the default `gateway-conformance-test` PR job
- cover the required `GATEWAY-HTTP` Core profile (`Gateway`, `HTTPRoute`, and `ReferenceGrant`)
- reuse the upstream embedded tests and manifests; Higress owns only the runner, kind lifecycle, network adapter, diagnostics, and CI integration
- fix controller behavior exposed by the Core suite: GatewayClass selection, listener/route hostname intersection, Kubernetes Service existence/lifecycle validation, and shared gateway address lookup

### GatewayClass reconciliation

Previously, Higress reconciled only the GatewayClass whose `metadata.name` matched the configured GatewayClass name, even when another GatewayClass used the same Higress `spec.controllerName`.

After this change, Higress reconciles every GatewayClass whose `spec.controllerName` matches the active Higress controller:

- default configuration: `higress.io/gateway-controller`
- when `global.gatewayClass=<name>` is configured: `higress.io/gateway-controller-<name>`

The configured/default GatewayClass is still created as before. The change allows multiple GatewayClass aliases to select the same Higress controller, as required by the Gateway API conformance tests.

### Kubernetes Service backend validation

Higress keeps its existing service-registry lookup as the primary resolution path. It also observes the Kubernetes Service informer so that a referenced backend can transition from `BackendNotFound` to resolved when the Service is created, and back to `BackendNotFound` when it is deleted. The informer dependency is retained after successful resolution so deletion triggers Route recomputation.

## Upgrade notice

Before upgrading, list GatewayClasses that use the active Higress `spec.controllerName`. A GatewayClass with a different name but the same controller name was previously ignored; after upgrading, it and the Gateways/Routes that reference it will be reconciled and may begin affecting status and data-plane configuration.

If such a GatewayClass must remain unmanaged by Higress, change its `spec.controllerName` or remove/migrate the associated resources before upgrading. Existing resources that use the configured/default GatewayClass keep their existing selection behavior.

Example audit for the default controller:

```bash
kubectl get gatewayclass -o jsonpath='{range .items[?(@.spec.controllerName=="higress.io/gateway-controller")]}{.metadata.name}{"\n"}{end}'
```

## Conformance scope

The CI job installs Higress in `gateway-conformance-infra` and runs the upstream v1.4.0 `GATEWAY-HTTP` Core profile. It does not claim that the suite alone proves Higress's production `higress-system` shared-gateway, cross-namespace TLS Secret data-plane path; that topology requires separate end-to-end coverage.

The controller image is built locally from the PR head SHA and loaded into kind. The Pilot and gateway images use the repository-wide fixed tags from `Makefile.core.mk`.

## Verification

- `go test ./pkg/ingress/kube/gateway/... -count=1`
- `go test ./test/gateway -run '^$' -count=1`
- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.7 .github/workflows/build-and-test.yaml`
- `TestHTTPRouteBackendServiceLifecycle`: missing -> created -> deleted, with `ResolvedRefs` transitions verified
- PR CI [`gateway-conformance-test`](https://github.com/higress-group/higress/actions/runs/29324817238/job/87059909531): 33 passed, 0 failed, 0 skipped
- PR CI [`higress-conformance-test`](https://github.com/higress-group/higress/actions/runs/29324817238/job/87059909552): passed after rebasing onto current `main`

The current shared gateway image is amd64-only. The default GitHub Actions job and the verified ACK environment are amd64.

Signed-off-by: EndlessSeeker <1766508902@qq.com>


johnlanni (approved):
LGTM

PR #4137, opened 2026-07-14
feat: add configurable AOF and RDB persistence for Redis

geekspeng:

Adds configurable AOF and RDB persistence settings for Redis when `persistence.enabled=true`.

### Behavior Matrix

| `persistence.enabled` | `aof.enabled` | `rdb.enabled` | Generated Config |
|--------|---------|---------|---------|
| `true` (default) | `false` (default) | `false` (default) | empty config (equivalent to `persistence.enabled=false`; the PVC is still mounted) |
| `true` | `true` | `true` | `dir /data` + AOF + RDB (dual persistence) |
| `true` | `false` | `true` | `dir /data` + RDB-only (snapshots) |
| `true` | `true` | `false` | `dir /data` + AOF-only |
| `true` | `true` | `true`, with `rdb.save=[]` | `dir /data` + AOF + `save ""` (RDB snapshots disabled) |

### New Configuration Options

**AOF Settings:**
- `aof.enabled` (default: false, opt-in) - Enable AOF persistence
- `aof.fsync` (default: "everysec") - AOF fsync policy: always, everysec, or no (validated at render time)
- `aof.rewritePercentage` (default: 100) - Auto-aof-rewrite-percentage
- `aof.rewriteMinSize` (default: "64mb") - Auto-aof-rewrite-min-size

**RDB Settings:**
- `rdb.enabled` (default: false, opt-in) - Enable RDB snapshots
- `rdb.save` (default: ["900 1", "300 10", "60 10000"]) - Snapshot schedule; an empty list `[]` renders `save ""`, which disables RDB snapshots

### Key Changes

- Add `aof.*` configuration block for fine-grained AOF control
- Add `rdb.*` configuration block for RDB snapshots
- AOF and RDB are **independent** - can enable both for dual persistence
- **Opt-in defaults**: `aof.enabled` and `rdb.enabled` default to `false` so existing `persistence.enabled=true` users keep their previous behavior (per review feedback)
- `dir /data` is rendered only when `aof.enabled` or `rdb.enabled` is true; when both are disabled, persistence can be driven via `extraConfig` (add `dir /data` there)
- `rdb.save=[]` renders `save ""` (the canonical Redis way to disable snapshots)
- `aof.fsync` is validated at render time against {always, everysec, no}
- Add trailing newline to values.yaml, fix extraConfig rendering to avoid blank lines
- Update README with new parameter documentation

No specific issue. Extends the original PR scope to provide flexible persistence configuration. This update addresses the review findings on this PR (opt-in defaults, `rdb.save=[]` behavior, `aof.fsync` validation, `dir /data` rendering condition).

This is a Helm template (YAML) change with no Go code. Verified by rendering the chart with `helm template` across all configuration combinations.

```bash
# Default: persistence enabled, AOF/RDB opt-in off -> empty config
helm template r ./helm/core/charts/redis --set redis.persistence.enabled=true --show-only templates/configmap.yaml

# Dual persistence (AOF + RDB)
helm template r ./helm/core/charts/redis --set redis.persistence.enabled=true --set redis.aof.enabled=true --set redis.rdb.enabled=true --show-only templates/configmap.yaml

# RDB-only
helm template r ./helm/core/charts/redis --set redis.persistence.enabled=true --set redis.rdb.enabled=true --show-only templates/configmap.yaml

# AOF-only
helm template r ./helm/core/charts/redis --set redis.persistence.enabled=true --set redis.aof.enabled=true --show-only templates/configmap.yaml

# Disable RDB snapshots via empty save list (renders `save ""`)
helm template r ./helm/core/charts/redis --set redis.persistence.enabled=true --set redis.aof.enabled=true --set redis.rdb.enabled=true --set-json 'redis.rdb.save=[]' --show-only templates/configmap.yaml

# Invalid aof.fsync fails at render time
helm template r ./helm/core/charts/redis --set redis.persistence.enabled=true --set redis.aof.enabled=true --set redis.aof.fsync=typo --show-only templates/configmap.yaml

# Parent chart path
helm template r ./helm/core --set global.enableRedis=true --set redis.redis.persistence.enabled=true --set redis.redis.aof.enabled=true --set redis.redis.rdb.enabled=true --show-only charts/redis/templates/configmap.yaml
```

- **Opt-in persistence**: `aof.enabled`/`rdb.enabled` default to `false`. Users who want persistence must enable at least one; this keeps the upgrade path safe for existing `persistence.enabled=true` deployments (addressing the HIGH review finding).
- **Dual persistence**: enable both for AOF (durability, ≤1s data loss with `everysec`) + RDB (fast restart, compact backups). Redis prefers AOF for recovery when both are enabled.
- **Default values**: all settings use Redis official defaults (`appendfsync everysec`, `save 900 1 300 10 60 10000`, etc.)
- **Backward compatible**: default behavior unchanged (`persistence.enabled: false` renders the same config as before; `persistence.enabled: true` no longer implicitly enables AOF/RDB)

## Ⅵ. AI Coding Tool Usage Checklist

This PR was produced with the assistance of **Claude Code** (Anthropic).

### AI Coding Summary

- **Key decisions**: (a) Make AOF and RDB independent toggles instead of mutually exclusive. (b) Default both to `false` (opt-in) per review feedback so existing `persistence.enabled=true` users keep their behavior. (c) Add fine-grained control over AOF fsync and rewrite settings. (d) Fix trailing newline and extraConfig rendering issues.
- **Review fixes** (commit 0eda200f): flipped `aof.enabled`/`rdb.enabled` defaults to `false`; render `save ""` when `rdb.save` is empty so documented RDB disabling works; added `fail` guard validating `aof.fsync`; skip `dir /data` when both AOF and RDB are disabled; synced comments and the higress README parameter table.
- **Major changes**: Added `aof.*` and `rdb.*` configuration blocks in values.yaml, updated configmap.yaml template, added README documentation.
- **Testing**: Verified all configuration combinations via `helm template` (sub-chart and parent chart paths), including the fsync fail guard and empty-save rendering; `helm lint` passes.


geekspeng:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Verified empirically with `helm template`: with `redis.redis.persistence.enabled=true` the rendered `redis-stack.conf` contains `dir /data`, the AOF directives (`appendonly`, `appendfsync`, rewrite thresholds), each `save` rule on its own line, and multiline `extraConfig` via `nindent` renders correctly; the PVC and `/data` volumeMount line up, and `helm lint` passes. Behavior with persistence disabled is unchanged.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

helm/core/charts/redis/templates/configmap.yaml
```diff
@@ -7,4 +7,21 @@ data:
   redis-stack.conf: |
     {{- if .Values.redis.password }}
     requirepass {{ .Values.redis.password }}
-    {{- end }}
\ No newline at end of file
+    {{- end }}
+    {{- if .Values.redis.persistence.enabled }}
```
CH3CHO, line 11: 🔴 **HIGH: Backward-incompatible behavior change for existing `persistence.enabled=true` users**

The PR description claims *"Backward compatible: Default behavior unchanged (`persistence.enabled: false`)"*, but this is misleading. The default `persistence.enabled: false` is preserved, but for users who have **already set `persistence.enabled=true`**, this PR silently flips on AOF and RDB by default:

- **Before**: `persistence.enabled=true` → only `dir /data` in config, **no persistence strategy**
- **After**: `persistence.enabled=true` → `dir /data` + AOF + RDB both enabled

Upgrade impact on existing deployments:
1. AOF file begins growing immediately (`appendonly yes`)
2. RDB snapshots fire on the default schedule (`save 900 1`, etc.)
3. Default PVC size is **1Gi** — AOF + RDB can fill this quickly under any non-trivial write load
4. I/O pattern changes (write amplification from `appendfsync everysec`)
5. First AOF rewrite triggers when file exceeds `rewriteMinSize=64mb`

**Recommended fix (pick one)**:
1. **Flip defaults to opt-in**: set `aof.enabled: false` and `rdb.enabled: false` in both `helm/core/charts/redis/values.yaml` and `helm/core/values.yaml`. This is the safest approach for an existing deployment base.
2. **Document as breaking change**: add a prominent "⚠️ Breaking change" callout to the PR description, release notes, and a new "Upgrading" section in `helm/higress/README.md`.

Also worth noting: in the behavior matrix, case 4 (`persistence=true, aof=false, rdb=false` → only `dir /data`) is functionally equivalent to `persistence.enabled=false` and just mounts an unused PVC — consider documenting this or removing the case.

helm/core/charts/redis/templates/configmap.yaml
```diff
@@ -7,4 +7,21 @@ data:
...
+    dir /data
+    {{- if .Values.redis.aof.enabled }}
+    appendonly yes
+    appendfsync {{ .Values.redis.aof.fsync }}
```
CH3CHO, line 20: 🟢 **LOW: No validation for `aof.fsync` value**

The template renders `appendfsync {{ .Values.redis.aof.fsync }}` without validating the value. If a user sets `aof.fsync: everysec_typo` or any invalid string, Redis will fail to start with a confusing error like *"Bad directive or wrong number of arguments"* — far from the values.yaml they just edited.

Same applies to `rdb.save` list elements (no format check on the `"<seconds> <changes>"` string).

**Recommended fix**: add `values.schema.json` to `helm/core/charts/redis/` validating `aof.fsync` against the set `{always, everysec, no}`, or add a `fail` guard in the template:

```yaml
{{- if not (or (eq .Values.redis.aof.fsync "always") (eq .Values.redis.aof.fsync "everysec") (eq .Values.redis.aof.fsync "no")) }}
{{- fail (printf "redis.aof.fsync must be one of: always, everysec, no (got: %q)" .Values.redis.aof.fsync) }}
{{- end }}
```

helm/core/charts/redis/values.yaml
```diff
@@ -46,4 +46,37 @@ redis:
...
+  # -- RDB (Redis Database) snapshot persistence settings. Only effective when persistence.enabled=true.
+  # RDB provides compact point-in-time snapshots, useful for backups and faster restart.
+  # Can be enabled alongside AOF for dual persistence strategy.
+  rdb:
+    # -- Enable RDB snapshots. Default is true. Set to false if you prefer AOF-only persistence.
+    enabled: true
+    # -- RDB snapshot schedule as "seconds changes" pairs. Redis saves when ANY rule is matched.
+    # Common defaults: "900 1" (15min/1 key), "300 10" (5min/10 keys), "60 10000" (1min/10000 keys).
+    # Set to empty list [] to disable RDB. See https://redis.io/docs/management/persistence/
+    save:
```
CH3CHO, line 76: 🟡 **MEDIUM: `rdb.save=[]` does NOT disable RDB as the comment claims**

The comment says: *"Set to empty list [] to disable RDB"*, but the template renders nothing when the list is empty:

```yaml
{{- range .Values.redis.rdb.save }}
save {{ . }}
{{- end }}
```

When `rdb.enabled=true` and `rdb.save=[]`, **no `save` directive is written to redis-stack.conf**. Redis treats this as "use defaults", which means RDB snapshots **still happen** on the built-in schedule (`save 3600 1`, `save 300 100`, `save 60 10000`). So the documented way to disable RDB does not actually disable it.

**Recommended fix (pick one)**:

1. **Fix the template** to emit `save ""` when the list is empty (this is the canonical way to disable RDB snapshots in Redis):
```yaml
{{- if .Values.redis.rdb.enabled }}
{{- if eq (len .Values.redis.rdb.save) 0 }}
save ""
{{- else }}
{{- range .Values.redis.rdb.save }}
save {{ . }}
{{- end }}
{{- end }}
{{- end }}
```

2. **Fix the comment** to reflect reality: *"Set to empty list [] to reset to Redis default save schedule (3600 1, 300 100, 60 10000). To fully disable RDB, set rdb.enabled=false."*

helm/core/charts/redis/templates/configmap.yaml
```diff
@@ -7,4 +7,21 @@ data:
   redis-stack.conf: |
     {{- if .Values.redis.password }}
     requirepass {{ .Values.redis.password }}
-    {{- end }}
\ No newline at end of file
+    {{- end }}
+    {{- if .Values.redis.persistence.enabled }}
+    dir /data
```
CH3CHO, line 13: 🟢 **LOW: Behavior matrix case 4 is functionally equivalent to `persistence.enabled=false`**

When `persistence.enabled=true` and both `aof.enabled=false` and `rdb.enabled=false`, the template renders only:
```
dir /data
```

But the StatefulSet still mounts the PVC at `/data` (since `persistence.enabled=true`). This combination just allocates a PVC and mounts it without any persistence strategy — data goes to RAM only and is lost on restart. It is functionally identical to `persistence.enabled=false`, except for the wasted disk allocation.

**Recommendation**: either remove this case from the behavior matrix in the PR description, or render nothing in the template when both AOF and RDB are disabled (skip the `dir /data` line as well). For example:

```yaml
{{- if and .Values.redis.persistence.enabled (or .Values.redis.aof.enabled .Values.redis.rdb.enabled) }}
dir /data
{{- if .Values.redis.aof.enabled }}
...
{{- end }}
{{- if .Values.redis.rdb.enabled }}
...
{{- end }}
{{- end }}
```

This makes the matrix case 4 collapse to "same as persistence.enabled=false", which is the principle of least surprise.

CH3CHO (changes requested):
## Code Review Summary

Thanks for the clear PR description and the behavior matrix — that made it easy to verify the intent. The implementation works as designed and the `helm template` verification is thorough.

However, the PR is essentially a **breaking change for anyone running with `persistence.enabled=true`** (default-on AOF+RDB), and one of the new `rdb.save=[]` defaults does not behave as documented. Both should be addressed before merge.

### Findings (in order of severity)

🔴 **HIGH — Backward-incompatible default behavior change**
`persistence.enabled=true` users will silently flip on AOF + RDB on upgrade, with `1Gi` default PVC. See [discussion](https://github.com/higress-group/higress/pull/4137#discussion_r3879095542).

🟡 **MEDIUM — `rdb.save=[]` does not disable RDB**
The doc comment promises one thing, the template delivers another (Redis falls back to built-in save defaults). See [discussion](https://github.com/higress-group/higress/pull/4137#discussion_r3879102827).

🟢 **LOW — No validation for `aof.fsync`**
A typo silently produces an invalid `appendfsync` directive that fails Redis startup. See [discussion](https://github.com/higress-group/higress/pull/4137#discussion_r3879096690).

🟢 **LOW — Behavior matrix case 4 is functionally equivalent to `persistence.enabled=false`**
Mounts a PVC that nothing writes to. See [discussion](https://github.com/higress-group/higress/pull/4137#discussion_r3879104880).

### Things that look good ✅

- `dir /data` correctly aligns with the volumeMount in `statefulset.yaml`
- AOF and RDB are independent toggles (not mutually exclusive as in the original chart)
- Defaults match Redis upstream recommendations (`everysec`, `64mb`, `100`, `900 1/300 10/60 10000`)
- `extraConfig` is appended correctly via `nindent`
- `redis.redis.*` in parent `values.yaml` correctly reaches the sub-chart without alias remapping

### Recommendation

**Request changes** until HIGH and MEDIUM are addressed. The HIGH issue is the most important: I'd lean toward flipping `aof.enabled` and `rdb.enabled` defaults to `false` so existing `persistence.enabled=true` users keep their previous behavior. New users who want dual persistence can opt-in explicitly.

helm/core/charts/redis/templates/configmap.yaml
geekspeng: Fixed in 0eda200f: flipped `aof.enabled`/`rdb.enabled` defaults to `false` (opt-in) in both `helm/core/charts/redis/values.yaml` and `helm/core/values.yaml`, so existing `persistence.enabled=true` users keep their previous behavior. The PR description no longer claims the old default-on behavior was backward compatible, and the behavior matrix documents the new default (empty config with the PVC still mounted).

helm/core/charts/redis/values.yaml
geekspeng: Fixed in 0eda200f: when `rdb.enabled=true` and `rdb.save` is an empty list, the template now renders `save ""` (the canonical Redis directive to disable snapshotting), so the documented empty-list behavior now matches what is generated.

helm/core/charts/redis/templates/configmap.yaml
geekspeng: Fixed in 0eda200f: added a `fail` guard validating `aof.fsync` against {always, everysec, no} — an invalid value now fails at render time with a message naming the offending value (`redis.aof.fsync must be one of: always, everysec, no (got: "typo")`). Went with the template guard rather than `values.schema.json` since the repo has no values.schema.json precedent yet.

helm/core/charts/redis/templates/configmap.yaml
geekspeng: Fixed in 0eda200f: `dir /data` is now rendered only when `aof.enabled` or `rdb.enabled` is true, so this combination collapses to the same config as `persistence.enabled=false` (the behavior matrix in the PR description was updated accordingly). The values.yaml comments document that driving persistence via `extraConfig` with both toggles off requires adding `dir /data` to `extraConfig` yourself.

helm/core/charts/redis/templates/configmap.yaml
CH3CHO: Verified locally with `helm template`: defaults are now opt-in (`aof.enabled: false`, `rdb.enabled: false` in both values.yaml files and synced in helm/higress/README.md). With `persistence.enabled=true` and no extra flags, the rendered config is empty as expected. ✅ Resolved.

helm/core/charts/redis/templates/configmap.yaml
CH3CHO: Verified: `helm template ... --set redis.aof.fsync=typo` now fails at render time with `redis.aof.fsync must be one of: always, everysec, no (got: "typo")`. The template guard is the right call given there is no `values.schema.json` precedent in the chart. ✅ Resolved.

helm/core/charts/redis/values.yaml
CH3CHO: Verified: `--set-json redis.rdb.save=[]` now renders `save ""` (canonical Redis way to disable snapshots). The comment in values.yaml was also clarified. ✅ Resolved.

helm/core/charts/redis/templates/configmap.yaml
CH3CHO: Verified: with `persistence.enabled=true` and both AOF/RDB disabled, the rendered configmap is empty (`dir /data` is now gated behind `or aof.enabled rdb.enabled`). The behavior matrix case 4 collapses as suggested. ✅ Resolved.

CH3CHO (approved):
All four review comments have been verified against commit 359e1048:

- `aof.enabled`/`rdb.enabled` defaults flipped to `false` (opt-in) in both `helm/core/charts/redis/values.yaml` and `helm/core/values.yaml`, with the README parameter table in sync.
- `rdb.save=[]` now renders the canonical `save ""` directive, and the values.yaml comment was clarified.
- `aof.fsync` is validated by a `{{- fail }}` guard at render time; `helm template ... --set redis.aof.fsync=typo` aborts with a clear error.
- `dir /data` is gated behind `(or aof.enabled rdb.enabled)`, so behavior-matrix case 4 collapses as expected.

Verified all behavior-matrix combinations via `helm template` (sub-chart and parent chart paths); `helm lint` passes. The behavior is now safe to upgrade for existing `persistence.enabled=true` users. 👍

PR #4142, opened 2026-07-15
feat: support Gateway API v1.6.0

EndlessSeeker:
## Summary

- upgrade the Higress production module to `sigs.k8s.io/gateway-api v1.6.0`
- update the Higress Istio fork to Gateway API v1.6.0 through merged [higress-group/istio#63](https://github.com/higress-group/istio/pull/63), then combine it with merged GIE support from [higress-group/istio#67](https://github.com/higress-group/istio/pull/67), pinned to latest `istio-1.27` commit `d17cf07b810bb877e93d87455a1410b8269e3173`
- use the v1.6 API types and semantics in the production Gateway API reconciliation and translation path
- provision an isolated Deployment and Service for each managed Gateway when the opt-in deployment controller is enabled
- keep the Gateway API v1.4.0 and v1.6.0 conformance dependencies isolated in independent Go submodules
- run the HTTP Core conformance profile for both versions in CI

## Why the Helm changes are required

Gateway API v1.6 HTTP Core includes multiple-Gateway isolation and listener-port behavior that cannot be implemented by routing every Gateway through the existing shared wildcard Service.

With the per-Gateway deployment controller removed, `HTTPRouteMultipleGateways` cross-routes traffic and the arbitrary listener-port case has no matching Service port. Enabling the controller makes each Gateway own a label-isolated Deployment and a Service whose ports follow its listeners.

The Helm additions are therefore product/runtime support, not ACK or local Kind scaffolding:

- an opt-in `global.enableGatewayAPIDeploymentController` value (default `false`)
- the Istio gateway injection template and values needed to create Higress gateway workloads
- the native Istio `PILOT_ENABLE_GATEWAY_API_DEPLOYMENT_CONTROLLER` environment flag; no duplicate Higress-specific flag is introduced
- RBAC for the resources watched by Istio's deployment-controller informers; without HPA/PDB list/watch access, Pilot never becomes ready

The conformance installation enables this opt-in value. Existing installations retain the shared gateway model by default.

The independent Helm value cannot reuse `global.enableGatewayAPI`: that existing value defaults to `true` and controls whether Gateway API resources are watched and translated, not whether every Gateway provisions workloads. Reusing it would silently change existing installations from the shared deployment model. `ENABLE_GATEWAY_API_MANUAL_DEPLOYMENT` is also a different capability: it permits binding a Gateway to an existing workload instead of provisioning one.

At runtime, Higress does reuse Istio's existing deployment-controller feature flag. Because upstream Istio defaults that flag to `true`, Higress additionally requires the environment variable to be explicitly set; this preserves Higress's default shared deployment model for non-Helm startup paths.

## Dependency and test isolation

- root `go.mod`: `sigs.k8s.io/gateway-api v1.6.0`
- Istio fork: Gateway API and conformance dependencies at `v1.6.0`
- `test/gateway/common`: shared dialer only, with no Gateway API dependency
- `test/gateway/v1.4`: standalone module pinned to Gateway API `v1.4.0`
- `test/gateway/v1.6`: standalone module pinned to conformance `v1.6.0`
- each CI matrix entry installs only its matching standard CRD bundle and runs only its matching test module

This prevents Go module resolution or installed CRDs from silently mixing v1.4 and v1.6.

## Advertised conformance scope

This PR advertises and runs only the minimum existing Core scope:

- profile: `GATEWAY-HTTP`
- features: `Gateway`, `HTTPRoute`, `ReferenceGrant`
- v1.4.0: 33 passed, 0 failed, 0 skipped
- v1.6.0: 37 passed, 0 failed, 0 skipped

TLS, GRPC, TCP, UDP, and experimental profiles are intentionally not part of this PR's conformance claim.

## Validation

- `go test ./pkg/ingress/kube/gateway/... -count=1`
- `go test -C test/gateway/common ./... -count=1`
- `go test -C test/gateway/v1.4 ./... -run '^$' -count=1`
- `go test -C test/gateway/v1.6 ./... -run '^$' -count=1`
- `helm lint helm/core`
- Helm rendering with the deployment controller enabled and with its default disabled value
- local official Gateway API v1.4.0 HTTP Core report: 33/33
- local official Gateway API v1.6.0 HTTP Core report: 37/37

The CI jobs build the Pilot image from the checked-out Istio submodule. This is required because the pre-existing published Pilot does not match the upgraded controller dependencies and caused the existing Higress conformance controller to remain unready.

## Conflict resolution and final CI validation for `1f310360`

The conflict with current `main` came from merged GIE v1.4 support changing both `istio/istio` and the InferencePool conversion test data. The resolution preserves both sides:

- `istio/istio` now points to `d17cf07b810bb877e93d87455a1410b8269e3173`, the latest `istio-1.27` commit containing Gateway API v1.6, GIE v1.4, and subsequent security fixes
- `conversion_test.go` retains the managed-Gateway coverage from this PR and the multi-port InferencePool coverage from `main`

Final clean-run reports for Higress commit `1f3103609843b8fc39d5e638f3e9e2f197bb129e`:

- [Gateway API v1.4.0 HTTP Core](https://github.com/higress-group/higress/actions/runs/31350097560/job/93339748063): 33 passed, 0 failed, 0 skipped
- [Gateway API v1.6.0 HTTP Core](https://github.com/higress-group/higress/actions/runs/31350097560/job/93339748071): 37 passed, 0 failed, 0 skipped
- [Gateway API Inference Extension v1.4.0 Core](https://github.com/higress-group/higress/actions/runs/31350097560/job/93339748057): 12 passed, 0 failed, 0 skipped
- Build, coverage, and lint checks in the same workflow passed


hansedong:
Any progress please?

EndlessSeeker:
@hansedong Thanks for following up. This feature will be included in the next Higress release.

johnlanni (approved):
LGTM

PR #4153, opened 2026-07-16
fix(jwt-auth): harden cookie token parsing

EndlessSeeker:

Harden `jwt-auth` cookie token parsing by skipping malformed cookie segments that do not contain `=`.

This mirrors the cookie parsing robustness fix previously applied to `cluster-key-rate-limit`:

- use `strings.SplitN(pair, "=", 2)` so cookie values containing `=` are preserved
- check `len(kv) == 2` before reading the cookie value
- add regression coverage for malformed cookie segments, empty cookies, missing keys, and values containing `=`

No public issue.

Added unit tests for `findCookie`.

```bash
cd plugins/wasm-go/extensions/jwt-auth
go test ./handler
```

This is a robustness bugfix for malformed Cookie headers in `jwt-auth` when `from_cookies` is configured.


johnlanni (approved):
LGTM

PR #4154, opened 2026-07-16
bugfix: map Claude adaptive effort to Bedrock output config

### LINKED ISSUE #4170: Claude adaptive effort is nested incorrectly in Bedrock Converse requests
- [x] I have searched the issues of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description

When `ai-proxy` converts a Claude request with adaptive thinking and an output effort level to an Amazon Bedrock Converse request, it places `effort` inside the `thinking` object.

Bedrock expects `thinking` and `output_config` to be sibling fields in `additionalModelRequestFields`. The current request shape does not match the Bedrock model request schema for adaptive thinking.

### Ⅱ. Describe what happened

Given an Anthropic-compatible request containing:

```json
{
  "thinking": {
    "type": "adaptive"
  },
  "output_config": {
    "effort": "high"
  }
}
```

`ai-proxy` currently produces Bedrock additional model request fields equivalent to:

```json
{
  "thinking": {
    "type": "adaptive",
    "effort": "high"
  }
}
```

The original `output_config` structure is lost and `effort` is sent at the wrong nesting level.

The Bedrock request should preserve the two sibling fields:

```json
{
  "thinking": {
    "type": "adaptive"
  },
  "output_config": {
    "effort": "high"
  }
}
```

Supported effort values should remain `low`, `medium`, and `high`. Unsupported values should continue to be omitted.

1. Configure the `ai-proxy` plugin with an Amazon Bedrock provider using the Converse API path.
2. Send a Claude-compatible request with `thinking.type` set to `adaptive` and `output_config.effort` set to `high`.
3. Inspect the generated Bedrock request.
4. Observe that `effort` is nested inside `thinking` and that no sibling `output_config` field is present.

The problem is limited to adaptive thinking effort mapping. Manual extended thinking with `budget_tokens`, unsupported effort values, and requests without adaptive thinking should remain unaffected.

### Ⅵ. Environment

- Higress version: current `main` before #4154
- Component: `plugins/wasm-go/extensions/ai-proxy`
- Provider: Amazon Bedrock Converse

storyicon:

This PR fixes the Bedrock Converse request mapping for Claude adaptive thinking.

Previously, the Anthropic-compatible request:

```json
{
  "thinking": {
    "type": "adaptive"
  },
  "output_config": {
    "effort": "high"
  }
}
```

was converted to:

```json
{
  "additionalModelRequestFields": {
    "thinking": {
      "type": "adaptive",
      "effort": "high"
    }
  }
}
```

Bedrock expects `thinking` and `output_config` to be sibling fields. This PR now generates:

```json
{
  "additionalModelRequestFields": {
    "thinking": {
      "type": "adaptive"
    },
    "output_config": {
      "effort": "high"
    }
  }
}
```

This keeps the original adaptive thinking configuration while placing `effort` in the Bedrock-compatible location.

Closes #4170

A regression test is included.

The test verifies that:

- `thinking.type` remains `adaptive`
- `effort` is not added to the `thinking` object
- `output_config.effort` is emitted as a sibling model request field
- unsupported effort values and effort without adaptive thinking remain omitted

Run the provider unit tests:

```shell
go test ./provider -count=1
```

Run the Bedrock plugin test:

```shell
go test . -run '^TestBedrock$'
```

Both commands should pass.

- The existing supported effort values remain unchanged: `low`, `medium`, and `high`.
- Unsupported effort values are still omitted.
- Manual extended thinking with `budget_tokens` is unaffected.
- Bedrock structured output configuration is unaffected.
- No fallback behavior is introduced.

### DIFF
--- a/plugins/wasm-go/extensions/ai-proxy/provider/bedrock.go
+++ b/plugins/wasm-go/extensions/ai-proxy/provider/bedrock.go
@@ -963,10 +963,12 @@ func (b *bedrockProvider) buildBedrockTextGenerationRequest(origRequest *chatCom
 
 	thinking := bedrockThinkingFromClaudeConfig(origRequest.ClaudeThinking)
 	if thinking != nil {
-		if origRequest.ClaudeThinking.Type == "adaptive" && origRequest.ClaudeOutputConfig != nil && bedrockSupportsAdaptiveEffort(origRequest.ClaudeOutputConfig.Effort) {
-			thinking["effort"] = origRequest.ClaudeOutputConfig.Effort
-		}
 		request.AdditionalModelRequestFields["thinking"] = thinking
+		if origRequest.ClaudeThinking.Type == "adaptive" && origRequest.ClaudeOutputConfig != nil && bedrockSupportsAdaptiveEffort(origRequest.ClaudeOutputConfig.Effort) {
+			request.AdditionalModelRequestFields["output_config"] = map[string]interface{}{
+				"effort": origRequest.ClaudeOutputConfig.Effort,
+			}
+		}
 	} else if origRequest.ReasoningEffort != "" {
 		thinkingBudget := 1024 // default
 		switch origRequest.ReasoningEffort {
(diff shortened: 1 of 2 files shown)


johnlanni (approved):
LGTM. The fix preserves `thinking.type=adaptive` while moving the supported effort value to the Bedrock-required sibling `additionalModelRequestFields.output_config.effort`. Unsupported effort values and non-adaptive/manual thinking paths remain unchanged.

Verified at head `9edb3d73` against current `origin/main`:
- focused Bedrock thinking/effort tests pass;
- full `go test ./provider -count=1` passes;
- `gofmt` and `git diff --check` are clean.

The branch is behind `main`, but GitHub reports it mergeable and the current patch applies cleanly, so this does not block the review.

PR #4178, opened 2026-07-21
fix(ai-data-masking): avoid loading jieba default dict to prevent OOM

### LINKED ISSUE #4157: ai-data-masking 插件开启后立即 OOM 崩溃 || ai-data-masking OOM crashes immediately after the plug-in is enabled
运行环境：
2 核 2G 的云服务器

 ## 问题描述                                                                                                                                                                
  开启 ai-data-masking Wasm 插件后，Envoy 进程内存立即暴涨，3-5 秒内被 Linux OOM Killer 杀死，全程不需要任何请求流量。
                                                                                                                                                        
  ## 复现步骤                                                                                                                                                                
  1. 使用 Higress All-in-One Docker 镜像启动                                                                                                                                 
  2. 只开启 ai-proxy 插件，Envoy 稳定运行，内存稳定在 ~480MB                                                                                                                 
  3. 在控制台开启 ai-data-masking 插件（空配置、默认配置都一样）                                                                                                             
  4. 观察 Envoy 进程内存：`watch -n 1 "ps aux | grep [e]nvoy | awk '{printf \"PID: %s  CPU: %.1f%%  MEM: %.1f%%  %.0f MB\\n\", \$2, \$3, \$4, \$6/1024}'"` 
 
                                                                                                                
                                                                                                                                                                             
  ## 实际结果                                                                                                                                                                
  - 开启插件后，内存从 482MB 开始暴涨                                                                                                                                        
  - 每秒增长 ~30MB，3-5 秒后达到 648MB                                                                                                                                       
  - Linux OOM Killer 杀死 Envoy 进程                                                                                                                                         
  - 日志输出：`Envoy exited with error: signal: killed`  

<img width="1192" height="386" alt="Image" src="https://github.com/user-attachments/assets/771bd3fa-a8d5-4bf8-81fa-dcc8988e66b7" />

## 我的想法
能否优化一下插件内存占用，或者在Envoy崩溃的时候提醒用户，不然最终的效果是：我开启插件后，ai 网管奔溃，无法接受任何流量
<!--This is a translation content dividing line, the content below is generated by machine, please do not modify the content below-->
---
Operating environment:
2-core 2G cloud server

 ## Problem description
  After turning on the ai-data-masking Wasm plug-in, the Envoy process memory immediately skyrocketed and was killed by the Linux OOM Killer within 3-5 seconds without any request traffic.
                                                                                                                                                        
  ## Reproduction steps
  1. Start using the Higress All-in-One Docker image
  2. Only enable the ai-proxy plug-in, Envoy runs stably, and the memory is stable at ~480MB
  3. Enable the ai-data-masking plug-in in the console (the empty configuration and the default configuration are the same)
  4. Observe the Envoy process memory: `watch -n 1 "ps aux | grep [e]nvoy | awk '{printf \"PID: %s CPU: %.1f%% MEM: %.1f%% %.0f MB\\n\", \$2, \$3, \$4, \$6/1024}'"`
 
                                                                                                                
                                                                                                                                                                             
  ## Actual results
  - After turning on the plug-in, the memory skyrocketed from 482MB.
  - Growth ~30MB per second, reaching 648MB after 3-5 seconds
  - Linux OOM Killer kills the Envoy process
  - Log output: `Envoy exited with error: signal: killed`

<img width="1192" height="386" alt="Image" src="https://github.com/user-attachments/assets/771bd3fa-a8d5-4bf8-81fa-dcc8988e66b7" />

## My thoughts
Can you optimize the memory usage of the plug-in, or remind the user when Envoy crashes, otherwise the final effect is: after I enable the plug-in, the ai network management crashes and cannot accept any traffic

yyyCode:

Enabling the `ai-data-masking` plugin caused Envoy memory to spike (~30MB/s) and get OOM-killed within a few seconds on a 2C2G host — even with an **empty/default config and no request traffic** (issue #4157).

**Root cause:** `DenyWord` used `#[derive(Default)]`, so `DenyWord::default()` built its `jieba` field via `Jieba::default()`, which loads jieba's embedded default dictionary (tens of thousands of entries) into a Cedar trie. Two paths hit this with an **empty** deny-word set:

- `RuleMatcher::parse_config` calls `AiDataMaskingConfig::default()` unconditionally when the plugin config object is empty (this is exactly the empty-config repro in the issue), and
- the `deny_words` field falls back to its default when omitted.

An empty deny-word set can never match anything, so loading a dictionary there is pure waste. With one Wasm VM per worker thread, it was enough to exhaust memory and get Envoy OOM-killed.

**Fix:** implement `Default` for `DenyWord` manually to return `empty()` (no dictionary), and short-circuit `check()` when the word set is empty.

The default dictionary is **still** loaded in `from_iter` (used by both the system sensitive-word dictionary and user-configured `deny_words`), because it is required for correct segmentation: without it, a sensitive word can be spuriously matched as a substring of normal text.

fixes #4157

Added unit tests in `deny_word.rs`:
- `empty()` / `default()` match nothing (regression guard for the OOM path — the default instance must not carry words or load a dictionary)
- `from_iter()` matches added words (CN + EN) and skips normal/blank text
- **substring guard**: with the default dictionary, the sentence `工信处女干事…亲口交代24口交换机…` is not falsely blocked by the sensitive word `口交`, while `口交` on its own is still matched. This mirrors the existing e2e conformance case in `test/e2e/conformance/tests/rust-wasm-ai-data-masking.go`.

1. Build the plugin (`make build` under `plugins/wasm-rust`, `PLUGIN_NAME=ai-data-masking`).
2. On a low-memory host (e.g. 2C2G), enable `ai-data-masking` with an empty/default config and observe Envoy memory — it should stay flat instead of spiking and being OOM-killed.
3. Configure `deny_words` and/or `system_deny: true` and confirm sensitive words are still blocked, and normal text containing a sensitive substring is not.

- No behavior/config change for users; matching semantics are unchanged. This only removes an unnecessary dictionary load on the empty-deny-word paths.
- I could not run the full `cargo build`/e2e locally (no MSVC C++ build tools for host proc-macro compilation, and the `proxy-wasm` git dependency host was unreachable). The segmentation behavior for the substring case was cross-verified against jieba (default dict + HMM), which confirms `口交` is not segmented out of `亲口交代…口交换机…` even after adding it as a deny word. Please rely on CI for the full Wasm build + conformance run.
- The first CI run of this PR failed the RUST conformance test because an earlier version of the fix removed the default dictionary entirely; this revision keeps it for matching and only avoids it for empty deny-word sets.

> Investigate issue #4157 (ai-data-masking plugin OOMs immediately when enabled, even with empty config and no traffic), find the root cause, fix the bug, test it, and submit a PR following the project's commit and PR conventions. (After CI failed, iterate on the fix.)

### AI Coding Summary

- **Key decisions:**
  - Root-caused the empty-config OOM to a derived `Default` loading jieba's default dictionary for an empty deny-word set, via `RuleMatcher::parse_config` calling `Config::default()`.
  - An initial fix that dropped the default dictionary everywhere broke the RUST conformance test (a sensitive word was matched as a substring of normal text). Corrected to keep the dictionary for actual matching (`from_iter`) and only skip it for empty sets (`Default`/`empty`), plus an early return in `check()`.
- **Major changes:** manual `Default` impl returning `empty()`, empty-set short-circuit in `check()`, unit tests incl. a substring regression guard.
- **Limitations:** could not compile/run tests in the authoring environment; relied on source-level analysis and cross-verification with jieba's segmentation, and on CI for the full build.


johnlanni (changes requested):
The core diagnosis and `Default -> empty()` fix are correct, but two issues should be addressed before merge.

1. `DenyWord::from_iter` constructs `Jieba::new()` before determining whether any valid words remain. As a result, explicit `deny_words: []` and blank-only lists still load the full Jieba dictionary (roughly 48 MiB per native instance in focused measurements), even though they are semantically empty. Please trim/filter/collect first and return `Self::empty()` when no valid words remain; only construct `Jieba::new()` for a genuinely non-empty set.

2. The current `default_is_empty` test does not guard the memory regression. I verified all five new tests pass against the pre-fix implementation that still derives `Default` and loads the dictionary. Since the test module can access the private field, add a structural/segmentation assertion that distinguishes an empty Jieba from the default dictionary, for example `assert!(!deny.jieba.has_word("我们"))`, and cover omitted `deny_words`, `deny_words: []`, and blank-only values through deserialization.

Keeping the default dictionary for non-empty custom/system words is correct; the substring regression test demonstrates why it is required.

Issue #83, opened 2022-11-30
pkg/bootstrap 提升单测覆盖率 (Increase unit test coverage)


johnlanni:
-> 20%

Charlie17Li:
pls assign it to me :)

Issue #81, opened 2022-11-30
支持通过 helm 参数集成 prometheus & grafana，提供 metrics 大盘模版（Support the integration of prometheus & grafana through helm parameters, and provide metrics template）


johnlanni:

cobb-tx:
@johnlanni Hi, i'd like to resolve this issue, can you assign it to me?

johnlanni:
> @johnlanni Hi, i'd like to resolve this issue, can you assign it to me?

Ok, @Lynskylate will give you some support

PR #4187, opened 2026-07-23
security: avoid image request allocation overflows

EndlessSeeker:
## Summary

- remove potentially overflowing len(slice) + constant capacity calculations from OpenAI image edit and variation URL collection
- remove the same pattern from Vertex image request part construction
- keep the existing slice length as the safe initial capacity and let append grow the slice for optional fields
- add regression tests for URL ordering and Vertex image-plus-prompt request structure

## CodeQL coverage

This addresses the three open High go/allocation-size-overflow alerts in Higress-owned AI Proxy code: alerts 81, 82, and 83. It does not dismiss or suppress the findings.

## Validation

- go test ./provider
- go test -timeout 10m ./...
- git diff --check

The complete AI Proxy test suite passed, including the main package in 343.654 seconds.

### DIFF
--- a/plugins/wasm-go/extensions/ai-proxy/provider/image_request_test.go
+++ b/plugins/wasm-go/extensions/ai-proxy/provider/image_request_test.go
@@ -1,0 +1,50 @@
+package provider
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestImageRequestsCollectAllURLs(t *testing.T) {
+	images := []imageInputURL{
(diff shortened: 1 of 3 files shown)


EndlessSeeker:
Closing after a reachability review. The reported len(slice)+1/+2 overflow would require a slice already within one or two elements of MaxInt; these element types cannot be materialized within Go's address space before that point. The request parser also only appends a fixed one or two optional values. Removing the capacity hint does not close a practical attack path and can add reallocations. CodeQL alerts #81-#83 will be dismissed as not practically exploitable.

PR #4188, opened 2026-07-23
security: validate plugin integer conversions

EndlessSeeker:
## Summary

- parse response-cache HTTP status values with an explicit 32-bit bound and only cache valid 100-599 responses
- parse custom-response status_code as uint32 and reject negative, overflowing, or non-HTTP values
- parse WAF INITIAL_PAGES as uint32, enforce the WebAssembly 1-65536 page range, and pass the checked type through to the module patcher
- add regression coverage for malformed, negative, out-of-range, and overflowing status values

## CodeQL coverage

This addresses three open High go/incorrect-integer-conversion alerts in Higress-owned plugin code: alerts 78, 60, and 40. No finding is dismissed or suppressed.

## Validation

- go test ./... in plugins/wasm-go/extensions/response-cache
- go test ./... in plugins/wasm-go/extensions/custom-response
- go test -c -o /tmp/higress-waf-magefiles.test . in plugins/wasm-go/extensions/waf/magefiles
- git diff --check

The WAF magefiles package has build-tool checks in init, so it was compiled without executing the build-tool initialization; response-cache and custom-response ran their complete test suites.

### DIFF
--- a/plugins/wasm-go/extensions/custom-response/main.go
+++ b/plugins/wasm-go/extensions/custom-response/main.go
@@ -125,8 +125,8 @@ func parseRuleItem(gjson gjson.Result, rule *CustomResponseRule) error {
 	rule.statusCode = 200
 	if gjson.Get("status_code").Exists() {
 		statusCode := gjson.Get("status_code")
-		parsedStatusCode, err := strconv.Atoi(statusCode.String())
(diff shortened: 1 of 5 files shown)


johnlanni (approved):
LGTM

PR #4189, opened 2026-07-23
security: validate core integer bounds

EndlessSeeker:
## Summary

- parse local rate-limit annotations directly as uint32 values
- reject zero rates, zero burst multipliers, negative/overflowing annotations, and burst multiplication overflow
- parse ZooKeeper endpoint ports with a 16-bit bound and skip invalid, zero, or out-of-range endpoints instead of emitting wrapped port values
- add boundary regression tests for annotation parsing, burst overflow, and invalid ZooKeeper ports

## CodeQL coverage

This addresses three open High go/incorrect-integer-conversion alerts in Higress-owned core code: alerts 29, 30, and 5. No finding is dismissed or suppressed.

## Validation

- gofmt completed for all changed Go files
- git diff --check passed
- targeted tests were added for both code paths

A local targeted go test invocation was blocked before compilation because the existing generated external/istio tree in this workspace requires Go 1.26 while the repository and CI select Go 1.24. No dependency or generated tree was changed to bypass that mismatch; the clean GitHub CI run is the authoritative test for this PR.

### DIFF
--- a/pkg/ingress/kube/annotations/local_rate_limit.go
+++ b/pkg/ingress/kube/annotations/local_rate_limit.go
@@ -15,6 +15,8 @@
 package annotations
 
 import (
+	"fmt"
+
 	"github.com/golang/protobuf/ptypes/duration"
 	networking "istio.io/api/networking/v1alpha3"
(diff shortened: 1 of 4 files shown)


johnlanni (approved):
LGTM

Issue #62, opened 2022-11-23
Implement a plugin that can replace the response body according to the configuration


johnlanni:
## Why do you need it?

实现一个插件，可以匹配应答和状态码返回自定义应答
---
Implement a plugin that can replace the response body according to the configuration

## How could it be?

config example:

```yaml
rules:
- match:
     status: 403
     body: "RBAC: access denied"
  replace:
     status: 200
     body: "{\"code\":403,\"message\":\"User is not authenticated\"}"
     content-type:  "application/json;utf-8"
- match:
     status: 503
     body: "no healthy upstream"
  replace:
     status: 200
     body: "{\"code\":503,\"message\":\"No Healthy Upstream\"}"
     content-type:  "application/json;utf-8"
```

zhangjiahaol:
Hi

PR #4190, opened 2026-07-23
security: require literal bot hostname separator

EndlessSeeker:
## What this fixes

Addresses CodeQL High alert #62 (`go/incomplete-hostname-regexp`) in the bot-detect default user-agent rules. The dot in `boitho.com-dc` is now escaped so it matches a literal hostname separator instead of any character.

## Validation

- Added a regression test that accepts `boitho.com-dc/1.0`
- Added a negative case that rejects `boithoXcom-dc/1.0`
- `go test ./config`
- `go test ./...`

This PR fixes the source pattern directly and does not suppress or dismiss the alert.


johnlanni (approved):
LGTM

Issue #37, opened 2022-11-08
License Compatibility check


slievrly:
commit id: 14421bd4144b08bc633dee3611810249499fb379

Dependency                                         |              License |                                                    Version
-------------------------------------------------- | -------------------- | ----------------------------------------------------------
cloud.google.com/go                                |           Apache-2.0 |                                                    v0.97.0
cloud.google.com/go/logging                        |           Apache-2.0 |                                                     v1.4.2
contrib.go.opencensus.io/exporter/prometheus       |           Apache-2.0 |                                                     v0.4.0
github.com/Azure/go-ansiterm                       |                  MIT |                         v0.0.0-20210617225240-d185dfc1b5a1
github.com/Azure/go-autorest                       |           Apache-2.0 |                                       v14.2.0+incompatible
github.com/Azure/go-autorest/autorest              |           Apache-2.0 |                                                   v0.11.20
github.com/Azure/go-autorest/autorest/adal         |           Apache-2.0 |                                                    v0.9.15
github.com/Azure/go-autorest/autorest/date         |           Apache-2.0 |                                                     v0.3.0
github.com/Azure/go-autorest/logger                |           Apache-2.0 |                                                     v0.2.1
github.com/Azure/go-autorest/tracing               |           Apache-2.0 |                                                     v0.6.0
github.com/MakeNowJust/heredoc                     |                  MIT |                         v0.0.0-20170808103936-bb23615498cd
github.com/Masterminds/goutils                     |           Apache-2.0 |                                                     v1.1.1
github.com/Masterminds/semver/v3                   |                  MIT |                                                     v3.1.1
github.com/Masterminds/sprig/v3                    |                  MIT |                                                     v3.2.2
github.com/Microsoft/go-winio                      |                  MIT |                                                     v0.5.0
github.com/Microsoft/hcsshim                       |                  MIT |                                                    v0.8.21
github.com/PuerkitoBio/purell                      |         BSD-3-Clause |                                                     v1.1.1
github.com/PuerkitoBio/urlesc                      |         BSD-3-Clause |                         v0.0.0-20170810143723-de5bf2ad4578
github.com/aws/aws-sdk-go                          |           Apache-2.0 |                                                    v1.41.7
github.com/beorn7/perks                            |                  MIT |                                                     v1.0.1
github.com/cenkalti/backoff/v4                     |                  MIT |                                                     v4.1.1
github.com/census-instrumentation/opencensus-proto |           Apache-2.0 |                                                     v0.3.0
[...]

PR #4191, opened 2026-07-23
test: verify conformance TLS certificates

EndlessSeeker:
## What this fixes

Addresses CodeQL High alerts #36 and #37 (`go/disabled-certificate-check`) in the conformance test round tripper. TLS clients now verify the gateway certificate against the CA supplied by each test instead of setting `InsecureSkipVerify`.

The ConfigMap HTTPS test replaces its expired fixture certificates at runtime with short-lived test certificates and trusts their generated CAs. The HTTPS-without-SNI test likewise supplies the CA it already generates.

## Validation

- `go test ./test/e2e/conformance/utils/roundtripper`
- Verified there are no remaining `InsecureSkipVerify` uses in the conformance round tripper
- Full conformance coverage is delegated to the repository CI because the local generated `external/istio` workspace requires Go 1.26 while the repository toolchain is Go 1.24.4

This PR fixes certificate authentication directly and does not suppress or dismiss either alert.

### DIFF
--- a/test/e2e/conformance/tests/configmap-https.go
+++ b/test/e2e/conformance/tests/configmap-https.go
@@ -15,8 +15,13 @@
 package tests
 
 import (
+	"context"
 	"testing"
 
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/alibaba/higress/v2/test/e2e/conformance/utils/cert"
 	"github.com/alibaba/higress/v2/test/e2e/conformance/utils/http"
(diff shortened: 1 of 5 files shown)


EndlessSeeker:
Closing after scope review. Both findings are in test/e2e conformance code and CodeQL classifies both instances as test code. They do not ship in the controller, gateway, CLI, or plugins. Trusting an ephemeral test endpoint is not a production vulnerability; the proposed CA fixture is optional test hardening rather than a required security fix. Alerts #36 and #37 will be dismissed as used in tests.

PR #4213, opened 2026-07-27
feat(wasm): add Qwen3Guard security plugin

JianweiWang:
## Summary

- add a Qwen3Guard-Gen Wasm plugin for OpenAI-compatible request and response moderation
- support request checks, non-streaming response checks, and buffered SSE response checks
- support `k8s`, `nacos`, `ip`, and `dns` service discovery with configurable risk thresholds and deny responses
- add vendor-neutral setup, verification, failure-mode, and security documentation

## Why

This allows Higress users to connect a self-hosted or otherwise reachable Qwen3Guard service without binding the plugin to a specific commercial deployment platform.

## Behavior and safety

- fail open when the guard service is unavailable or returns an unsupported response
- bound request, response, and streaming moderation buffers with `maxBodyBytes`
- switch oversized streaming responses to pass-through mode instead of continuing to accumulate Wasm memory
- default an omitted Kubernetes namespace to `default`
- return OpenAI-compatible JSON or SSE deny responses when the configured risk threshold is reached

## Validation

- `go test ./...`
- `GOOS=wasip1 GOARCH=wasm go build -buildmode=c-shared`
- verified the Wasm export contains `proxy_abi_version_0_2_100`
- `git diff --check`


rinfx (approved):
LGTM

PR #4219, opened 2026-07-28
fix(ai-proxy): detect Baidu embeddings API paths

123123213weqw:
## What changed

- recognize Baidu Qianfan's `/v2/embeddings` endpoint in `GetApiName`
- add regression coverage for chat, embeddings, prefixed embeddings, and unknown paths

## Why

Baidu Qianfan includes embeddings in its default capabilities, but original-protocol requests are classified by the provider-specific `GetApiName` hook. That hook only recognized chat completions, leaving direct `/v2/embeddings` requests with an empty API name.

## Validation

- `go test ./provider`
- `git diff --check`


johnlanni:

PR #4220, opened 2026-07-28
fix(ai-proxy): detect hunyuan embeddings API paths

123123213weqw:
## Problem

hunyuanProvider.GetApiName() unconditionally returned ApiNameChatCompletion regardless of the request path. Since DefaultCapabilities() declares support for ApiNameEmbeddings, embeddings requests (/v1/embeddings) were silently misprocessed as chat completions — signed with the wrong TC-Action (ChatCompletions) and sent to Tencent Cloud with an incompatible payload.

## Fix

Add path detection for the embeddings endpoint before falling back to the chat completion default.

## Testing

Added hunyuan_test.go with unit tests covering:
- Chat completions path detection
- Embeddings path detection
- Embeddings with route prefix
- Unknown path defaults to chat completion


johnlanni:

PR #4224, opened 2026-07-28
bugfix: fix nil pointer panic in ai-load-balancer cluster_metrics type assertions

wc4440222:
## Ⅰ. 描述此 PR 做了什么

Fix unsafe type assertions in `ai-load-balancer/cluster_metrics/lb_policy.go` that can cause Wasm plugin panics.

`HandleHttpStreamingResponseBody` and `HandleHttpStreamDone` use bare type assertions (`.(string)`, `.(int64)`) on context values (`candidate`, `request_start`, `statusCode`). These context values are only set in `HandleHttpRequestHeaders` and `HandleHttpResponseHeaders`. When Envoy invokes response callbacks without first calling request header callbacks (e.g. upstream connection errors, Envoy internal errors), these values are `nil`, causing the bare type assertions to panic and crash the entire Wasm instance.

## Ⅱ. 是否修复某个 Issue

Related to general stability improvements for the ai-load-balancer plugin.

## Ⅲ. 为何不加测试用例

The `cluster_metrics` package currently has no test files. The fix is a straightforward defensive programming change (comma-ok pattern) that prevents panics in edge cases. Adding unit tests would require mocking the `wrapper.HttpContext` interface which is non-trivial.

## Ⅳ. 如何验证

- `cd plugins/wasm-go/extensions/ai-load-balancer`
- `go test ./...` — existing tests pass

## Ⅴ. 评审特别说明

The fix uses the comma-ok pattern for all 6 unsafe type assertions across two functions:
- `HandleHttpStreamingResponseBody`: 3 assertions fixed (candidate, request_start, statusCode)
- `HandleHttpStreamDone`: 3 assertions fixed (candidate, request_start, statusCode)

When context values are missing, the functions return early (`return data` for streaming body, `return` for stream done) instead of panicking.

## Ⅵ. AI Coding 工具使用清单

- [x] 使用了 AI 编码工具: Yes (GitHub Copilot / ChatGPT assisted with code analysis)
- [ ] 新独立插件场景: N/A (bug fix in existing code)
- [x] 常规修改场景: Provided the analysis of unsafe type assertions and the fix pattern


johnlanni (approved):
LGTM. This fixes reachable missing-context panics in StreamDone/streaming response lifecycle paths. The candidate context and inflight increment are written together, so skipping cleanup when candidate is absent is correct; when candidate exists, decrement happens before optional latency recording, preserving counter pairing. Missing response status is correctly treated as a failed request for load-balancer feedback. Non-blocking follow-up: upstream focused fake-context tests and add debug diagnostics on fallback paths.

PR #4256, opened 2026-08-04
bugfix: fix unsafe type assertions in ai-quota and ai-cache plugins

wc4440222:
## I. What does this PR do

Fixes unsafe type assertions in the ai-quota and ai-cache plugins that could panic if context values contain unexpected types.

### Root Cause

Both plugins use bare type assertions on `ctx.GetContext()` values. While nil checks exist in most locations, type checks are missing — if a context key is set to a different type (e.g. by a plugin interaction or race condition), the bare assertion panics.

### Fix

**ai-quota/main.go** (3 assertions in `onHttpStreamingResponseBody`):
- `inputToken.(int64)` → comma-ok pattern, return `data` if not int64
- `outputToken.(int64)` → comma-ok pattern, return `data` if not int64
- `consumer.(string)` → comma-ok pattern, return `data` if not string

**ai-cache/util.go** (8 assertions across 5 functions):
- `handleNonStreamChunk`: type check for `tempContentI.([]byte)`
- `handleStreamChunk`: type check for `partialMessageI.([]byte)`
- `processNonStreamLastChunk`: type check for `tempContentI.([]byte)`
- `processStreamLastChunk`: type checks for `partialMessageI.([]byte)` and `tempContentI.(string)`
- `processSSEMessage`: type checks for cache content string assertions (2 locations)

## II. Fixes Issue

Related to general stability improvements for ai-quota and ai-cache plugins.

## III. Why no test cases

Both plugins require Wasm runtime and Redis/cache infrastructure to test. The fix follows the same defensive pattern approved in #4226.

## IV. How to verify

- `cd plugins/wasm-go/extensions/ai-quota && GOOS=wasip1 GOARCH=wasm go build ./...` — compiles
- `cd plugins/wasm-go/extensions/ai-cache && GOOS=wasip1 GOARCH=wasm go build ./...` — compiles

- Same pattern as #4226 (WAF plugin, approved by @johnlanni) and #4255 (SSE proxy)
- When context types don't match, functions return early with safe defaults instead of panicking
- ai-quota: returns `data` (passing through the streaming body unchanged)
- ai-cache: falls back to treating the chunk as the initial content or returns empty string

- [x] Regular modification scenario: AI analyzed the ai-quota and ai-cache plugin context usage to identify all unsafe type assertions

### DIFF
--- a/plugins/wasm-go/extensions/ai-cache/util.go
+++ b/plugins/wasm-go/extensions/ai-cache/util.go
@@ -17,7 +17,11 @@ func handleNonStreamChunk(ctx wrapper.HttpContext, c config.PluginConfig, chunk 
 		ctx.SetContext(CACHE_CONTENT_CONTEXT_KEY, chunk)
 		return nil
 	}
-	tempContent := tempContentI.([]byte)
+	tempContent, ok := tempContentI.([]byte)
+	if !ok {
+		ctx.SetContext(CACHE_CONTENT_CONTEXT_KEY, chunk)
+		return nil
+	}
 	tempContent = append(tempContent, chunk...)
 	ctx.SetContext(CACHE_CONTENT_CONTEXT_KEY, tempContent)
 	return nil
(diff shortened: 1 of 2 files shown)


johnlanni (approved):
LGTM. The changes preserve normal-path quota deduction and cache assembly semantics while making unexpected context types fail open instead of panicking. Streaming response bytes remain unchanged, quota deduction stays end-of-stream and exactly-once, and invalid cache state degrades to a cache miss rather than poisoning stored values. Non-blocking follow-up: add bounded diagnostics and focused wrong-type context tests.

PR #4259, opened 2026-08-04
bugfix: always include system_fingerprint in ai-proxy streaming responses

### LINKED ISSUE #1658: 使用AI-Proxy插件代理 qwen 模型进行流式返回时因字段不全造成 golang sdk 解析失败

- [x] I have searched the [issues](https://github.com/alibaba/higress/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description

使用 Higress 配置 ai-proxy 插件来代理 dashscope / openai 等模型，在使用社区 [golang sdk](github.com/sashabaranov/go-openai) 进行流式返回的时候发现使用 qwen-plus 模型时， golang 侧解析出来的response结构体为空，但使用 gpt-4o-mini 模型时并没有问题

通过在终端使用 curl 测试两者区别发现使用 qwen 系列模型时返回结构中缺少部分值为 null 的字段

qwen系列返回数据
image

openai 系列返回数据
image

在 sdk 的仓库里看到结构体定义中， system_fingerprint 、finish_reason 等字段是必须存在的(没有添加标签omitempty)，导致了结构体解析失败
https://github.com/sashabaranov/go-openai/blob/2a0ff5ac63e460cbe44cccd0d4199d51bf8682a4/chat_stream.go#L38
https://github.com/sashabaranov/go-openai/blob/2a0ff5ac63e460cbe44cccd0d4199d51bf8682a4/chat_stream.go#L53

### Ⅱ. Describe what happened

  If there is an exception, please attach the exception trace:

```
Just paste your stack trace here!
```

希望在返回的结构中与 Openai 保持一致，补全缺失的字段信息

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- Higress version:  
- OS :
- Others:

wc4440222:
## I. What does this PR do

Fixes a compatibility issue where the `system_fingerprint` field is missing from ai-proxy streaming responses, causing older versions of the `sashabaranov/go-openai` SDK to fail JSON parsing.

### Root Cause

The `chatCompletionResponse` struct in `provider/model.go` defines `SystemFingerprint` with `json:"system_fingerprint,omitempty"`. When the value is an empty string (which it always is for Qwen and other non-OpenAI providers), `omitempty` causes the field to be completely omitted from the JSON output.

OpenAI's official API always includes `system_fingerprint` in streaming responses (as `null` or a fingerprint string). The older `sashabaranov/go-openai` SDK versions define this field without `omitempty`, so when the field is missing from the response, JSON unmarshalling fails and the parsed response struct is empty.

This affects:
1. **Native dashscope mode**: `buildChatCompletionStreamingResponse` sets `SystemFingerprint: ""`, which gets omitted due to `omitempty`
2. **OpenAI compatible mode**: upstream responses are passed through as-is, but any provider response missing this field will trigger the SDK issue

### Fix

Remove `omitempty` from `SystemFingerprint` in `chatCompletionResponse` so it is always serialized. When the value is empty, it will be serialized as `"system_fingerprint":""` which is valid JSON and compatible with all SDK versions.

This is a 1-line change that aligns Higress with OpenAI's official behavior.

## II. Fixes Issue

Fixes #1658

## III. Why don't you add test cases

The fix is a struct tag change. Existing tests continue to pass. The behavioral change (field always present in JSON output) can be verified by checking the serialized JSON includes the `system_fingerprint` key.

## IV. How to verify

- `cd plugins/wasm-go/extensions/ai-proxy && GOOS=wasip1 GOARCH=wasm go build . ./provider/` — compiles
- `go test . -run Test_normalizeOpenAiRequestBody` — existing tests pass
- Stream a Qwen request through ai-proxy and verify `system_fingerprint` is present in each SSE chunk's JSON

- This is a minimal, non-breaking change: `system_fingerprint` is now always present as an empty string instead of being omitted entirely
- OpenAI's official API always includes this field, so this aligns Higress with the expected behavior
- The change affects all providers using `chatCompletionResponse` (which is the standard OpenAI-compatible response type), but only adds a field that was supposed to be there anyway
- Comment posted on #1658 asking to confirm this should be fixed at the gateway level — this PR implements the fix proactively since the change is safe and reversible

- [x] Used AI coding tools: Yes (AI-assisted code analysis)
- [ ] New standalone plugin scenario: N/A (bug fix)
- [x] Regular modification scenario: AI analyzed the chatCompletionResponse struct and omitempty behavior

### DIFF
--- a/plugins/wasm-go/extensions/ai-proxy/provider/model.go
+++ b/plugins/wasm-go/extensions/ai-proxy/provider/model.go
@@ -42,39 +42,39 @@ type thinkingParam struct {
 
 type chatCompletionRequest struct {
 	NonOpenAIStyleOptions
-	Messages             []chatMessage          `json:"messages"`
-	Model                string                 `json:"model"`
-	Store                bool                   `json:"store,omitempty"`
-	ReasoningEffort      string                 `json:"reasoning_effort,omitempty"`
-	ClaudeThinking       *claudeThinkingConfig  `json:"claude_thinking,omitempty"`
-	ClaudeOutputConfig   *claudeOutputConfig    `json:"claude_output_config,omitempty"`
-	ClaudeAnthropicBeta  []string               `json:"claude_anthropic_beta,omitempty"`
-	Metadata             map[string]string      `json:"metadata,omitempty"`
-	FrequencyPenalty     float64                `json:"frequency_penalty,omitempty"`
-	LogitBias            map[string]int         `json:"logit_bias,omitempty"`
-	Logprobs             bool                   `json:"logprobs,omitempty"`
-	TopLogprobs          int                    `json:"top_logprobs,omitempty"`
-	MaxTokens            int                    `json:"max_tokens,omitempty"`
-	MaxCompletionTokens  int                    `json:"max_completion_tokens,omitempty"`
-	N                    int                    `json:"n,omitempty"`
-	Modalities           []string               `json:"modalities,omitempty"`
-	Prediction           map[string]interface{} `json:"prediction,omitempty"`
-	Audio                map[string]interface{} `json:"audio,omitempty"`
-	PresencePenalty      float64                `json:"presence_penalty,omitempty"`
-	ResponseFormat       map[string]interface{} `json:"response_format,omitempty"`
-	Seed                 int                    `json:"seed,omitempty"`
-	ServiceTier          string                 `json:"service_tier,omitempty"`
-	Stop                 []string               `json:"stop,omitempty"`
-	Stream               bool                   `json:"stream,omitempty"`
-	StreamOptions        *streamOptions         `json:"stream_options,omitempty"`
-	PromptCacheRetention string                 `json:"prompt_cache_retention,omitempty"`
-	PromptCacheKey       string                 `json:"prompt_cache_key,omitempty"`
-	Temperature          float64                `json:"temperature,omitempty"`
-	TopP                 float64                `json:"top_p,omitempty"`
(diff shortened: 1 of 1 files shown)


johnlanni:

wc4440222:
Thanks for the feedback, @johnlanni. A dedicated issue with a concrete reproduction has been created:

#4323 — ai-proxy omits system_fingerprint field in streaming responses, breaking older go-openai SDK versions

The issue includes:
- Exact Higress version (v2.2.3, commit `88d7769`)
- Minimal ai-proxy configuration with Qwen provider
- Client code example using `sashabaranov/go-openai` SDK
- Step-by-step reproduction procedure
- Expected vs actual JSON output (with/without `system_fingerprint` field)
- Root cause location (`provider/model.go:174`)
- Confirmation that it reproduces on current `main`

This PR (#4259) is the proposed fix for #4323. Related original report: #1658.

wc4440222:
@johnlanni @rinfx Gentle ping — this PR was submitted on Aug 4 and the linked issue #4323 was created per your requirement. Could you please review when you have time? Happy to make any changes if needed.

wc4440222:
Friendly ping @rinfx — this PR has been waiting for review for over 2 weeks. Could you please take a look?

wc4440222:
@johnlanni @rinfx Friendly reminder — this PR has been open for over 3 weeks. It fixes the `system_fingerprint` field omission in ai-proxy streaming responses (#4323, originally reported in #1658). The fix is minimal and aligns the output with OpenAI's official API behavior. Would appreciate a review when you have time. Thanks!

PR #4268, opened 2026-08-05
bugfix: avoid panic on malformed multimodal content

yuluo-yx:

`ParseContent` used unchecked string assertions for `image_url.url`, `input_audio.data`, `input_audio.format`, and `file.file_id`. Malformed OpenAI-compatible JSON could therefore panic.

This change validates all four required fields with comma-ok assertions and skips the entire malformed structured part. Conversion of valid image, audio, and file parts is unchanged.

For Bedrock, a message containing only malformed parts converts to empty content. The request builder now omits that empty message and verifies content length before inspecting the first block while merging tool results. This prevents an all-malformed user message followed by a tool message from causing a secondary slice-index panic or producing an empty Bedrock message.

Closes OSS-24

Regression tests cover:

- missing and non-string required multimodal fields;
- valid parts before and after malformed parts, preserving valid ordering;
- content containing only malformed structured parts;
- an all-malformed user message followed by a Bedrock tool result;
- unchanged conversion of valid image, audio, and file parts.

- Before the Bedrock fix, the focused regression test panicked at `provider/bedrock.go:924`.
- `go test ./provider -count=1` passes.
- Focused regression tests pass with `-race`.
- `git diff --check` passes.
- Provider package coverage is 31.9% of statements; each new behavior branch has focused regression coverage.
- `go test -timeout=3m ./... -count=1` passed `config`, `provider`, and `util`, but the existing root-package WASM suite timed out while wazero was compiling `TestTogetherAI/wasm` on ARM64.
- `go vet ./provider` reports only pre-existing JSON tags on unexported fields in `context.go`, `failover.go`, `provider.go`, and `retry.go`.
- Diff size: 248 additions and 7 deletions across four files.

Kubernetes, kind, Docker, conformance, and WASM E2E tests were not run locally; those heavier environments are left to the existing CI workflows.

The behavior change is limited to malformed or provider-unsupported structured content: invalid parts are skipped, and Bedrock messages with no convertible content are omitted. No migration is required. Rollback by reverting commits `30b32910` and `b7e8063a` in that order.

Fix the panic caused by malformed multimodal fields in `ParseContent`: safely assert required strings and skip invalid items, preserve valid conversions, and add table-driven regression tests. Address the review finding that all-malformed content followed by a Bedrock tool message still panics, including mixed-order and all-invalid regression coverage. Do not run local container- or cluster-based tests.

### AI Coding Summary

- Key decision: preserve successful parsing of valid parts while treating each malformed structured part as independently skippable; omit provider messages when no content remains after conversion.
- Major changes: replace four unchecked string assertions with validated comma-ok assertions, guard Bedrock tool-result merging, omit empty Bedrock messages, and add focused regression tests.
- Important considerations: the patch intentionally changes only malformed or unsupported-input behavior. Lightweight provider tests passed; container, cluster, conformance, and WASM E2E coverage is delegated to existing CI.


johnlanni (changes requested):
The targeted comma-ok guards are correct and the new tests genuinely reproduce the pre-fix type-assertion panic. However, the malformed-input path is not yet panic-free for Bedrock.

When all structured parts in a user message are malformed, `ParseContent` now returns an empty content slice. If the following message is a tool message, `buildBedrockTextGenerationRequest` reaches `messages[len(messages)-1].Content[0]` at `provider/bedrock.go:924` without checking `len(Content)`, causing `panic: runtime error: index out of range [0] with length 0`. A focused throwaway regression test reproduces this on the current PR head. Empty content arrays could trigger the underlying bug before this PR, but this change makes additional malformed multimodal inputs reach it, and the PR goal is specifically to prevent malformed multimodal content from panicking.

Please guard the `Content[0]` access (or otherwise handle empty converted content) and add regression coverage for an all-malformed user message followed by a tool message. It would also be useful to cover mixed valid/malformed ordering and all-invalid content explicitly.

Non-blocking semantic note: silently dropping a malformed part can change prompt meaning. A future validation path that returns a clear 400 would be more transparent, but the current skip-part behavior is consistent with the existing `ParseContent` design.

yuluo-yx:
### Implementation Rationale

Exact head: `30b32910b9b5d225d755529a9d845200234c7e21`

- `provider/model.go:chatMessage.ParseContent` uses comma-ok assertions for every required multimodal string so one malformed structured part is discarded without affecting valid neighbors.
- `provider/bedrock.go:buildBedrockTextGenerationRequest` omits converted messages whose content is empty. This avoids serializing an invalid Bedrock `content: []` entry after every part was malformed or unsupported.
- The tool-result merge checks both the previous-message index and its content length before accessing `Content[0]`, removing the reviewed slice-index panic even if future callers introduce another empty-message path.
- Regression coverage now includes valid/malformed ordering, all-invalid content, and the exact all-malformed-user-then-tool sequence reported in review.

Verification: full `provider` tests and focused race tests pass. The module-wide root WASM suite reached its 3-minute timeout while wazero compiled the existing `TestTogetherAI/wasm`; `config`, `provider`, and `util` passed in that same run.

PR #4278, opened 2026-08-06
fix(mcp-session): handle multiple MCP server domains

### LINKED ISSUE #4277: mcp-session skips all hosts when an MCP Server has multiple domains
- [x] I have searched the [issues](https://github.com/higress-group/higress/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description

When an MCP Server is configured with more than one domain, the generated `mcp-session` rule does not match any of the configured domains.

### Ⅱ. Describe what happened

I ran into this while validating a setup where the same MCP Server is available through an external hostname and an internal hostname.

The Ingress used this annotation:

```yaml
higress.io/mcp-server: "true"
higress.io/mcp-server-match-rule-domains: "orders.example.com,orders-internal.example.com"
higress.io/mcp-server-match-rule-type: "prefix"
higress.io/mcp-server-match-rule-value: "/mcp"
higress.io/mcp-server-upstream-type: "sse"
```

The annotation parser keeps both domains, but the controller writes them into one `match_rule_domain` value:

```text
(orders.example.com|orders-internal.example.com)
```

`mcp-session` does not treat that value as a regular expression. Its host matcher supports an exact hostname or a leading/trailing `*`, so the generated value is compared as a literal hostname. I reproduced the matcher result locally and neither configured host matches:

```text
generated matcher does not match configured host "orders.example.com"
generated matcher does not match configured host "orders-internal.example.com"
```

There is no configuration error. The request simply does not match the rule, so `mcp-session` continues without processing it.

Both configured hostnames should match the same MCP Server path and session settings. One compatible representation would be a separate match rule for each domain.

1. Configure an MCP Server with two comma-separated values in `higress.io/mcp-server-match-rule-domains`.
2. Generate the `mcp-session` filter configuration.
3. Inspect `plugin_config.value.match_list[0].match_rule_domain`.
4. Try to match either configured hostname with the current `mcp-session` host matcher.

The controller currently generates one regex-like string instead of two matcher-compatible domain rules.

The multiple-domain annotation is already covered by the annotation parser tests, so the input itself is supported. The mismatch happens between controller-side config generation and the data-plane host matcher.

I used OpenAI Codex to trace the generated configuration, reproduce the matcher behavior, and search existing issues and pull requests. No direct duplicate or in-progress fix was found.

### Ⅵ. Environment:

- Higress version: `main` at `b79b5cf12359750849b8adfe9c3f518cf08dc8e7`
- OS: Windows 11, 64-bit
- Others: Go 1.24.4; source-level configuration and matcher reproduction

jesseedcp:
Fixes #4277

Generate one `mcp-session` match rule for each domain configured on an MCP Server instead of joining multiple domains into one regex-like literal string.

The existing host matcher supports exact hostnames and leading/trailing `*` patterns. It therefore treated the old value `(domain-a|domain-b)` as one literal hostname and matched neither configured domain.

The change preserves the existing empty-domain fallback, single-domain behavior, domain order, and the priority of manually configured match rules.

Yes. It fixes the issue linked above.

Regression tests are included. They cover empty, single, and multiple provider domains, verify all copied rule fields, and confirm that manually configured rules remain ahead of provider-generated rules.

The multiple-domain case fails on the previous implementation because it receives one `(domain-a|domain-b)` rule instead of two independent rules.

```bash
go test ./pkg/ingress/kube/configmap -run '^TestMcpServerController_constructMcpSessionStructProviderDomains$' -count=1 -v
go test ./pkg/ingress/kube/configmap -count=1
go test ./pkg/ingress/kube/annotations ./pkg/ingress/kube/mcpserver -count=1
(cd plugins/golang-filter && go test ./mcp-session/common -count=1)
go vet ./pkg/ingress/kube/configmap
```

All commands pass locally with Go 1.24.4.

This PR intentionally does not add regular-expression support to `mcp-session`, reorder or deduplicate domains, or change the existing fallback for providers that return an empty domain list. It only converts the existing domain slice into matcher-compatible rules.

Inspect the MCP Server multi-domain configuration path on the current Higress `main` branch. Reproduce the mismatch between the controller-generated value and the `mcp-session` host matcher using non-sensitive example domains; search existing issues, pull requests, branches, and file history for duplicates; create a bug issue based on the verified local reproduction; implement the smallest backward-compatible fix test-first; run the related control-plane and matcher tests; and follow an issue-then-immediate-PR contribution flow.

### AI Coding Summary

- Confirmed that the controller joined multiple domains into a regex-like value while the consumer treated it as a literal hostname.
- Chose to emit one rule per domain rather than expand the matcher with regular-expression semantics.
- Preserved empty-domain, single-domain, ordering, and first-match behavior.
- Added a regression test that fails on the old implementation and covers empty, single, and multiple domain lists.
- Ran the focused test, related package suites, the `mcp-session/common` suite, and `go vet`.
- Validation is source- and unit-test-based; no full Kubernetes gateway deployment was used.


johnlanni (approved):
LGTM. The controller previously emitted `(domain-a|domain-b)` as one `match_rule_domain`, but the mcp-session dataplane matcher supports only exact, leading-`*` suffix, and trailing-`*` prefix matching—not regex alternation—so the old value matched neither domain. Emitting one adjacent rule per domain is the minimal correct fix and preserves manual-rule precedence, provider/server/domain order, single-domain behavior, copied rule fields, and wildcard semantics.

The focused regression test passes on this head and fails only for the multiple-domain case on the old implementation. Related configmap/annotation/mcpserver suites, mcp-session tests, race, vet, and formatting checks passed.

Non-blocking follow-up: trim and filter comma-separated annotation domains so values such as `a.example.com, b.example.com` do not retain leading whitespace.

PR #4282, opened 2026-08-06
fix(ai-statistics): make request body buffer limit configurable

### LINKED ISSUE #4267: ai-statistics: hard-coded 100 MiB request-body buffer causes 413 for non-AI multipart uploads
## Summary

The `ai-statistics` Wasm plugin unconditionally enables request-body buffering with a hard-coded 100 MiB limit.

In `plugins/wasm-go/extensions/ai-statistics/main.go`:

```go
defaultMaxBodyBytes uint32 = 100 * 1024 * 1024
```

The plugin then calls:

```go
ctx.SetRequestBodyBufferLimit(defaultMaxBodyBytes)
```

When `ai-statistics` is enabled globally (`defaultConfigDisable: false`), any route receiving a request body larger than approximately 100 MiB can be rejected by Envoy with HTTP 413 before the request reaches the upstream application.

## Impact

This also affects non-LLM routes, including `multipart/form-data` file uploads.

Observed Envoy access-log characteristics:

```text
response_code=413
response_code_details=request_payload_too_large
upstream_service_time=-
bytes_received=104873984
```

The upstream application does not receive the request.

## Steps to Reproduce

1. Enable `ai-statistics` globally with `defaultConfigDisable: false`.
2. Configure any normal HTTP route; it does not need to be an AI/LLM route.
3. Send a `multipart/form-data` request with a body larger than 100 MiB.
4. Envoy returns `413 Payload Too Large` before forwarding to the upstream service.

## Expected Behavior

`ai-statistics` should not impose an undocumented hard-coded request-body limit on unrelated routes.

The request-body buffer limit should be configurable and support route-level overrides. The plugin should also avoid buffering request bodies when body inspection is not required.

## Proposed Improvement

Please consider adding a configurable field such as:

```yaml
max_request_body_bytes: 10485760
```

with support for `matchRules` overrides:

```yaml
matchRules:
- ingress:
  - upload-route
  config:
    max_request_body_bytes: 524288000
```

Additionally, request-body buffering should be conditional:

- Buffer only when the configured attributes require request-body extraction.
- Avoid buffering `multipart/form-data` and other non-LLM request bodies by default.
- Allow an explicit opt-in if request-body buffering is required for such content types.

## Current Workaround

A locally rebuilt `ai-statistics` WASM with the limit raised to 500 MiB allows the large request to pass, which confirms that the hard-coded plugin buffer limit is the source of the 413 response.

srpatcha:
## Ⅰ. Describe what this PR does

Fixes #4267

The `ai-statistics` plugin unconditionally set the request-body buffer limit to a hard-coded 100 MiB (`defaultMaxBodyBytes`). When the plugin is enabled globally (`defaultConfigDisable: false`), any route — including non-AI `multipart/form-data` uploads — receiving a body larger than ~100 MiB was rejected by Envoy with **HTTP 413** before reaching the upstream application.

## Ⅱ. Changes

- **`main.go`** — add a configurable `max_request_body_bytes` field to `AIStatisticsConfig`:
  - Parsed in `parseConfig`, defaulting to `defaultMaxBodyBytes` (100 MiB).
  - Supports **matchRules route-level overrides** (larger limits for upload routes) since the plugin uses `wrapper.ParseConfig`.
  - An explicit `0` falls back to the default, since a zero limit would disable buffering and break model-field extraction.
  - `onHttpRequestHeaders` now uses `config.maxRequestBodyBytes` instead of the hard-coded constant.
- **`README.md` / `README_EN.md`** — document the new option.
- **`main_extra_test.go`** — tests for the default, explicit-value, and zero-fallback cases.

## Ⅲ. Example: route-level override for an upload route

```yaml
max_request_body_bytes: 10485760   # 10 MiB default for AI routes
matchRules:
- ingress:
  - upload-route
  config:
    max_request_body_bytes: 524288000   # 500 MiB for large uploads
```

## Ⅳ. Notes

The model field is still always extracted from the request body for AI routes, so buffering behavior is preserved; only the limit becomes configurable. Backward compatible — existing deployments keep the 100 MiB default.


EndlessSeeker:
Thanks for working on this. I see that #4267 reports `multipart/form-data` uploads over 100 MiB being rejected when `ai-statistics` is enabled globally. Could you clarify which real-world workload motivates making the buffer limit configurable? Have you encountered AI requests over 100 MiB in a deployment, or is the main use case large non-AI uploads passing through a gateway with this plugin enabled globally?

If you have a concrete example, the approximate request size, content type, and plugin/route configuration would help us assess the right behavior. I would expect 100 MiB to cover most AI requests; for non-AI uploads, avoiding request-body buffering when inspection is unnecessary may address the underlying problem more directly. Does this PR intentionally focus on the configurable-limit part of #4267?

PR #4294, opened 2026-08-07
fix(nacos): make watcher timeout configurable

### LINKED ISSUE #4293: Nacos registry watcher gRPC timeout causes intermittent connection failures and XDS push storms
## Problem

When using `nacos3` registry type with a large number of services (200+), the controller intermittently logs gRPC connection errors:

ERROR  rpc/rpc_client.go:630  Send request fail, request=ServiceListRequest, body={"namespace":"...","groupName":"DEFAULT","pageNo":N,"pageSize":50}, error=rpc error: code = Canceled desc = grpc: the client connection is closing
ERROR  fetch nacos service list failed, err:rpc error: code = DeadlineExceeded desc = context deadline exceeded, pageNo:N
ERROR  fetch nacos service list failed, err:client not connected, current status:UNHEALTHY, pageNo:N

These errors occur periodically (~every 10 minutes) and are accompanied by:
- Readiness probe timeouts (`context deadline exceeded`)
- XDS push storms (200+ full pushes in a single burst upon recovery)
- 3-4 minute stalls where no config updates are processed

### 1. `DefaultNacosTimeout` is hardcoded to 5000ms

In `registry/nacos/v2/watcher.go:48`:

```go
const DefaultNacosTimeout = 5000

This value is passed directly to the Nacos SDK client config with no way to override via McpBridge CRD or any external
configuration:

constant.WithTimeoutMs(DefaultNacosTimeout)

When the Nacos server takes >5s to respond to call (which happens with 200+ services across4+ pages), the gRPC DeadlineExceeded fires.

2. Retry logic has no backoff

In fetchAllServices() (registry/nacos/v2/watcher.go:385-464):

if err != nil {
    if tries > 10 {
        return err
    }
    log.Errorf("fetch nacos service list failed, err:%v, pageNo:%d", err, page)
    page--
    tries++
    continue  // immediate retry, no backoff
}

When the first DeadlineExceeded occurs, the Nacos SDK internally marks the connection as UNHEALTHY and begins
reconnection (typically takes 10-30s). All 10iately and hit the UNHEALTHY connection,exhausting the retry budget within ~3 seconds. The function then returns an error, and the next attempt only happens
on the next ticker cycle (30s later).

The error cascade:
DeadlineExceeded (5s timeout) → SDK closes connection → UNHEALTHY × N retries → 30s wait → retry cycle

3. fetchAllServices holds mutex during entire retry loop

fetchAllServices() acquires w.mutex.Lock() at the start (line 386) and holds it through the entire pagination + retry
loop. During the 5s timeout + retry period, n can proceed.

4. Subscribe callback bursts trigger XDS push

When fetchAllServices() eventually succeeds a accumulated subscribe callbacks firesimultaneously, triggering a burst of 200+ XDS full pushes in a single second. This causes CPU spikes and readiness
probe timeouts.

Environment

- Higress version: 2.2.2
- Registry type: nacos3
- Service count: 200+ in single namespace/gro
- Node CPU: ~74% allocated

Suggestions

Short-term (code fix)

1. Make DefaultNacosTimeout configurable via WatcherOption and McpBridge CRD (nacosTimeout field), similar to how
nacosRefreshInterval is already configurable.
2. Add exponential backoff in retry logic:
time.Sleep(time.Duration(tries) * time.Second
2. This allows the Nacos SDK time to reconnect between retries.
3. Increase DefaultNacosTimeout default from  for production use.

Long-term

4. Reduce mutex scope in fetchAllServices — ronly hold during WatchingServices mapread/write.
5. Debounce subscribe callbacks — batch rapidXDS push storms when recovering from a failureperiod.

Affected files

- registry/nacos/v2/watcher.go (primary)
- registry/nacos/watcher.go (same pattern)
- registry/nacos/mcpserver/watcher.go (same pattern)
- registry/reconcile/reconcile.go (needs With

Aias00:
Fixes #4293.

## What changed
- Added `nacosTimeout` to McpBridge registry config and exposed it in generated CRD schemas.
- Increased the default Nacos client timeout from 5s to 30s for Nacos v1, Nacos v2/v3, and the Nacos MCP watcher.
- Passed the configured timeout from the registry reconciler into the Nacos watchers.
- Added linear backoff for Nacos service-list fetch retries to avoid tight retry loops during transient SDK/gRPC failures.
- Moved Nacos service-list and MCP config list calls outside the watcher mutex so timeout/backoff waits do not block watcher state updates.

## Validation
- `buf generate --path networking --path extensions`
- `git diff --check`
- `kubectl apply --dry-run=client --validate=false -f api/kubernetes/customresourcedefinitions.gen.yaml`
- `kubectl apply --dry-run=client --validate=false -f helm/core/crds/customresourcedefinitions.gen.yaml`
- `ruby -e 'require "yaml"; YAML.load_stream(File.read(ARGV[0])); puts "ok"' helm/core/crds/customresourcedefinitions.gen_lt1.16.yaml`
- `go test -mod=mod -modfile=/tmp/higress-go-*.mod ./registry/nacos ./registry/nacos/v2 ./registry/nacos/mcpserver -run 'TestWithNacosTimeout|TestFetchRetryBackoff'`

Full targeted package test note: `go test -count=1 -mod=mod -modfile=/tmp/higress-go-full-*.mod ./registry/nacos ./registry/nacos/v2 ./registry/nacos/mcpserver` still hits existing `generateServiceEntry` assertions in `registry/nacos` and `registry/nacos/v2`; `registry/nacos/mcpserver` passes.


johnlanni (approved):
Approve — reviewed at exact head `1372173f555ae1b03519ca831f421e90c5db1885`.

Verified against #4293: the base implementation's hardcoded 5s `WithTimeoutMs`, immediate retry loop, and mutex held around paginated service discovery explain the reported deadline/UNHEALTHY/stall cascade. `nacosTimeout` is consistently wired as an int64 millisecond field through the proto, regenerated descriptor, all CRD schema copies, reconciler, nacos v1/v2 watchers, and nacos3 MCP watcher. Zero/negative values safely fall back to 30s; a registry change recreates the watcher through the existing DeepEqual-based reconcile path. The v1 fetch now fails closed rather than reconciling against a partial service list.

CI is green on this head. Non-blocking follow-ups: mention the 5s→30s default behavior change in release notes; v1 could gain v2's per-page shutdown check; and the separate golang-filter MCP Nacos client still has its own hardcoded 5s timeout outside this CRD/controller path.

Issue #3992, opened 2026-06-17
model-mapper: service constraint silently skips matching on McpBridge-based AI routes


Aaron199:
## Bug Description

When an AI route uses McpBridge as its backend (the default for Higress AI routes), the model-mapper WasmPlugin's `service` constraint fails to match, causing model name mapping rules to be silently skipped. This results in the original model name being forwarded to upstream providers unchanged, leading to 400 errors from providers that require different model names.

## Root Cause

The `service` field in WasmPlugin matchRules matches against the Envoy route's backend service. For McpBridge-based AI routes, the backend is `McpBridge/default`, not the actual upstream K8s Service (e.g., `llm-supplier-xxx.internal.dns`). Therefore, a model-mapper rule configured with:

```yaml
matchRules:
  - ingress:
      - ai-route-example.internal
    service:
      - llm-supplier-xxx.internal.dns    # never matches on McpBridge routes
    config:
      modelMapping:
        "claude-opus-4-6": "cd-opus-4.6"
```

will never apply, because the actual backend service observed at the WasmPlugin layer is `McpBridge/default`.

## Reproduce

1. Create an AI route with McpBridge backend and multiple upstreams
2. Configure model-mapper with a `service` constraint targeting one of the upstream services
3. Send a request matching that upstream
4. Observe that the model name is **not** transformed (the original model name reaches the upstream unchanged)
5. Check Gateway logs — no model-mapper execution for that route

## Expected Behavior

Either:
- The `service` constraint should correctly match against the route's configured upstream services on McpBridge routes, OR
- The Higress Console / controller should not generate `service`-constrained model-mapper rules for McpBridge-based routes (use empty `service: []` for global matching), OR
- A warning should be emitted when `service` constraints are used on McpBridge routes to alert users

## Workaround

Add `modelMapping` directly to the ai-proxy provider configuration in `defaultConfig.providers` instead of using the model-mapper WasmPlugin. The ai-proxy plugin resolves the provider correctly via `clusterProviderMap` and applies per-provider modelMapping during request processing.

## Environment

- Higress version: 2.2.2
- Route backend type: McpBridge (default)

CH3CHO:
Your understanding isn't right. When using McpBridge as the backend, there shall be an annotation `higress.io/destination` marking the actual upstream and its value will be used in WasmPlugin for matching. Please provide full steps to reproduce this issue.

PR #4295, opened 2026-08-07
test(ingress): cover McpBridge destination routes

Aias00:
Refs #3992.

## What changed
- Added regression coverage for Ingress v1 McpBridge resource backends using `higress.io/destination`.
- Added the same coverage for the legacy Ingress v1beta1 path.
- The tests assert that normal HTTP routes use the destination annotation's real upstream service (`llm-supplier-xxx.internal.dns:443`) rather than the `McpBridge/default` resource backend as the route destination.

## Notes
The current control-plane path already routes McpBridge resource backends through `higress.io/destination`; this PR locks that behavior with tests for the AI-route-style HTTP path described in #3992. If the reporter still sees model-mapper service matching fail, the issue likely needs a concrete reproduction where the generated route lacks or differs from the destination annotation.

## Validation
- `go test -mod=mod -modfile=/tmp/higress-3992-go-*.mod ./pkg/ingress/kube/ingress ./pkg/ingress/kube/ingressv1 -run TestHTTPRouteUsesMCPDestinationForResourceBackend -count=1`
- `go test -mod=mod -modfile=/tmp/higress-3992-full-*.mod ./pkg/ingress/kube/ingress ./pkg/ingress/kube/ingressv1 -count=1`
- `git diff --check`

### DIFF
--- a/pkg/ingress/kube/ingress/controller_test.go
+++ b/pkg/ingress/kube/ingress/controller_test.go
@@ -1026,6 +1026,77 @@ func TestSSLPassthroughIgnoresNonRootPath(t *testing.T) {
 	}
 }
 
+func TestHTTPRouteUsesMCPDestinationForResourceBackend(t *testing.T) {
+	c := controller{}
+	apiGroup := "networking.higress.io"
+	wrapper := &common.WrapperConfig{
+		Config: &config.Config{
+			Meta: config.Meta{
+				Namespace: "default",
+				Name:      "ai-route",
+			},
+			Spec: ingress.IngressSpec{
+				Rules: []ingress.IngressRule{
+					{
+						Host: "ai.example.com",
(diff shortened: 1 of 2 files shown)


johnlanni (approved):
Approve — reviewed at exact head `e947062719a56b91c93ec7738c92019bd45ff483`.

This is appropriately test-only and uses `Refs #3992`, not a closing keyword: the issue still needs a concrete end-user reproduction, while the test locks in the maintainer-confirmed controller behavior. Both legacy and `networking.k8s.io/v1` Ingress controllers cover a valid `McpBridge/default` resource backend and verify the generated route uses the real `higress.io/destination` upstream rather than `McpBridge/default`.

CI is green on this head. Non-blocking coverage opportunities: assert the route weight/total and add a weighted multi-destination case closer to the issue scenario.

Issue #4201, opened 2026-07-24
AI路由的重试插件和全局路由超时配置问题 || AI routing retry plug-in and global routing timeout configuration issues


yuiej:
### Ⅰ. Issue Description
higress全局配置
```yaml
downstream:
  idleTimeout: 1800
  routeTimeout: 3600
upstream:
  idleTimeout: 600
```
创建AI路由，请求服务，服务会在8min中止。
多次尝试后，发现在关闭重试插件后即可正常响应

### Ⅱ. Describe what happened
截取的部分网关日志
    "response_body": "upstream request timeout",
    "protocol": "HTTP/1.1",
    "response_flags": "URX,UT",
    "response_code_details": "upstream_per_try_timeout",
    "response_code": "504",
    "response_type": "normal",
    "duration": 480066,

### Ⅲ. Describe what you expected to happen
服务处理此次请求的时间会比较长，大概10min。希望能使用全局超时配置

### Ⅵ. Environment:

- Higress version:  2.2.2
- OS: k8s
- Others: 

---
### Ⅰ. Issue Description
higress global configuration
```yaml
downstream:
  idleTimeout: 1800
  routeTimeout: 3600
upstream:
  idleTimeout: 600
```
Create an AI route and request the service. The service will be terminated in 8 minutes.
After several attempts, I found that the response can be normal after closing the retry plug-in.

### Ⅱ. Describe what happened
Part of the intercepted gateway logs
    "response_body": "upstream request timeout",
    "protocol": "HTTP/1.1",
    "response_flags": "URX,UT",
    "response_code_details": "upstream_per_try_timeout",
    "response_code": "504",
    "response_type": "normal",
    "duration": 480066,

### Ⅲ. Describe what you expected to happen
The service will take a long time to process this request, about 10 minutes. Hope to use global timeout configuration

### Ⅵ. Environment:

- Higress version: 2.2.2
- OS: k8s
- Others:

Aias00:
Root cause traced against the v2.2.2 source.

The Console AI-route service supplies a retry default of `enabled=true`, `attempts=3`, and `timeout=120`, then writes `higress.io/proxy-next-upstream-timeout: "120"` to the generated Ingress:

- https://github.com/higress-group/higress-console/blob/534872f11ad8e4dbe85d665dc78ee1488012fd75/backend/sdk/src/main/java/com/alibaba/higress/sdk/service/ai/AiRouteServiceImpl.java#L80-L81
[...]

PR #4298, opened 2026-08-07
fix: let retry inherit route timeout by default

Aias00:
## What changed

- Do not set `HTTPRetry.PerTryTimeout` when retry annotations enable retry but `proxy-next-upstream-timeout` is not explicitly configured.
- Keep explicit `proxy-next-upstream-timeout` behavior unchanged.
- Update retry annotation tests to cover both default inheritance and explicit per-try timeout cases.

This lets retry-enabled routes continue to inherit the route/global timeout by default, which addresses #4201.

## Validation

- `go test -mod=mod -modfile=/tmp/higress-4201-go-*.mod ./pkg/ingress/kube/annotations -count=1`
- `git diff --check`

### DIFF
--- a/pkg/ingress/kube/annotations/retry.go
+++ b/pkg/ingress/kube/annotations/retry.go
@@ -50,9 +50,8 @@ func (r retry) Parse(annotations Annotations, config *Ingress, _ *GlobalContext)
 	}
 
 	retryConfig := &RetryConfig{
-		retryCount:      defaultRetryCount,
-		perRetryTimeout: &duration.Duration{},
-		retryOn:         defaultRetryOn,
+		retryCount: defaultRetryCount,
+		retryOn:    defaultRetryOn,
 	}
 	defer func() {
 		config.Retry = retryConfig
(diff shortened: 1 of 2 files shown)


johnlanni (changes requested):
Request changes — reviewed at exact head `eda92ee4ac320d8f068eeed0a900e282b26297c7`.

The Issue is clear and the annotation-level tests pass, but I cannot yet verify that this changes the generated Envoy behavior. On the pinned dependency path, Istio copies `HTTPRetry.PerTryTimeout` only when it is non-nil; however Envoy constructs its internal retry timeout with `PROTOBUF_GET_MS_OR_DEFAULT(..., per_try_timeout, 0)`. Consequently both the old explicit `0s` field and the new absent field resolve to `0ms`, and Envoy's timeout logic treats a zero per-try value as using the global route timeout.

Please add a regression at the relevant xDS/data-plane boundary that demonstrates the old configuration produces the observed `upstream_per_try_timeout` behavior and the new configuration inherits the configured route timeout. If another Higress layer distinguishes explicit `0s` from an absent field, identify it and cover that path. The current tests only assert Go-object field presence, so they do not establish that #4201 is fixed.

PR #4305, opened 2026-08-07
fix(ai-proxy): reduce default failure retry timeout

Aias00:
## What changed

- Align the `retryOnFailure.retryTimeout` default with the documented 30000ms value.
- Add a named default constant for failure retry timeout.
- Add coverage proving explicit `retryTimeout` values still override the default.

This is a narrow mitigation for #4179: when AI upstreams fail and `retryOnFailure` is enabled without an explicit timeout, retries now hold resources for 30s by default instead of 60s. It does not claim to cover the issue’s broader deployment tuning items such as circuit breaking, route-not-found polling, or body logging behavior.

## Validation

- `go test ./provider -count=1` from `plugins/wasm-go/extensions/ai-proxy`
- `git diff --check`

### DIFF
--- a/plugins/wasm-go/extensions/ai-proxy/provider/retry.go
+++ b/plugins/wasm-go/extensions/ai-proxy/provider/retry.go
@@ -7,14 +7,15 @@ import (
 	"net/http"
 
 	"github.com/alibaba/higress/plugins/wasm-go/extensions/ai-proxy/util"
+	"github.com/higress-group/proxy-wasm-go-sdk/proxywasm"
 	"github.com/higress-group/wasm-go/pkg/log"
 	"github.com/higress-group/wasm-go/pkg/wrapper"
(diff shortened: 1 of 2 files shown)


johnlanni (approved):
Approve — reviewed at exact head `5f27d0d40cf249bf722fb4604b5d9a0f9dda4f9e`.

This is a narrowly scoped mitigation for #4179 that resolves a verified documentation/code mismatch: `retryOnFailure.retryTimeout` was documented as 30000ms but the zero/omitted fallback was 60000ms. The new named 30000ms default preserves explicit positive values, and the timeout reaches `ClusterClient.Post` in milliseconds as intended.

I ran `go test ./provider -count=1` from `plugins/wasm-go/extensions/ai-proxy`; the default and explicit-override coverage pass, and `git diff --check` is clean. Non-blocking follow-up: negative retry timeout values still wrap through the existing `uint32` cast; that predates this PR.

Issue #4133, opened 2026-07-13
AI路由不支持一个https(忽略ssl证书校验的前提), 一个http进行负载均衡 || AI routing does not support an https (ignoring the premise of ssl certificate verification), an http for load balancing


lunzilunzi:
### Ⅲ. 描述您期望发生的情况
AI路由管理配置中, 一个路由配置两个目标AI服务
<img width="618" height="145" alt="Image" src="https://github.com/user-attachments/assets/9cb9e196-5b84-468a-adbc-1e08554351d3" />
一个是http协议的, 一个是https协议(需要跳过ssl验证)的, 希望访问路由地址时, 能按要求进行负载

### Ⅳ. 如何重现（尽量精简和准确）
1. AI服务提供者管理, 填写一个http协议的模型, 再填写一个https协议的模型
2. 创建一个路由, 在"目标AI服务"中填上两个模型
3. 填写注解, 跳过https校验 higress.io/backend-protocol=HTTPS
4. 访问路由. 路由到https的模型时, 返回正常, 路由到http模型时, 则出现错误
<img width="1175" height="508" alt="Image" src="https://github.com/user-attachments/assets/6814e41e-bb3d-4236-85a2-eb884e7cf143" />
5. 如果不填写注解, 则路由到http模型时, 返回正常, 路由到https模型时, 则出现错误
<img width="1212" height="492" alt="Image" src="https://github.com/user-attachments/assets/d74430c8-3625-459e-b2aa-a110f1442865" />

### Ⅴ. 其他需要补充的信息？
我们没办法提供证书信息, 所以想在基于忽略ssl验证的前提下完成这个负载均衡需求

### Ⅵ. 环境信息：

- Higress 版本：2.2.2
- 操作系统：docker for linux
- 其他：

---
### Ⅲ. Describe what you expect to happen
In the AI routing management configuration, one route is configured with two target AI services.
<img width="618" height="145" alt="Image" src="https://github.com/user-attachments/assets/9cb9e196-5b84-468a-adbc-1e08554351d3" />
One is for the http protocol, and the other is for the https protocol (ssl verification needs to be skipped). I hope that when accessing the routing address, the load can be carried out as required.

### Ⅳ. How to reproduce (try to be as concise and accurate as possible)
1. AI service provider management, fill in a model of http protocol, and then fill in a model of https protocol
2. Create a route and fill in the two models in "Target AI Service"
3. Fill in the annotation and skip https verification higress.io/backend-protocol=HTTPS
[...]

PR #4306, opened 2026-08-07
fix: support mixed protocols for mcp destinations

Aias00:
## What changed

- Allow `higress.io/destination` entries to carry `http://` or `https://` schemes while keeping route destinations as plain host/port.
- Preserve per-destination backend protocol metadata and make it trigger DestinationRule generation.
- Apply per-destination UpstreamTLS config for McpBridge resource backends, so HTTP and HTTPS targets in one route can get different traffic policies.
- Keep existing scheme-less `host[:port] [subset]` destination behavior unchanged.
- Add controller-level regression coverage for legacy Ingress and `networking.k8s.io/v1` Ingress, including port-level and service-level DestinationRules.
- Document the destination URI syntax and its interaction with `higress.io/backend-protocol` in Chinese and English.

This addresses #4133 by allowing mixed HTTP/HTTPS AI route targets without applying `higress.io/backend-protocol=HTTPS` to every target.

## Validation

- `go test -mod=mod -modfile=.codex_test.mod ./pkg/ingress/config ./pkg/ingress/kube/annotations ./pkg/ingress/kube/common ./pkg/ingress/kube/ingress ./pkg/ingress/kube/ingressv1 -count=1`
- `git diff --check origin/main...HEAD`


johnlanni (approved):
Approve — reviewed at exact head `f75ff4b3d4bd45f10df50d02c391838ef6319773`.

The linked report now provides a concrete mixed HTTP/HTTPS reproduction (Higress 2.2.2, Docker for Linux, `higress.io/backend-protocol=HTTPS`, expected/actual behavior). The follow-up adds the missing controller-level proof: `IngressConfig.convertDestinationRule` is exercised for both legacy Ingress and `networking.k8s.io/v1`, at port and service level, and asserts explicit HTTP remains plaintext while explicit HTTPS and a scheme-less entry inheriting `backend-protocol: HTTPS` produce SIMPLE TLS.

I also verified the parser/traffic-policy wiring and the Chinese/English docs: explicit schemes override the global backend protocol only for that destination, while omitted schemes retain the prior fallback. The repository CI is green, including build, coverage, lint, gateway/higress/inference conformance, CodeQL, DCO, and license checks.

One non-blocking follow-up: unknown URI schemes are not rejected by the current parser; they are treated as part of the host string. The docs correctly list only `http://` and `https://` as supported forms, but validation/error reporting for unsupported schemes can be improved separately.

PR #4307, opened 2026-08-07
fix(ai-proxy): reject oversized content length early

### LINKED ISSUE #3986: 当请求体大小比较大，如200M，higress内存压力会暴涨，即使预期出现HTTP 413请求被拦截，但 bytes_received 仍然是完整的 209,715,359 字节（约 200MB） || When the request body size is relatively large, such as 200M, the higress memory pressure will skyrocket. Even if the HTTP 413 request is expected to be intercepted, bytes_received is still a complete 209,715,359 bytes (about 200MB)
发送 200MB JSON 请求到 /v1/chat/completions 后：

返回 HTTP 413（response_code_details: "request_payload_too_large"）
但 bytes_received 仍然是完整的 209,715,359 字节（约 200MB）
请求耗时 ~2.5 秒
容器内存 额外冲高 1GB+（200MB body × 多个 Wasm VM 各持引用 + TCP 缓冲）
问题：413 是在完整接收 200MB 请求体后才返回的，不是"边收边拒绝"。

是否可以 超过限制立即拒绝（不缓冲完整 body），避免内存暴涨

---
After sending a 200MB JSON request to /v1/chat/completions:

Returns HTTP 413 (response_code_details: "request_payload_too_large")
But bytes_received is still a full 209,715,359 bytes (~200MB)
The request took ~2.5 seconds
Container memory increased to 1GB+ (200MB body × multiple Wasm VMs holding references + TCP buffer)
Problem: 413 is returned after completely receiving the 200MB request body, not "rejection while receiving".

Is it possible to reject the limit immediately (without buffering the complete body) to avoid memory explosion

Aias00:
## What changed

- Reject ai-proxy requests in the request-header phase when `Content-Length` is larger than the existing 100MB request body buffer limit.
- Return local HTTP 413 with `request_payload_too_large` before setting the wasm request body buffer limit.
- Keep the existing body-buffer path unchanged when `Content-Length` is missing, invalid, or within the limit.
- Handle `Content-Length` case-insensitively for the early reject check.

## Scope note

This only enables immediate rejection when the client declares an oversized `Content-Length`. For chunked/no-`Content-Length` requests, the gateway still cannot know the body is oversized before reading enough data, so rejection continues to depend on the configured Envoy/Wasm body buffer limit.

Fixes #3986.

## Validation

- `go test . -run 'Test_contentLengthExceedsLimit|TestOpenAI/openai_chat_completion_rejects_oversized|TestOpenAI/openai_chat_completion_request_headers' -count=1 -v`
- `git diff --check`

Note: local `go test ./... -count=1` and `go test . -count=1` were started, but did not complete within a reasonable local wait window and were interrupted with no failure output.

### DIFF
--- a/plugins/wasm-go/extensions/ai-proxy/main.go
+++ b/plugins/wasm-go/extensions/ai-proxy/main.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	"fmt"
+	"net/http"
 	"net/url"
 	"regexp"
 	"strconv"
@@ -35,6 +36,8 @@ const (
 	ctxOriginalAuth                = "original_auth"
 	ctxUpstreamErrorResponseStatus = "upstream_error_response_status"
 )
+
+const headerContentLength = "Content-Length"
 
 type pair[K, V any] struct {
 	key   K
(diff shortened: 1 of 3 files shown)


johnlanni (approved):
Approve — reviewed at exact head `bb5db27dacfdb86e4d63967e2ebd9a495da2de56`.

The early 413 applies only to requests already subject to the ai-proxy 100 MiB body-buffer limit (a body plus a `RequestBodyHandler`), before the plugin removes `Content-Length` and starts buffering. Missing, malformed, or negative `Content-Length` values safely retain the existing body-stage limit; chunked/no-length requests retain that fallback too.

Focused boundary tests and the OpenAI header flow pass in Go and Wasm modes. The header-flow tests verify `ActionPause` plus a local 413 before a body is delivered, and `git diff --check` is clean. Non-blocking: documenting that this prevents gateway/Wasm buffering but cannot pre-reject a chunked request would make the limitation explicit.

PR #4314, opened 2026-08-08
fix(mcp-server): allow custom server version

### LINKED ISSUE #4056: 在Higress创建的MCP服务Version固定为1.0.0无法修改
## Why do you need it?
 Is your feature request related to a problem? Please describe in details

当 MCP 服务经过多次迭代更新后，所有版本都显示为 1.0.0，无法通过版本号区分不同迭代的服务实例，不利于服务的版本追踪和管理。

## How could it be?
A clear and concise description of what you want to happen. You can explain more about input of the feature, and output of it.

在 Higress 控制台创建或编辑 MCP 服务时，允许用户自定义服务的 Version 字段。

Aias00:
## What changed

- Added an optional MCP server version field parsed from `server.version` for single servers and `toolSet.version` for composed servers.
- Kept the existing default `serverInfo.version` as `1.0.0` when the field is omitted or blank.
- Returned the configured version in the MCP `initialize` response under `serverInfo.version` instead of always returning `1.0.0`.
- Documented `server.version` in the mcp-server README.

Fixes #4056.

## Validation

- `go test ./server -count=1` from `plugins/wasm-go/pkg/mcp`
- `go test . -run '^TestRestMCPServerBasicFlow$' -count=1` from `plugins/wasm-go/extensions/mcp-server`
- `git diff --check`

Note: local `go test . -count=1` in `plugins/wasm-go/extensions/mcp-server` was also attempted. It reached the existing REST tool-call branch after the new initialize version test had passed, then crashed with a Go runtime `SIGBUS`; no assertion failure was reported before the runtime crash.

### DIFF
--- a/plugins/wasm-go/extensions/mcp-server/README.md
+++ b/plugins/wasm-go/extensions/mcp-server/README.md
@@ -11,13 +11,20 @@
 
 # 配置说明
 
-本插件通过编译期注册 MCP Server，**WasmPlugin 的 `defaultConfig` 通常无需额外字段**。具体工具列表、参数与鉴权由各子 Server 实现决定，开发新 MCP Server 请参考 [MCP Server 实现指南](../../mcp-servers/README.md)。
+本插件通过编译期注册 MCP Server，**WasmPlugin 的 `defaultConfig` 通常无需额外字段**。如需在 MCP `initialize` 响应中展示自定义服务版本，可在 `server.version` 中配置；未配置时默认返回 `1.0.0`。具体工具列表、参数与鉴权由各子 Server 实现决定，开发新 MCP Server 请参考 [MCP Server 实现指南](../../mcp-servers/README.md)。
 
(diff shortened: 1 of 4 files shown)


johnlanni (approved):
Approve — reviewed at exact head `4f3c6b7ac651d63912b5ca1ec016ae958d429bdb`.

This is a focused implementation of the feature request in #4056. `server.version` and composed `toolSet.version` share the same trim/default (`1.0.0`) behavior, and the `initialize` response now uses the parsed value while existing configurations keep their previous value.

Verified with focused parsing tests and the REST MCP initialize flow in both Go and compiled-Wasm modes; the latter asserts configured `2.5.0` in the actual `serverInfo.version` response. `git diff --check` is clean. Non-blocking documentation nit: README documents `server.version`; please add the equivalent `toolSet.version` form when documenting composed server configuration.

PR #4316, opened 2026-08-08
fix(ai-agent): format path arguments safely

### LINKED ISSUE #4299: bug(ai-agent): non-string path parameters can panic request processing
## Bug
`ai-agent` substitutes path-template arguments with `value.(string)`. Valid JSON tool arguments may be numbers or booleans, causing a Go panic and aborting request processing.

## Expected
Path arguments should be formatted consistently with query arguments and URL-escaped without unsafe type assertions.

## Reproduction
Use an API tool path such as `/users/{id}` and invoke it with `{id: 123}`.

Aias00:
### What this PR does

Fixes #4299.

This updates `ai-agent` tool path parameter substitution so JSON arguments are formatted the same way as query parameters before URL escaping. Non-string values such as numbers and booleans no longer panic request processing.

### Validation

```bash
go test . -run '^TestOnHttpResponseBody$' -count=1 -v
go test . -count=1
git diff --check
```

### DIFF
--- a/plugins/wasm-go/extensions/ai-agent/main.go
+++ b/plugins/wasm-go/extensions/ai-agent/main.go
@@ -446,7 +446,7 @@ func toolsCall(ctx wrapper.HttpContext, llmClient wrapper.HttpClient, llmInfo LL
 									// 删除已经使用过的
 									delete(data, param)
 									// 替换模板中的占位符
-									urlParts[i] = url.QueryEscape(value.(string))
+									urlParts[i] = url.QueryEscape(fmt.Sprintf("%v", value))
 								}
 							}
 						}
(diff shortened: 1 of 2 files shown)


johnlanni (approved):
Approve — reviewed at exact head `3a018a90776799839eb36de146a0d4be2ea2106b`.

I reproduced the original failure on the base: JSON `action_input` values are decoded as `float64`/`bool`, so the previous path substitution's `value.(string)` panics for `/users/{id}` with `{id: 123}`. The PR's regression test fails on the base with `interface conversion: interface {} is float64, not string` and passes at this head; the full `ai-agent` package test suite also passes.

Formatting with `fmt.Sprintf("%v", value)` before the existing `url.QueryEscape` preserves string behavior and deliberately matches the adjacent query-argument handling. The test could later assert the final URL too, but that is non-blocking.

PR #4317, opened 2026-08-08
fix(mcp): pass through REST authorization header

### LINKED ISSUE #4221: server.passthroughAuthHeader 配置不生效 || server.passthroughAuthHeader configuration does not take effect

Aias00:
### What this PR does

Fixes #4221.

This fixes `server.passthroughAuthHeader` for REST-to-MCP tools:

- when no downstream security is configured and `passthroughAuthHeader` is enabled, the incoming `Authorization` header is copied into the REST tool outbound request headers
- when passthrough is disabled, the existing fallback still removes the incoming `Authorization` header
- `RestMCPServer.Clone()` now preserves `passthroughAuthHeader` and default security settings so cloned REST servers do not lose the parsed auth behavior

### Validation

```bash
go test ./server -run 'TestAppendPassthroughAuthHeader|TestRestServer_Clone_Independence' -count=1 -v
go test ./server -count=1
go test ./... -count=1
git diff --check
```

### DIFF
--- a/plugins/wasm-go/pkg/mcp/server/rest_server.go
+++ b/plugins/wasm-go/pkg/mcp/server/rest_server.go
@@ -26,8 +26,8 @@ import (
 	"github.com/higress-group/proxy-wasm-go-sdk/proxywasm"
 	"github.com/tidwall/sjson"
 
-	"github.com/higress-group/wasm-go/pkg/log"
 	"github.com/alibaba/higress/plugins/wasm-go/pkg/mcp/utils"
+	"github.com/higress-group/wasm-go/pkg/log"
 	"github.com/higress-group/wasm-go/pkg/wrapper"
 )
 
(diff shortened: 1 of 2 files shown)


johnlanni (dismissed):
The REST Authorization passthrough and `Clone()` preservation changes are useful, but this head introduces a credential-forwarding regression that needs to be fixed before merge.

When effective downstream security is configured with a non-Authorization scheme (for example, an API key in `X-Api-Key` or a query parameter) and `passthroughAuthHeader` is disabled, the incoming client `Authorization` header must still be removed before routing to the REST backend. Current main does this through the fallback removal keyed on the tool-level security configuration. This PR changes that condition to the effective `downstreamSecurity.ID`; because that ID is non-empty, the fallback removal is skipped even though the configured scheme did not consume/remove `Authorization`. `RouteCall` preserves unmodified request headers, so the stray client credential can reach the backend.

Please preserve this invariant:

- passthrough enabled: copy/preserve the incoming Authorization value, unless configured upstream security intentionally replaces it;
- passthrough disabled: remove any remaining incoming Authorization header after downstream credential processing, regardless of whether the effective downstream scheme itself is non-empty.

Please also add a regression test through the production `RestMCPTool.Call`/routing path for a non-Authorization downstream security scheme with `passthroughAuthHeader=false`, asserting that the backend receives no stray Authorization header. The existing helper and Clone tests do not cover this branch.

Reviewed at exact head `19be8c71f64c6c938b058a503262623275d9529c`.

PR #4320, opened 2026-08-08
fix(transformer): skip unmatched pattern rules

### LINKED ISSUE #4132: 请求/响应转换插件, 不同匹配路径下, 无法为相同header的key设置不同的值, 只有第一次生效 || Request/response conversion plug-in, under different matching paths, different values cannot be set for the key of the same header, and it will only take effect the first time
### Ⅲ. 描述您期望发生的情况
插件设置全局配置, 想让/test和/prod开头的地址, 都添加一个access-appId的请求头, 
/test路径, access-appId值为ij2b6Nfg
/prod路径, access-appId值为Y6JK5J7T
实际情况, 同一个header里, 如果key相同, 那么只有第一个key是生效的

### Ⅳ. 如何重现（尽量精简和准确）

1. 进入首页 -> 插件配置 -> 请求/响应转换 -> 配置
2. 填写"请求/响应转换"配置
```
reqRules:
- operate: add
  headers:
  - key: access-appId
    value: "ij2b6Nfg"
    host_pattern: "^/test/.*"
  - key: access-appId
    value: "Y6JK5J7T"
    host_pattern: "^/prod/.*"
```
### Ⅴ. 其他需要补充的信息？

由于access-appId在请求头和请求路径参数中都能生效, 所以最终我是这样改的, 最终效果能解决我的问题, 但是这个bug希望修复
```
reqRules:
- operate: add
  headers:
  - key: "access-appId"
    value: "ij2b6Nfg"
    host_pattern: "^/test/.*"
- operate: add
  querys:
  - key: "access-appId"
    value: "Y6JK5J7T"
    host_pattern: "^/prod/.*"
```

### Ⅵ. 环境信息：

- Higress 版本：2.2.2
- 操作系统：docker for linux
- 其他：

---
### Ⅲ. Describe what you expect to happen
The plug-in sets the global configuration. I want the addresses starting with /test and /prod to add an access-appId request header.
/test path, access-appId value is ij2b6Nfg
/prod path, access-appId value is Y6JK5J7T
In actual situation, if the keys in the same header are the same, only the first key is valid.

### Ⅳ. How to reproduce (try to be as concise and accurate as possible)

1. Go to the home page -> Plug-in configuration -> Request/response conversion -> Configuration
2. Fill in the "Request/Response Conversion" configuration
```
reqRules:
- operate: add
  headers:
  - key: access-appId
    value: "ij2b6Nfg"
    host_pattern: "^/test/.*"
  - key: access-appId
    value: "Y6JK5J7T"
    host_pattern: "^/prod/.*"
```
### Ⅴ. Any other information that needs to be supplemented?

Since the access-appId can take effect in the request header and request path parameters, I finally changed it like this. The final effect can solve my problem, but I hope this bug can be fixed.
```
reqRules:
- operate: add
  headers:
  - key: "access-appId"
    value: "ij2b6Nfg"
    host_pattern: "^/test/.*"
- operate: add
  queries:
  - key: "access-appId"
    value: "Y6JK5J7T"
    host_pattern: "^/prod/.*"
```

### Ⅵ. Environmental information:

- Higress version: 2.2.2
- Operating system: docker for linux
- Others:

Aias00:

Fixes transformer pattern-scoped operations so `host_pattern` / `path_pattern` act as match conditions. If a pattern does not match the current request host/path, the corresponding `replace`, `add`, or `append` item is skipped instead of writing its literal value.

This allows multiple same-name header rules with different path patterns to select the matching value, instead of the first item creating the header and blocking later matching items.

fixes #4132

Added focused transformer regression tests for unmatched path patterns and same-name headers with different path patterns.

```bash
go test . -run '^TestRequest_Headers_AddWithPathPattern_NoMatchSkipsAdd$' -count=1 -v
go test . -run '^TestRequest_Headers_AddSameHeaderWithDifferentPathPatterns$' -count=1 -v
go test . -count=1
git diff --check
```

All commands were run from `plugins/wasm-go/extensions/transformer`.

This changes the previous fallback behavior where an unmatched pattern still wrote the unexpanded literal value. The documented meaning of `host_pattern` / `path_pattern` is a match rule, so skipping unmatched items avoids leaking values from unrelated path-specific rules.

Handle Higress issue #4132 and keep the change focused with tests and DCO.

### AI Coding Summary

- Changed transformer regex-scoped `replace`, `add`, and `append` handling to skip unmatched `host_pattern` / `path_pattern` items.
- Added regression coverage for unmatched path patterns and same-name headers with different path patterns.
- Verified the transformer package tests pass in both Go and wasm test modes.

### DIFF
--- a/plugins/wasm-go/extensions/transformer/main.go
+++ b/plugins/wasm-go/extensions/transformer/main.go
@@ -934,7 +934,11 @@ func (h kvHandler) handle(host, path string, kvs map[string][]string, mapSourceD
 			for _, replace := range kvtOp.replaceKvtGroup {
 				key, newValue := replace.key, replace.newValue
 				if replace.reg != nil {
-					newValue = replace.reg.matchAndReplace(newValue, host, path)
+					var matched bool
+					newValue, matched = replace.reg.matchAndReplace(newValue, host, path)
+					if !matched {
+						continue
+					}
 				}
 				kvs[key] = []string{newValue}
 			}
@@ -946,7 +950,11 @@ func (h kvHandler) handle(host, path string, kvs map[string][]string, mapSourceD
 					continue
 				}
 				if add.reg != nil {
-					value = add.reg.matchAndReplace(value, host, path)
+					var matched bool
+					value, matched = add.reg.matchAndReplace(value, host, path)
+					if !matched {
+						continue
+					}
 				}
 				kvs[key] = []string{value}
 			}
@@ -956,7 +964,11 @@ func (h kvHandler) handle(host, path string, kvs map[string][]string, mapSourceD
 			for _, append_ := range kvtOp.appendKvtGroup {
 				key, appendValue := append_.key, append_.appendValue
 				if append_.reg != nil {
-					appendValue = append_.reg.matchAndReplace(appendValue, host, path)
+					var matched bool
+					appendValue, matched = append_.reg.matchAndReplace(appendValue, host, path)
+					if !matched {
+						continue
+					}
 				}
 				kvs[key] = append(kvs[key], appendValue)
 			}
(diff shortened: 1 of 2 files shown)


johnlanni (approved):
Approve — reviewed at exact head `2eff8b77aaaf087a50a459f1775feba26c4b0281`.

The original same-header/pattern failure was independently reproduced on the base: for `/prod/chat`, the unmatched first `add` rule claimed the key, so the matching second rule could not apply. The focused regression tests fail on the base and pass at this head in both Go and wasm modes. The six relevant string-pattern call sites (kv/JSON `replace`, `add`, and `append`) consistently skip an unmatched rule, while rules without a pattern are unchanged.

Non-blocking follow-ups: documenting the no-match=`skip` behavior explicitly and adding host-pattern/replace-or-append coverage would make the contract clearer.

PR #4318, opened 2026-08-08
feat: support Gateway API Inference Extension v1.4 conformance

EndlessSeeker:
## What changed

This PR adds the Higress runtime support and a required CI gate for Gateway API Inference Extension v1.4.0.

Runtime support:

- create deterministic synthetic Service ports for every InferencePool targetPort;
- route InferencePool HTTPRoute backends to the first synthetic cluster port;
- preserve InferencePool-specific ports in controller conversion tests;
- update Envoy, go-control-plane, and Istio submodules for served-endpoint metadata and data-parallel endpoint aggregation;
- align the golang-filter Envoy Go bindings with the updated Envoy submodule.

CI integration:

- build Higress Controller and Pilot from the PR source in a clean GitHub Actions runner;
- create an isolated Kind and MetalLB environment;
- install the exact upstream GIE tag, Gateway API CRDs, InferencePool CRDs, Istio CRDs, EPP TLS wiring, and Higress;
- pin and verify the prevalidated Envoy 1.36 gateway image by SHA256 digest;
- run the unmodified upstream conformance suite and fail unless the report has the expected pass count with zero failures and zero skips;
- upload the conformance report and diagnostics, then clean the Kind environment;
- make this job a dependency of publish.

The workflow uses a version matrix. Adding v1.5.0 later only requires another matrix row containing the GIE version, matching Gateway API and Go versions, expected pass count, and gateway image tag/digest. Each GIE tag is downloaded into a version-specific source directory and produces a version-specific report artifact.

## Dependency PRs

- higress-group/envoy#31
- higress-group/go-control-plane#4
- higress-group/istio#67

## Validation

All native data-plane builds were performed on a remote Linux/amd64 development host. No Envoy image was built on macOS.

Focused validation:

- Higress pkg/ingress/kube/gateway/istio package: passed;
- go-control-plane override-host generated package: compiled;
- Istio InferencePool model/cluster/EDS focused tests: passed;
- Envoy override-host config/load-balancer/integration tests: 3/3 passed;
- golang-filter built with Go 1.23.12 against Envoy dc8a83fe53d0 and embedded in the ACK gateway image.

ACK pre-validation:

- Gateway API Inference Extension: exact v1.4.0 tag;
- Gateway API: v1.5.0 standard CRDs;
- Higress base: v2.2.3 plus this change;
- Envoy: 1.36.4;
- result: 12 passed, 0 failed, 0 skipped.

[...]

PR #4352, opened 2026-08-08
feat(wasm): add an AI endpoint picker with configurable prefix precision

### LINKED ISSUE #4296: Proposal: 为 ai-load-balancer 引入可组合的智能调度框架 || Proposal: Introducing a composable intelligent scheduling framework for ai-load-balancer
## 摘要

建议在 `ai-load-balancer` 中引入一个可组合、可观测、可渐进降级的智能调度框架，将当前相互独立的队列长度、KV Cache 使用率、LoRA 亲和、前缀亲和、并发数、延迟和错误反馈等能力，统一到 **Filter → Normalize → Score → Pick → Feedback** 流水线中。

这个提案不是增加又一个孤立的负载均衡算法，而是建立稳定的调度抽象，使 Higress 可以同时利用多个信号做一次可解释的决策，并在信号缺失、过期或冲突时安全退化。现有策略继续兼容；#4050 的 `AdaptiveScore` 可作为该框架的首个集群级评分 Profile，而 #4289/#4290 的内置 endpoint metrics 可作为信号来源之一。

## 背景与现状

当前 `ai-load-balancer` 的实现更接近“一次选择一个原子策略”：

- endpoint 级策略可以按队列、LoRA、KV Cache 使用率等条件逐级筛选；
- `prefix_cache` 通过 Redis 记录历史请求与后端映射，属于流量历史亲和，不代表推理引擎当前真实持有的 KV block；
- `cluster_metrics` 与拟议中的 `AdaptiveScore` 面向集群级延迟、并发和失败反馈；
- Gateway API Inference Extension 的内置模式（#4289/#4290）计划提供 endpoint metrics，但 Phase 1 仍以指标采集和单项选择为主。

这些能力各自有价值，但缺少一个统一模型回答以下问题：一个后端前缀命中最好、但队列已经很长时应该如何取舍；某项指标过期时是否还应参与决策；多个候选分数如何归一化；选择结果如何用真实完成状态持续校准。

## 目标

1. 让多种异构信号在一次调度决策中组合，而不是只能选择单一策略或固定级联。
2. 明确区分硬约束、软偏好和选择算法，避免将“不健康”和“得分较低”混为一谈。
3. 为 endpoint 级和 cluster 级调度提供同一套阶段语义和可复用接口。
4. 对每项信号记录来源、采样时间、置信度和缺失策略，使降级行为可预测。
5. 保持现有配置和行为兼容，允许用户按路由/模型池渐进启用。
6. 提供决策可观测性和离线回放能力，使权重调整有数据依据。

## 非目标

- 本提案不试图在第一阶段自动学习权重或引入强化学习。
- 不要求第一阶段必须依赖 Redis 或其他全局状态存储。
- 不替代 Envoy 的连接、重试、熔断和被动健康检查能力。
- 不在本提案中规定 vLLM KVEvents 的接入细节；该能力由独立提案提供真实前缀缓存信号。
- 不改变已经显式选择现有原子策略的用户行为。

### 1. Filter：先处理不可选条件

Filter 仅表达硬约束，例如：

- endpoint 健康状态、熔断/冷却状态；
- 模型、LoRA adapter、硬件或协议能力匹配；
- 指标或事件状态超过允许的最大陈旧时间；
- endpoint 已达到明确的队列或并发硬上限。

Filter 应返回候选集合以及排除原因。所有候选都被过滤时，框架必须按配置执行 fail-open 或回退到 Envoy 默认负载均衡，不能因可选遥测异常直接中断请求。

### 2. Normalize：把异构信号转成可组合值

每个 Signal Provider 返回原始值，并携带：

- `source`：Prometheus、请求生命周期反馈、真实 KV event、静态能力等；
- `observed_at` / `freshness`；
- `confidence`；
- `missing_reason`。

Normalize 将“越小越好”的队列/延迟/使用率和“越大越好”的前缀命中率/亲和度转换为统一的 `[0, 1]` 区间。每个信号必须定义边界、异常值处理、陈旧衰减和缺失策略，避免简单拼接不可比的原始数值。

### 3. Score：按 Profile 组合软偏好

建议使用可解释的加权评分作为第一版：

```text
final_score(endpoint) = Σ(weight_i × normalized_signal_i × confidence_i)
                        - overload_penalty
                        - failure_penalty
```

框架应支持少量内置 Profile，而不是一开始暴露大量难以校准的参数，例如：

- `balanced`：兼顾缓存收益和排队/并发压力；
- `cache-first`：适合长公共前缀、缓存重算成本高的流量；
- `latency-first`：更重视队列、TTFT 和尾延迟；
- `adaptive-cluster`：承接 #4050 的 EWMA 延迟、inflight 和失败惩罚模型。

权重和非线性惩罚应有版本，并能在指标中暴露实际采用的 Profile。高级自定义权重可以后续增加，不应成为第一阶段的前置条件。

### 4. Pick：从高质量候选中做稳定选择

Pick 与 Score 解耦，第一阶段可支持：

- `max-score`：选择最高分；
- `top-k-p2c` 或在 Top-K 内加权随机：降低同分和瞬时热点导致的请求同步聚集；
- 对同分候选使用稳定哈希，减少不必要的抖动。

Pick 必须保留最终选择原因，并对候选为空、全部低置信度和并列等情况定义确定的回退行为。

### 5. Feedback：用真实请求结果更新短期状态

在请求开始、首 token、流结束/取消和上游失败时，向框架反馈：

- 本地 inflight；
- TTFT、总时延和流式持续时间的 EWMA；
- 成功、超时、取消和可重试错误；
- 实际选择的 endpoint、使用的 Profile 与主要得分项。

第一阶段允许每个 gateway replica 维护本地反馈。需要跨 gateway 一致性的场景可后续接入共享状态，但共享存储不可成为基础框架的强依赖。

## 接口与兼容性建议

- 为框架新增独立策略名（建议在 Design 阶段确定，例如 `composite`），现有 `round_robin`、`least_request`、`prefix_cache`、`endpoint_metrics`、`cluster_metrics` 等配置不变。
- Signal Provider、Normalizer、Scorer、Picker 和 Feedback Sink 使用稳定的小接口，数据快照只读且带版本，便于在 WASM 与网关内置原生组件之间复用语义。
- endpoint 与 cluster 使用同一阶段模型，但候选 ID 和信号集合可不同；不强制将两者放进同一个实现对象。
- #4050 的 `AdaptiveScore` 不再演化成平行框架，而是作为 `cluster_metrics` 信号适配器和内置 Profile 接入。
- #4289/#4290 提供的主动指标采集成为 provider，不改变其“无 EPP 的内置模式”和“有 EPP 的外部模式”边界。

## 可观测性

至少暴露以下低基数指标，并提供采样调试日志：

- 每个 Profile 的请求数和回退次数；
- 各 Filter 排除候选的计数；
- signal missing/stale/invalid 计数；
- 被选 endpoint 的总分分布和主要贡献项；
- Pick 结果、请求结果、TTFT/总时延和错误率；
- 决策耗时以及候选数量。

不得默认将 prompt、token 或高基数 endpoint 明细写入普通日志。调试信息需可采样并可关闭。

### Phase 1：框架骨架与兼容适配

- 实现 Filter、Normalize、Score、Pick、Feedback 的阶段接口和请求级决策上下文；
- 接入现有健康、队列、KV 使用率、LoRA、inflight、延迟和错误信号；
- 提供 `balanced` Profile、`max-score`/Top-K Pick、陈旧/缺失降级和观测指标；
- 保持现有原子策略回归行为不变。

### Phase 2：真实前缀缓存信号

- 接入独立 vLLM KVEvents 提案提供的精确 per-endpoint prefix cache score；
- 对比真实事件评分、现有 Redis 历史亲和以及无缓存感知基线。

### Phase 3：校准与扩展

- 根据回放和 benchmark 固化 `cache-first`、`latency-first`、`adaptive-cluster` Profile；
- 评估跨 gateway 聚合、动态权重或更多推理引擎 provider，需另行提案。

## 验收标准

- 同一请求可以同时基于至少“缓存收益”和“负载压力”两个独立信号评分，且两项贡献可观测。
- 单项指标缺失或过期不会导致请求失败，并能按既定策略降级。
- 相同输入快照和配置产生可复现的 Normalize/Score 结果；Pick 的随机部分可通过固定种子测试。
- 现有负载均衡策略的配置和回归测试保持不变。
- 提供离线决策回放/单元测试夹具，覆盖前缀命中高但队列过载、低负载但无缓存、指标过期、全部候选被过滤等场景。
- 基准测试证明框架开销受控，并分别报告 TTFT、P95/P99 延迟、吞吐、缓存命中收益和负载偏斜，避免仅以单一指标判断效果。

## 相关工作

- #4050：`AdaptiveScore` 集群级负载均衡模式，可作为本框架的首个集群级 Profile。
- #4289 / #4290：GIE 内置 `ai-load-balancer` Proposal/Design，可作为 endpoint 指标和配置入口。
- #3407：基于 KV Cache 感知的负载均衡讨论。
- #3810：前缀缓存亲和相关能力。

---

## Summary

It is recommended to introduce a composable, observable, and progressively degraded intelligent scheduling framework into `ai-load-balancer`, and unify the currently independent capabilities such as queue length, KV Cache usage, LoRA affinity, prefix affinity, concurrency, delay and error feedback into the **Filter → Normalize → Score → Pick → Feedback** pipeline.

This proposal is not to add another isolated load balancing algorithm, but to establish a stable scheduling abstraction so that Higress can make an explainable decision using multiple signals at the same time, and safely degrade when signals are missing, expired, or conflicting. Existing policies continue to be compatible; #4050's `AdaptiveScore` is available as the framework's first cluster-level scoring Profile, and #4289/#4290's built-in endpoint metrics are available as one of the signal sources.

## Background and current situation

The current implementation of `ai-load-balancer` is closer to "select one atomic strategy at a time":

- Endpoint-level policies can be filtered step by step based on queue, LoRA, KV Cache usage and other conditions;
- `prefix_cache` records historical requests and backend mapping through Redis, which belongs to traffic history affinity and does not represent the KV block currently held by the inference engine;
- `cluster_metrics` and the proposed `AdaptiveScore` for cluster-level latency, concurrency and failure feedback;
- The built-in mode (#4289/#4290) of the Gateway API Inference Extension is planned to provide endpoint metrics, but Phase 1 is still focused on metric collection and single-item selection.

Each of these capabilities has value, but there is a lack of a unified model to answer the following questions: how to choose when a backend prefix hits the best but the queue is already very long; whether a certain indicator should still participate in the decision-making when it expires; how to normalize multiple candidate scores; how to continuously calibrate the selection results with the real completion status.

## Target

1. Let multiple heterogeneous signals be combined in one scheduling decision, instead of having to choose only a single strategy or fixed cascade.
2. Clearly distinguish between hard constraints, soft preferences and selection algorithms to avoid confusing “unhealthy” with “lower scores”.
3. Provide the same set of stage semantics and reusable interfaces for endpoint-level and cluster-level scheduling.
4. Record the source, sampling time, confidence level, and missing policy for each signal to make degradation behavior predictable.
5. Maintain compatibility with existing configurations and behaviors, allowing users to enable progressive activation by route/model pool.
6. Provide decision-making observability and offline playback capabilities so that weight adjustments are based on data.

## non-target

- This proposal does not attempt to automatically learn weights or introduce reinforcement learning in the first stage.
- There is no requirement that the first stage must rely on Redis or other global state stores.
- Does not replace Envoy's connection, retry, circuit breaking and passive health checking capabilities.
- Access details for vLLM KVEvents are not specified in this proposal; this capability is provided by a separate proposal to provide true prefix cache signaling.
- Does not change the behavior of users who have explicitly selected an existing atomic policy.

### 1. Filter: handle non-optional conditions first

Filter only expresses hard constraints, for example:

- endpoint health status, fuse/cooling status;
- Model, LoRA adapter, hardware or protocol capability matching;
- The indicator or event status exceeds the maximum allowed staleness;
- The endpoint has reached an explicit queue or concurrency hard cap.

Filter should return a set of candidates along with reasons for exclusion. When all candidates have been filtered, the framework must perform a fail-open or fallback to Envoy's default load balancing as configured, and cannot directly interrupt requests due to optional telemetry exceptions.

### 2. Normalize: Convert heterogeneous signals into combinable values

Each Signal Provider returns the original value, carrying:

- `source`: Prometheus, request life cycle feedback, real KV events, static capabilities, etc.;
- `observed_at` / `freshness`;
- `confidence`;
- `missing_reason`.

Normalize converts "smaller is better" queue/latency/usage and "bigger is better" prefix hit/affinity into a uniform `[0, 1]` interval. Boundaries, outlier handling, stale attenuation and deletion strategies must be defined for each signal to avoid simply splicing incomparable raw values.

### 3. Score: Combine soft preferences by Profile

It is recommended to use interpretable weighted scoring as a first version:

```text
final_score(endpoint) = Σ(weight_i × normalized_signal_i × confidence_i)
                        - overload_penalty
                        - failure_penalty
```

The framework should support a small number of built-in profiles instead of initially exposing a large number of parameters that are difficult to calibrate, such as:

- `balanced`: taking into account cache revenue and queuing/concurrency pressure;
- `cache-first`: suitable for traffic with long public prefixes and high cache recalculation costs;
- `latency-first`: pay more attention to queue, TTFT and tail latency;
- `adaptive-cluster`: inherits the EWMA delay, inflight and failure penalty models from #4050.

Weights and non-linear penalties should be versioned and the actual adopted profile exposed in the indicator. Advanced custom weights can be added later and should not be a prerequisite for the first phase.

### 4. Pick: Make stable selections from high-quality candidates

Pick and Score are decoupled, and the first stage can support:

- `max-score`: select the highest score;
- `top-k-p2c` or weighted random within Top-K: reduce the synchronization aggregation of requests caused by the same score and instant hot spots;
- Use stable hashing for identical candidates to reduce unnecessary jitter.

Pick must retain the final selection reason and define certain fallback behavior for cases where candidates are empty, all have low confidence, and are tied.

### 5. Feedback: Update short-term status with real request results

Feedback to the framework on request start, first token, flow end/cancellation, and upstream failure:

- local inflight;
- EWMA for TTFT, total latency and streaming duration;
- Success, timeout, cancellation and retryable errors;
- The actual selected endpoint, used Profile and main scoring items.

The first phase allows each gateway replica to maintain local feedback. Scenarios that require cross-gateway consistency can subsequently access shared state, but shared storage cannot become a strong dependency on the basic framework.

## Interface and compatibility recommendations

- Add a new independent strategy name to the framework (it is recommended to determine it in the design stage, such as `composite`), and the existing `round_robin`, `least_request`, `prefix_cache`, `endpoint_metrics`, `cluster_metrics` and other configurations remain unchanged.
- Signal Provider, Normalizer, Scorer, Picker and Feedback Sink use stable small interfaces, and data snapshots are read-only and versioned, which facilitates the reuse of semantics between WASM and the built-in native components of the gateway.
- Endpoint and cluster use the same stage model, but the candidate ID and signal set can be different; they are not forced to be put into the same implementation object.
- #4050's `AdaptiveScore` is no longer evolved into a parallel framework, but is instead plugged in as a `cluster_metrics` signal adapter and built-in Profile.
- #4289/#4290 Active metric collection provided by the provider does not change its "built-in mode without EPP" and "external mode with EPP" boundaries.

## Observability

Expose at least the following low-base metrics and provide sampled debug logs:

-The number of requests and rollbacks for each Profile;
- Count of excluded candidates for each Filter;
- signal missing/stale/invalid count;
- The total score distribution and main contributions of the selected endpoints;
- Pick results, request results, TTFT/total delay and error rate;
- Decision time and number of candidates.

Prompt, token, or high-cardinality endpoint details must not be written to the normal log by default. Debugging information needs to be sampled and turned off.

### Phase 1: Framework skeleton and compatible adaptation

- Implement stage interfaces and request-level decision context for Filter, Normalize, Score, Pick, and Feedback;
- Access to existing health, queue, KV usage, LoRA, inflight, latency and error signals;
- Provides `balanced` Profile, `max-score`/Top-K Pick, stale/missing degradation and observation indicators;
- Keep existing atomic strategy regression behavior unchanged.

### Phase 2: Real prefix cache signal

-Access to the precise per-endpoint prefix cache score provided by the independent vLLM KVEvents proposal;
- Compare real event scores, existing Redis history affinity, and cache-less awareness baselines.

### Phase 3: Calibration and Expansion

- Solidify `cache-first`, `latency-first`, `adaptive-cluster` Profile based on playback and benchmark;
- Evaluating cross-gateway aggregation, dynamic weighting, or more inference engine providers requires separate proposals.

## Acceptance criteria

- The same request can be scored based on at least two independent signals of "cache revenue" and "load pressure" at the same time, and both contributions are observable.
- Missing or expired individual indicators will not cause the request to fail, and can be downgraded according to the established strategy.
- Identical input snapshots and configurations produce reproducible Normalize/Score results; the random portion of Pick can be tested with a fixed seed.
- Configuration and regression testing of existing load balancing policies remain unchanged.
- Provide offline decision playback/unit testing fixtures, covering scenarios such as high prefix hits but queue overload, low load but no cache, indicator expiration, and all candidates being filtered.
- Benchmark tests prove that the framework overhead is under control, and report TTFT, P95/P99 latency, throughput, cache hit revenue and load skew separately to avoid judging the effect only by a single metric.

## Related work

- #4050: `AdaptiveScore` cluster-level load balancing mode can be used as the first cluster-level Profile of this framework.
- #4289 / #4290: GIE has built-in `ai-load-balancer` Proposal/Design, which can be used as endpoint indicator and configuration entry.
- #3407: KV Cache-aware load balancing discussion.
- #3810: Prefix cache affinity related capabilities.

### LINKED ISSUE #4346: Design: 新增独立的 ai-endpoint-picker 插件 || Design: Added independent ai-endpoint-picker plug-in
- Proposal Issue: 4296

## 设计摘要

本设计通过新增独立 WASM Go 插件 `ai-endpoint-picker` 落地 #4296，而不是继续扩展已经承担 cluster/endpoint、Redis、指标解析和多种历史策略的 `ai-load-balancer`。

插件只负责单个 upstream cluster 内的 endpoint 选择，采用与 llm-d Endpoint Picker 相同的核心分层：**Filter → Normalize/Score → Pick → Feedback**。首个版本提供 `balanced` Profile，组合 vLLM queue depth、KV cache utilization、LoRA affinity、gateway-local inflight 和失败反馈；后续 #4297 的真实 prefix cache score 通过稳定 Signal Provider 接口加入，不改动调度主流程。

插件名选择 `ai-endpoint-picker`，原因是它与 GIE 的 `endpointPickerRef`/EPP 术语直接对齐；`Router` 在 llm-d 中表示 Proxy + EPP 的整体，且 Higress 已有 `model-router`，使用 `router` 容易产生边界歧义。

## 参考与差异

设计参考 llm-d Router commit `aa0e31371891f283ceb56cfe3dbddbfdb62fb648`：

- `SchedulerProfile` 顺序执行 filters、weighted scorers、picker；
- scorer 输出 `[0,1]`，按权重累加；
- queue scorer 使用候选集合 min/max 归一化，队列越短得分越高；
- KV utilization scorer 使用 `1-utilization`；
- max-score picker 在同分候选间随机打散。

参考源码：

- https://github.com/llm-d/llm-d-router/blob/aa0e31371891f283ceb56cfe3dbddbfdb62fb648/pkg/epp/scheduling/scheduler_profile.go
- https://github.com/llm-d/llm-d-router/tree/aa0e31371891f283ceb56cfe3dbddbfdb62fb648/pkg/epp/framework/plugins/scheduling

Higress 版本不复制 EPP 的 Kubernetes controller、ext_proc server、flow control 和动态插件装载，而是在 WASM 请求路径内完成 endpoint snapshot、评分和 `SetUpstreamOverrideHost`，从而保持零额外 Deployment。

### 本次实现

- 新增 `plugins/wasm-go/extensions/ai-endpoint-picker/`，作为独立 Go module 和插件镜像。
- 从 `proxywasm.GetUpstreamHosts()` 获取候选 endpoint、健康状态和主动抓取的 metrics snapshot。
- 解析当前和旧版 vLLM 指标名，并允许 LoRA 指标缺失。
- 实现可测试的 Filter、Signal、Normalize、Score、Pick 数据模型。
- 实现 `balanced` Profile 和 fail-open 回退。
- 维护 gateway-local endpoint feedback：inflight、TTFT、总时延和失败 EWMA。
- 提供低基数自定义指标、采样 debug 日志、中英文 README、配置示例和单元测试。

### 非本次实现

- vLLM ZMQ KVEvents 与精确 prefix block 索引（#4297）。
- 独立队列、优先级、公平性、admission control 或完整 llm-d flow control。
- cluster 间/provider 间选择；仍由现有 cluster 级能力负责。
- GIE controller 自动生成 WasmPlugin 绑定；本次只稳定插件配置和语义，为后续控制面绑定提供目标。
- 动态加载任意用户 scorer 或在线学习权重。

## 目录与组件

```text
plugins/wasm-go/extensions/ai-endpoint-picker/
├── main.go                    # WASM 生命周期、请求/响应回调、override host
├── config.go                  # 用户配置、默认 Profile、校验
├── metrics.go                 # 低基数 proxy-wasm 指标
├── scheduling/
│   ├── types.go               # EndpointSnapshot、SignalValue、Decision、Feedback
│   ├── parser.go              # Prometheus/vLLM metrics 解析
│   ├── pipeline.go            # Filter → Normalize → Score → Pick
│   ├── feedback.go            # gateway-local endpoint 状态与 EWMA
│   └── *_test.go
├── go.mod / go.sum
├── VERSION
├── README.md
└── README_EN.md
```

`scheduling` 不依赖 proxy-wasm hostcall，使用纯 Go 输入输出，便于单元测试和未来被原生 gateway filter 复用。只有 `main.go` 负责 SDK 交互。

## 核心数据模型

```go
type EndpointSnapshot struct {
    Address string
    Healthy bool
    Signals map[SignalName]SignalValue
}

type SignalValue struct {
    Value      float64
    Available  bool
    Confidence float64
    Source     string
    Reason     string
}

type Decision struct {
    Address       string
    Score         float64
    CandidateCount int
    FallbackReason string
}
```

MVP Signal 集合固定为：

- `queue`：`vllm:num_requests_waiting`，候选内 min/max 归一化，越短越好；
- `kv_cache`：优先 `vllm:kv_cache_usage_perc`，兼容 `vllm:gpu_cache_usage_perc`，`1-clamp(value,0,1)`；
- `lora_affinity`：请求 `model` 存在于 `vllm:lora_requests_info` active adapters 时为 1，否则为 0；指标缺失时不可用，不作为解析错误；
- `inflight`：gateway-local 当前并发，候选内 min/max 归一化，越少越好；
- `failure`：gateway-local 失败 EWMA，使用 `1-failureEWMA`。

每个 Signal 带可用性、置信度和原因。缺失 Signal 的贡献为 0，不改变总权重分母，因此不会因为缺少指标获得额外优势。若所有 endpoint 的外部指标都缺失，本地 inflight/failure 仍可参与；若所有得分相同，Picker 在最高分集合中均匀随机选择。

## 配置

第一版只暴露少量经过定义的参数：

```yaml
profile: balanced
weights:
  queue: 2
  kvCache: 2
  loraAffinity: 1
  inflight: 1
  failure: 1
feedback:
  ewmaAlpha: 0.2
picker:
  mode: max-score
debug:
  sampleRate: 0
```

约束：

- `profile` MVP 只接受 `balanced`；未知值配置失败，避免静默采用错误策略。
- 权重必须为非负有限数，且至少一个大于 0；未设置时使用上述默认值。
- `ewmaAlpha` 必须在 `(0,1]`。
- `sampleRate` 必须在 `[0,1]`；默认不输出候选明细。
- 不暴露 prompt、token、endpoint 地址作为 metric label。

## 请求生命周期

1. Request Headers：如存在 body，返回 StopIteration 以等待完整 OpenAI 请求；无 body 时 fail-open。
2. Request Body：解析 `model`，读取 upstream hosts 并构建 snapshot。
3. Filter：排除非 Healthy endpoint；metrics 缺失不是硬过滤条件。
4. Normalize/Score：各 scorer 生成 `[0,1]` 分数，按固定总权重累加。
5. Pick：选择最高分，随机打散同分候选；无健康候选或 hostcall/解析异常时返回 Continue，让 Envoy 默认 LB 接管。
6. 成功 override 后：记录所选 endpoint、开始时间，inflight +1。
7. Response Headers/Streaming Body：记录状态码和首次响应数据时间；不得修改响应数据。
8. Stream Done：inflight -1，更新 TTFT、总时延和失败 EWMA；确保取消/异常路径也释放 inflight。

同一 stream 只能完成一次 feedback，避免 response end 与 stream done 重复扣减。

### Filter

MVP 只有健康 Filter。Filter 返回保留集合和低基数原因计数，不使用 queue/KV 等软信号做硬剔除，避免旧 `endpoint_metrics` 的固定级联导致某一信号完全压制其他信号。

### Normalize 与 Score

每个 scorer 输出 `[0,1]`，最终得分为：

```text
score(endpoint) = Σ(weight(signal) × normalized(signal) × confidence(signal))
                  / Σ(configured weights)
```

统一分母使缺失信号成为保守惩罚。所有候选的某项 min/max 相同时，可用信号统一返回 1，表示该信号不区分候选且不引入偏差。

### Pick

MVP `max-score` 与 llm-d 一致：选最高分 endpoint，同分随机。随机源由 Picker 注入，单元测试使用固定 seed；生产使用 WASM 可用随机源。

### Feedback

FeedbackStore 为配置实例共享的有界 map，只为当前 upstream hosts 保留状态；每次请求清理已不在候选集合且 inflight 为 0 的旧 endpoint，防止地址变化造成无限增长。EWMA 仅在有完整观测时更新。

## 指标和日志

使用 proxy-wasm custom metrics，固定指标名、不带 endpoint label：

- `ai_endpoint_picker_decisions_total`
- `ai_endpoint_picker_fallback_total`
- `ai_endpoint_picker_missing_signal_total`
- `ai_endpoint_picker_feedback_total`
- `ai_endpoint_picker_inflight`

由于 proxy-wasm metrics 不适合动态高基数 label，MVP 不为 scorer/profile/endpoint 动态创建 metric。调试日志仅在 sample 命中时输出候选数量、Signal 可用性、分数和选择原因；不得输出 body/prompt/token。

## 兼容与降级

- 新插件完全独立，不修改 `ai-load-balancer` 的解析和行为。
- 配置错误在 plugin config 阶段失败；运行时 hostcall、body、metrics 或评分错误一律 fail-open，让 Envoy 默认 LB 继续处理。
- 单个 metric 缺失只影响对应 Signal；vLLM 没有启用 LoRA 时不会使整个 endpoint 失效。
- 支持新 `vllm:kv_cache_usage_perc`，同时兼容旧 `vllm:gpu_cache_usage_perc`。
- `SetUpstreamOverrideHost` 失败时不建立 feedback lease，避免虚假的 inflight。

## 与 GIE 的后续绑定

当 Higress 支持 GIE `InferencePool` 且 `endpointPickerRef` 为空时，控制面可以默认生成/绑定 `ai-endpoint-picker` 配置，并把 pool endpoints 的主动 metrics 写入现有 upstream host metadata。这样可在不部署 EPP/ext_proc 的情况下获得与 llm-d EPP 核心调度层近似的 Filter/Scorer/Picker 能力。

边界必须明确：这与“完整 llm-d Router/EPP 对等”不同。llm-d 还包括 request parsing、flow control、admission、优先级、多 profile handler、精确 cache index 等能力。Higress 只有在后续分别补齐 #4297、GIE 控制面绑定和必要 flow-control 后，才能扩大“对等”范围。

## 测试计划

- 配置测试：默认值、非法 profile/权重/EWMA/sample rate。
- parser 测试：新旧 KV metric、queue、LoRA 可选、损坏 Prometheus 文本、部分缺失。
- pipeline 测试：高 KV 空闲但长队列、短队列但 KV 高占用、LoRA affinity、缺失 Signal、全部同分、无健康候选。
- feedback 测试：成功、5xx、取消、重复完成、endpoint 清理、EWMA。
- 回归：`go test ./...`、`go vet ./...`、WASM build。

## 实施与发布

- 插件版本从 `1.0.0-alpha` 开始，使 Higress 批量插件构建能在 CI 中编译。
- 首个 PR 只新增插件目录和必要的插件索引/文档入口，不改现有策略默认值。
- 后续 GIE 默认绑定、#4297 Signal Provider、更多 Profile 分别提交独立 Design/PR。

---

- Proposal Issue: 4296

## Design Summary

This design is implemented in #4296 by adding the independent WASM Go plug-in `ai-endpoint-picker`, instead of continuing to extend the `ai-load-balancer` that already takes on cluster/endpoint, Redis, indicator analysis and various historical strategies.

The plug-in is only responsible for endpoint selection within a single upstream cluster, using the same core layering as llm-d Endpoint Picker: **Filter → Normalize/Score → Pick → Feedback**. The first version provides `balanced` Profile, which combines vLLM queue depth, KV cache utilization, LoRA affinity, gateway-local inflight and failure feedback; the real prefix cache score of #4297 will be added later through the stable Signal Provider interface without changing the main scheduling process.

The plug-in name is `ai-endpoint-picker` because it is directly aligned with GIE's `endpointPickerRef`/EPP terminology; `Router` represents the whole of Proxy + EPP in llm-d, and Higress already has `model-router`, and using `router` is prone to boundary ambiguity.

## Reference and Differences

Design reference llm-d Router commit `aa0e31371891f283ceb56cfe3dbddbfdb62fb648`:

- `SchedulerProfile` executes filters, weighted scorers, and pickers sequentially;
- scorer outputs `[0,1]`, accumulated according to weight;
- The queue scorer is normalized using the candidate set min/max. The shorter the queue, the higher the score;
- KV utilization scorer uses `1-utilization`;
- max-score picker randomly disperses candidates with the same score.

Reference source code:

- https://github.com/llm-d/llm-d-router/blob/aa0e31371891f283ceb56cfe3dbddbfdb62fb648/pkg/epp/scheduling/scheduler_profile.go
- https://github.com/llm-d/llm-d-router/tree/aa0e31371891f283ceb56cfe3dbddbfdb62fb648/pkg/epp/framework/plugins/scheduling

The Higress version does not replicate EPP's Kubernetes controller, ext_proc server, flow control, and dynamic plugin loading, but instead completes the endpoint snapshot, scoring, and `SetUpstreamOverrideHost` within the WASM request path, thus maintaining zero additional deployment.

### This implementation

- Added `plugins/wasm-go/extensions/ai-endpoint-picker/` as an independent Go module and plug-in image.
- Get candidate endpoints, health status, and actively captured metrics snapshots from `proxywasm.GetUpstreamHosts()`.
- Parse current and legacy vLLM metric names and allow missing LoRA metrics.
- Implement testable Filter, Signal, Normalize, Score, and Pick data models.
- Implement `balanced` Profile and fail-open fallback.
- Maintain gateway-local endpoint feedback: inflight, TTFT, total latency and failure EWMA.
- Provide low-cardinality custom indicators, sampling debug logs, Chinese and English README, configuration examples and unit tests.

### Not implemented this time

- vLLM ZMQ KVEvents with exact prefix block indexing (#4297).
- Independent queues, priority, fairness, admission control or full llm-d flow control.
- Inter-cluster/inter-provider selection; still the responsibility of existing cluster-level capabilities.
- The GIE controller automatically generates the WasmPlugin binding; this time it only stabilizes the plug-in configuration and semantics to provide targets for subsequent control plane binding.
- Dynamically load any user scorer or learn weights online.

## Directory and components

```text
plugins/wasm-go/extensions/ai-endpoint-picker/
├── main.go # WASM life cycle, request/response callback, override host
├── config.go # User configuration, default profile, verification
├── metrics.go # Low-cardinality proxy-wasm metrics
├── scheduling/
│ ├── types.go # EndpointSnapshot、SignalValue、Decision、Feedback
│ ├── parser.go # Prometheus/vLLM metrics analysis
│ ├── pipeline.go # Filter → Normalize → Score → Pick
│ ├── feedback.go # gateway-local endpoint status and EWMA
│ └── *_test.go
├── go.mod / go.sum
├── VERSION
├── README.md
└── README_EN.md
```

`scheduling` does not rely on proxy-wasm hostcall and uses pure Go input and output, which facilitates unit testing and future reuse by native gateway filters. Only `main.go` is responsible for SDK interaction.

## Core Data Model

```go
type EndpointSnapshot struct {
    Address string
    Healthy bool
    Signals map[SignalName]SignalValue
}

type SignalValue struct {
    Value float64
    Available bool
    Confidence float64
    Source string
    Reason string
}

type Decision struct {
    Address string
    Score float64
    CandidateCount int
    FallbackReason string
}
```

The MVP Signal collection is fixed to:

- `queue`: `vllm:num_requests_waiting`, min/max normalization within candidates, the shorter the better;
- `kv_cache`: Prioritize `vllm:kv_cache_usage_perc`, compatible with `vllm:gpu_cache_usage_perc`, `1-clamp(value,0,1)`;
- `lora_affinity`: 1 when the request `model` exists in `vllm:lora_requests_info` active adapters, otherwise 0; unavailable when the indicator is missing, it is not treated as a parsing error;
- `inflight`: gateway-local current concurrency, min/max normalization within candidates, the less the better;
- `failure`: gateway-local fails EWMA, use `1-failureEWMA`.

Each signal comes with availability, confidence, and reason. The missing Signal has a contribution of 0 and does not change the total weight denominator, so there is no additional advantage gained from the missing indicator. If the external indicators of all endpoints are missing, local inflight/failure can still participate; if all scores are the same, the Picker randomly selects evenly from the highest score set.

## Configuration

The first version exposed only a few defined parameters:

```yaml
profile: balanced
weights:
  queue: 2
  kvCache: 2
  loraAffinity: 1
  inflight: 1
  failure: 1
feedback:
  ewmaAlpha: 0.2
picker:
  mode: max-score
debug:
  sampleRate: 0
```

Constraints:

- `profile` MVP only accepts `balanced`; unknown value configuration fails to avoid silently adopting wrong strategies.
- The weights must be non-negative finite numbers, and at least one is greater than 0; if not set, the above default value is used.
- `ewmaAlpha` must be in `(0,1]`.
- `sampleRate` must be in `[0,1]`; candidate details are not output by default.
- Do not expose prompt, token, and endpoint addresses as metric labels.

## Request life cycle

1. Request Headers: If there is a body, return StopIteration to wait for a complete OpenAI request; fail-open when there is no body.
2. Request Body: parse `model`, read upstream hosts and build snapshot.
3. Filter: exclude non-Healthy endpoints; missing metrics is not a hard filter condition.
4. Normalize/Score: Each scorer generates `[0,1]` scores, which are accumulated according to a fixed total weight.
5. Pick: Select the highest score and randomly scatter candidates with the same score; when there are no healthy candidates or hostcall/parsing exceptions, return Continue and let Envoy's default LB take over.
6. After successful override: record the selected endpoint, start time, and inflight +1.
7. Response Headers/Streaming Body: Record the status code and first response data time; the response data must not be modified.
8. Stream Done: inflight -1, updates TTFT, total latency and failure EWMA; ensure cancellation/exception paths also release inflight.

The same stream can only complete feedback once to avoid repeated deductions for response end and stream done.

### Filter

MVP only has a health filter. Filter returns the reserved set and low cardinality reason count, and does not use soft signals such as queue/KV for hard elimination, to avoid the fixed cascade of the old `endpoint_metrics` causing one signal to completely suppress other signals.

### Normalize and Score

Each scorer outputs `[0,1]`, and the final score is:

```text
score(endpoint) = Σ(weight(signal) × normalized(signal) × confidence(signal))
                  /Σ(configured weights)
```

Unifying the denominator makes missing signals a conservative penalty. When a certain min/max of all candidates is the same, the available signal uniformly returns 1, indicating that the signal does not distinguish candidates and does not introduce bias.

### Pick

MVP `max-score` is consistent with llm-d: select the highest-scoring endpoint, and the same score will be randomized. The random source is injected by Picker, and unit testing uses a fixed seed; production uses the WASM available random source.

### Feedback

FeedbackStore is a bounded map shared by configuration instances and only retains status for the current upstream hosts; every request cleans up old endpoints that are no longer in the candidate set and have an inflight of 0 to prevent unlimited growth caused by address changes. EWMA is only updated when complete observations are available.

## Metrics and logs

Use proxy-wasm custom metrics with fixed metric name and no endpoint label:

- `ai_endpoint_picker_decisions_total`
- `ai_endpoint_picker_fallback_total`
- `ai_endpoint_picker_missing_signal_total`
- `ai_endpoint_picker_feedback_total`
- `ai_endpoint_picker_inflight`

Since proxy-wasm metrics are not suitable for dynamic high-cardinality labels, MVP does not dynamically create metrics for scorer/profile/endpoint. The debug log only outputs the number of candidates, signal availability, score, and selection reason when the sample hits; body/prompt/token must not be output.

## Compatibility and downgrade

- The new plugin is completely independent and does not modify the parsing and behavior of `ai-load-balancer`.
- Configuration errors fail in the plugin config phase; hostcall, body, metrics or scoring errors at runtime will all fail-open, allowing Envoy's default LB to continue processing.
- The loss of a single metric only affects the corresponding Signal; vLLM will not invalidate the entire endpoint when LoRA is not enabled.
- Support new `vllm:kv_cache_usage_perc` and compatible with old `vllm:gpu_cache_usage_perc`.
- Do not establish feedback lease when `SetUpstreamOverrideHost` fails to avoid spurious inflights.

## Subsequent binding to GIE

When Higress supports GIE `InferencePool` and `endpointPickerRef` is empty, the control plane can generate/bind the `ai-endpoint-picker` configuration by default and write active metrics of pool endpoints into the existing upstream host metadata. This can obtain Filter/Scorer/Picker capabilities similar to those of the llm-d EPP core scheduling layer without deploying EPP/ext_proc.

The boundaries must be clear: this is not the same as "full llm-d Router/EPP peering". llm-d also includes capabilities such as request parsing, flow control, admission, priority, multiple profile handlers, and precise cache index. Higress can only expand the "peer-to-peer" scope after completing #4297, GIE control plane binding and necessary flow-control respectively.

## Test plan

- Configuration test: default value, illegal profile/weight/EWMA/sample rate.
- Parser test: old and new KV metric, queue, LoRA optional, corrupted Prometheus text, partially missing.
- Pipeline test: high KV idle but long queue, short queue but high KV occupancy, LoRA affinity, missing Signal, all equal scores, no healthy candidates.
- feedback testing: success, 5xx, cancellation, repeated completion, endpoint cleanup, EWMA.
- Regression: `go test ./...`, `go vet ./...`, WASM build.

## Implementation and Release

- Plugin version starting from `1.0.0-alpha` enables Higress batch plugin builds to compile in CI.
- The first PR only adds the plug-in directory and necessary plug-in index/document entries, and does not change the existing policy defaults.
- Subsequent GIE default binding, #4297 Signal Provider, and more Profiles will be submitted to independent Design/PR respectively.

### LINKED ISSUE #4347: Implement: 新增 ai-endpoint-picker 插件 || Implement: Added ai-endpoint-picker plug-in
- Design Issue: 4346

## 实现目标

在独立分支 `codex/ai-endpoint-picker` 上新增 `plugins/wasm-go/extensions/ai-endpoint-picker/`，实现 #4346 定义的 endpoint-only 智能调度 MVP。现有 `ai-load-balancer` 不做任何修改。

## 设计不变量

1. 请求决策严格遵循 Filter → Normalize/Score → Pick，Feedback 只消费实际 override 成功的请求生命周期。
2. `scheduling` 子包为纯 Go，不调用 proxy-wasm；host ABI、body/stream 回调和 custom metrics 只存在于 plugin main/config 边界。
3. queue、KV utilization、LoRA、local inflight 和 failure 的分数统一为 `[0,1]`，使用固定配置总权重；缺失信号不能获得额外优势。
4. 运行时 body、host metadata、metrics、调度或 override 异常全部 fail-open 到 Envoy LB。
5. 不记录 prompt/token，不创建 endpoint 高基数 metric。
6. 所有代码、测试、文档和模块文件都限定在新插件目录。

## 实现范围

- plugin 生命周期和配置校验；
- upstream host snapshot 与 vLLM Prometheus metrics parser；
- health filter、balanced weighted scorers、max-score/random-tie picker；
- inflight、TTFT、总时延、failure EWMA feedback；
- proxy-wasm 低基数 counters/gauge 与采样 debug；
- 配置/parser/pipeline/feedback 单元测试；
- `VERSION=1.0.0-alpha`、README/README_EN、go.mod/go.sum；
- Go test、vet、WASM compile。

## 明确不包含

- #4297 ZMQ KVEvents；
- GIE 控制面自动绑定；
- cluster 级/provider 级调度；
- 独立 EPP/ext_proc、flow control 或 admission queue；
- 对现有 `ai-load-balancer` 的重构或迁移。

## 工作空间与交付

- Integration branch: `codex/ai-endpoint-picker`
- Base revision: `f21b9fa50cbfc57ded4b8db31b31d169f2426b3b`
- Task: `TASK-4346001`
- 变更由单一实现角色在 issue-spec managed worktree 完成一个 DCO commit。
- 合入集成分支后，由不同角色在精确 revision 上完成独立 review；修复回到原实现角色。

## 最低验证

```bash
cd plugins/wasm-go/extensions/ai-endpoint-picker
go test ./...
go vet ./...
GOOS=wasip1 GOARCH=wasm go build -buildmode=c-shared -o /tmp/ai-endpoint-picker.wasm ./
```

额外检查：

- `git diff --name-only <base>...HEAD` 只包含 `plugins/wasm-go/extensions/ai-endpoint-picker/`；
- 当前/旧 KV metric、LoRA 缺失、metrics 损坏、同分、无健康 endpoint、override 失败和 feedback 重复完成均有测试或明确的 host-boundary 证明；
- `ai-load-balancer` 目录无变化。

## 完成条件

- TASK-4346001 的 checklist 全部有实现或测试证据；
- SPEC-4296001 至 SPEC-4296004 均有 PROCESS 与独立 REVIEW 覆盖；
- PR 关联 Proposal #4296、Design #4346 和本 Implement issue；
- CI/本地验证没有已知阻塞项。

---

- Design Issue: 4346

## achieve goals

Added `plugins/wasm-go/extensions/ai-endpoint-picker/` on the independent branch `codex/ai-endpoint-picker` to implement the endpoint-only intelligent scheduling MVP defined in #4346. No modifications will be made to the existing `ai-load-balancer`.

## Design invariants

1. The request decision strictly follows Filter → Normalize/Score → Pick, and Feedback only consumes the request life cycle in which the actual override is successful.
2. The `scheduling` sub-package is pure Go and does not call proxy-wasm; host ABI, body/stream callbacks and custom metrics only exist at the plugin main/config boundary.
3. The scores of queue, KV utilization, LoRA, local inflight and failure are unified to `[0,1]`, using a fixed configuration total weight; no additional advantages can be obtained for missing signals.
4. All runtime body, host metadata, metrics, scheduling or override exceptions fail-open to Envoy LB.
5. Do not record prompt/token, and do not create endpoint high-cardinality metric.
6. All code, tests, documentation and module files are restricted to the new plugin directory.

## Implementation scope

- plugin life cycle and configuration verification;
- upstream host snapshot and vLLM Prometheus metrics parser;
- health filter, balanced weighted scorers, max-score/random-tie picker;
- inflight, TTFT, total delay, failure EWMA feedback;
- proxy-wasm low-cardinality counters/gauge and sampling debug;
- Configure /parser/pipeline/feedback unit test;
- `VERSION=1.0.0-alpha`, README/README_EN, go.mod/go.sum;
- Go test,vet,WASM compile.

## Explicitly not included

- #4297 ZMQ KVEvents;
- GIE control plane is automatically bound;
- cluster level/provider level scheduling;
- Standalone EPP/ext_proc, flow control or admission queue;
- Refactoring or migration of existing `ai-load-balancer`.

## Workspace and Delivery

- Integration branch: `codex/ai-endpoint-picker`
- Base revision: `f21b9fa50cbfc57ded4b8db31b31d169f2426b3b`
- Task: `TASK-4346001`
- Changes made by a single implementation role complete a DCO commit in the issue-spec managed worktree.
- After merging into the integration branch, different roles will complete independent reviews on the exact revision; the repair will return to the original implementation role.

## Minimum verification

```bash
cd plugins/wasm-go/extensions/ai-endpoint-picker
go test ./...
go vet ./…
GOOS=wasip1 GOARCH=wasm go build -buildmode=c-shared -o /tmp/ai-endpoint-picker.wasm ./
```

Additional checks:

- `git diff --name-only <base>...HEAD` only contains `plugins/wasm-go/extensions/ai-endpoint-picker/`;
- Current/old KV metrics, missing LoRA, corrupted metrics, same score, no healthy endpoint, override failure and feedback duplicate completion all have testing or clear host-boundary proof;
- No changes to the `ai-load-balancer` directory.

## Completion conditions

- All checklists of TASK-4346001 have implementation or testing evidence;
- SPEC-4296001 to SPEC-4296004 are covered by PROCESS and independent REVIEW;
- PR links Proposal #4296, Design #4346 and this Implement issue;
- CI/local validation has no known blockers.

EndlessSeeker:
## Why

`ai-load-balancer` already combines cluster selection, endpoint selection, Redis state, metrics parsing, and several historical policies. This PR adds a dedicated endpoint picker whose boundary matches the Gateway API Inference Extension (GIE) `endpointPickerRef` / EPP terminology and the scheduling core of llm-d EPP.

## What

- Add the standalone WASM Go plugin `ai-endpoint-picker` for endpoint selection inside one upstream cluster.
- Implement `Filter -> Normalize -> Score -> Pick -> Feedback` with llm-d Router `aa0e31371891f283ceb56cfe3dbddbfdb62fb648` queue-depth 2, KV-cache 2, and approximate-prefix 3 weights, plus gateway-local inflight weight 1.
- Enable gateway-local inflight scoring by default as a low-weight correction between `/metrics` updates. Keep LoRA affinity and failure EWMA opt-in with zero default weight. Missing signals contribute zero under the fixed configured-weight denominator.
- Prefer `vllm:kv_cache_usage_perc`, accept legacy `vllm:gpu_cache_usage_perc`, and use `vllm:cache_config_info` to size the weighted prefix LRU.
- Approximate prefix locality from semantic Chat messages/tools or Completion prompts. The index is gateway-local routed-request history, not vLLM's exact KV state; exact KVEvents remain tracked by #4297.
- Randomize equal maximum scores, record prefix history only after a successful upstream override, and settle feedback exactly once.
- Add `prefix.toolMode=none|identity|full` while keeping scheduling weights unchanged. The default `identity` mode hashes only ordered tool type/name identities; `none` skips tools and `full` preserves bounded complete canonicalization.

### Review hardening in `7f2b3e445d9c3df319b20b4e9fad4e5a4814baf7`

- Preflight request structure with a fixed-stack, non-recursive scanner before any recursive JSON API. At or below 64 levels, perform full JSON validation and semantic extraction; over-depth valid envelopes keep model/queue/KV scheduling but disable only approximate prefix locality.
- Treat valid batched Completion token prompts as prefix-unsupported, so other scorers continue. Flat token IDs remain supported; invalid or mixed shapes remain explicit extraction errors.
- Skip only a host whose metadata, health state, or Prometheus snapshot is malformed instead of failing open the whole cluster decision.
- Record prefix chains suffix-first so weighted-LRU pressure preserves a usable head prefix; delete unhealthy-host prefix state and distinguish that from healthy hosts without `num_gpu_blocks`, which retain default capacity.
- Emit fixed low-cardinality selection reasons (`max_score`, `random_tie`) and signal/skip bitmasks and counts. Logs do not contain endpoint addresses, request bodies, prompts, tokens, or dynamic error strings.
- Add adversarial depth, prompt-shape, bad-host isolation, eviction, cleanup, and observability-contract regression tests; update both READMEs.

### Configurable tool-prefix precision in `f66badc2899850c0a16200f1a0e32c4b9d84300b`

- `identity` is the zero-configuration default and reads only up to 64 ordered `tool.type` / `tool.function.name` identities within an 8192-byte identity budget. It does not canonicalize descriptions, parameters, or complete schemas.
- `none` minimizes gateway work by ignoring top-level tools; `full` retains the existing bounded canonical tools behavior for dynamically changing schemas.
- Reaching the identity cap preserves the prefix already produced and continues with messages. All three modes affect scheduling hints only, not model-output correctness.
- Unit coverage includes mode semantics, invalid configuration, caps, field-order stability, and allocation bounds against very large schemas.

### Bounded request work and compact metrics snapshots in `eb4af7264a357180a9358a0f2f94cbac671981b3`

- Add `prefix.maxBlocks` with a request-wide default of 32 and a validated 1..128 range.
- Add `prefix.blockSizeTokens` with an unchanged default of 1024 and a validated 1..1024 range. It controls the approximate scheduler hash-block granularity for both semantic text and flat token-ID prompts; it does not change vLLM's actual KV-cache block size. GIE built-in routes continue to inherit the plugin default unless users provide plugin configuration. Tools, messages, Completion string arrays, and flat token IDs share the same budget; semantic decoding, canonicalization, and array visitation stop as soon as it is exhausted.
- Keep the full linear JSON validity/depth preflight for safety, while bounding prompt-bearing semantic work separately. Message metadata and structured content now share the existing canonical-node budget, so very wide objects disable only prefix locality instead of creating unbounded field materialization or sorting work.
- Cache compact per-endpoint scheduling snapshots for 250ms. A TTL hit avoids another `GetUpstreamHosts()` call; an unchanged metrics fingerprint reuses the model-neutral parsed result, while LoRA affinity is derived per request without mutating cached maps.
- When metrics change, retain only the five vLLM families used by queue, current/legacy KV, LoRA, and cache configuration before invoking the Prometheus parser. The relevant subset is capped at 64KiB and starts with a small allocation, so large unrelated families neither build DTOs nor force a 64KiB temporary buffer.
- Expired snapshots are never served after refresh failure. The plugin cache retains neither raw host metadata nor raw metrics bodies.
- Envoy still retains and exposes the complete health-check response when `store_metrics: true`; removing that native storage/hostcall copy cost requires a separate Envoy/Istio source-side filtering change and is not claimed by this plugin-only update.
- GIE built-in integration can continue emitting an empty plugin configuration and inherits these defaults without duplicating them in the control plane.

## Memory and buffering bounds

- Buffer only requests with a positive trustworthy `Content-Length` up to `limits.maxRequestBodyBytes` (4 MiB by default, configurable up to 100 MiB). Oversized and chunked/unknown-length requests skip the optional picker and continue unchanged.
- Cap the gateway-local prefix LRU with `prefix.maxCacheBlocksPerEndpoint` (31,250 by default). Missing or oversized `num_gpu_blocks` values cannot expand it past that configured cap.
- Add `limits.vmRebuildThresholdBytes` (200 MiB by default, 0 disables, up to 4 GiB). It requests VM rebuild at the threshold, matching ai-proxy's policy without upgrading wasm-go dependencies. This is explicitly documented as a soft recovery threshold, not a hard memory quota.
- Fingerprint only the compact five-family metrics subset. Unrelated `/metrics` churn no longer triggers another Prometheus parse; the 250ms snapshot TTL and 64KiB relevant-subset cap remain.
- No `go.mod` or `go.sum` change was required.

## GIE compatibility

GIE v1.4 ExternalEPP support from #4318 remains unchanged. A normal core `Service` `endpointPickerRef` still uses external EPP/ext_proc. The built-in route path is selected only by this well-known reference:

```yaml
endpointPickerRef:
  group: extensions.higress.io
  kind: WasmPlugin
  name: ai-endpoint-picker
```

The corresponding control-plane work is #4608 plus higress-group/istio#69. It generates route-scoped plugin rules and inherits this plugin's defaults. In BuiltIn GIE mode, the controller configures the final inference cluster with exactly one `/metrics` health check and `store_metrics: true`, replacing other health checks on that cluster. Standalone WasmPlugin deployments still configure these upstream metrics capabilities explicitly.

## Validation

Current implementation head: `02365362a548b3bf7ed89dbc22be8e53a2411ee2`

- `go test -count=1 ./...` — PASS
- `go test -race -count=1 ./...` — PASS
- `go vet ./...` — PASS
- `go mod tidy -diff` — clean
- `GOOS=wasip1 GOARCH=wasm go build -buildmode=c-shared` — PASS
- `git diff --check` — PASS
- Added regression coverage for request-body buffering/fail-open behavior, VM threshold decoding, metrics fingerprint reuse, cache-cap clamping, invalid limit configuration, default/custom block sizing, text segmentation, flat token-ID segmentation, and request-wide block budgets.
- Independent exact-diff review from `ab18cf14d421788f0df019cb7db04055ed94d116` to `02365362a548b3bf7ed89dbc22be8e53a2411ee2` found no actionable P0/P1/P2 issues.

The earlier real proxy-Wasm adversarial verification remains applicable to the prefix parser and scheduler behavior. Pinned real proxy-Wasm baseline/fixed verification used `higress/gateway:v2.2.3@sha256:cf1fa0e100e79890b1de15c74f41a21fabba8f854b9d004d8f5b8ccb48a95c5d` and the same fixtures for three repetitions each:

- Baseline `1e7502978a100127ec3f2949e653bf2f02dc8d94`, Wasm SHA-256 `8b701395a114b2694dd528e5cc07f58b27533badbbd32d44bb5aab7af2f78903`: batched tokens triggered full plugin fail-open 3/3; depth-20,000 triggered V8 `Maximum call stack size exceeded` and HTTP 503 3/3.
- Fixed `7f2b3e445d9c3df319b20b4e9fad4e5a4814baf7`, Wasm SHA-256 `1162d877501ec20e2d886a8f9d14ccd93b47140241ce6926bd7dfaaf4f57cfec`: flat tokens, batched tokens, and depth-20,000 all returned HTTP 200 (9/9); no stack overflow or plugin fail-open; all 9 requests produced a normal scheduling decision.
- Containers, Compose network, volumes, and localhost ports 10000/9901 were removed after capture.

GitHub CI at the pre-rebase equivalent head completed with 23 successful checks, 1 intentional skip, 0 failures, and 0 pending checks, including changed-plugin unit tests, GO/RUST Wasm e2e, Gateway API v1.4/v1.6, Inference Extension v1.4, Higress conformance, CodeQL, DCO, CLA, license, validation, build, lint, coverage, and publish.

## Design references

- Plugin Proposal: #4296
- Plugin Design and review-hardening amendment: #4346
- Plugin implementation tracking: #4347, TASK-4346004, TASK-4346005, TASK-4346006, and TASK-4346007 in #4346
- GIE built-in integration Proposal/Design: #4289 / #4290
- GIE v1.4 ExternalEPP implementation: #4318
- Exact vLLM KVEvents follow-up: #4297
- llm-d Router reference: https://github.com/llm-d/llm-d-router/tree/aa0e31371891f283ceb56cfe3dbddbfdb62fb648

## Agent-assisted contribution declaration

This PR materially used a coding agent for design analysis, implementation, tests, review follow-up, runtime verification, and PR documentation. The agent was constrained to the issue-spec Design/TASK scope, preserved fail-open and public-configuration compatibility, and stopped before merge.

This in-flight PR uses the repository's verified administrator exception rather than claiming retroactive maintainer approval:

- PR: #4352
- authenticated GitHub login: `EndlessSeeker`
- canonical-repository `role_name`: `admin`
- actual PR author: `EndlessSeeker`
- `GH_TOKEN` and `GITHUB_TOKEN` were unset for every verified-exception identity, permission, and PR-author command
- rationale: the existing Proposal/Design have no explicit independent maintainer approval record; the authenticated administrator and PR author is continuing an already-open review while retaining issue-spec Design, TASK, test, and independent-review traceability

The accepting or merging maintainer must independently revalidate the current PR author and canonical-repository role with both token override variables unset. This declaration is review context, not approval or merge authority.

Issue-spec managed closing links:
Closes #4296
Closes #4346
Closes #4347

Latest exact-head note: the branch was rebased without semantic changes onto canonical `main` at `ce209a2c4c3ad95e80084c1028617f329ca4e32c`; CI completed against `eb4af7264a357180a9358a0f2f94cbac671981b3`: 23 successful checks, 1 intentional skip, 0 failures, and 0 pending checks.

Previous exact-head CI at `78675fa14c39883f26babdce3cf33b0cd3b174d0` completed successfully. CI for exact head `02365362a548b3bf7ed89dbc22be8e53a2411ee2` completed successfully: all required checks passed, including changed-plugin tests, GO/RUST Wasm e2e, Gateway API v1.4/v1.6, GIE v1.4, Higress conformance, build, lint, coverage, CodeQL, DCO, CLA, license, validation, and publish.

Runtime follow-up on the preserved ACK GPU environment used the unchanged 1024-token block default together with inflight weight 1 and the control-plane 2-second `/metrics` interval. Three hot-state runs completed 900/900 requests with zero picker fallback, 148.63 ms mean TTFT, 5971.65 ms mean RT, 10621.13 ms mean P99 RT, 422.24 token/s, 39.23 ms/token TPOT, and a representative 74.37% prefix-cache hit rate. These are small steady-state gains over the prior BuiltIn baseline; the clearer result is that full-run endpoint distributions stayed near three-way balance (92/111/97, 93/106/101, and 95/106/99).


plugins/wasm-go/extensions/ai-endpoint-picker/main.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	now := time.Now()
+	feedback := &requestFeedback{lease: config.store.Begin(decision.Address, now)}
+	ctx.SetContext(feedbackContextKey, feedback)
+	config.metrics.missing(missingSignals)
+	config.metrics.decision()
+	config.metrics.beginFeedback()
+	if config.sampleRate > 0 && config.random.Float64() < config.sampleRate {
+		log.Debugf("ai-endpoint-picker candidates=%d score=%.4f missing_signals=%d", decision.CandidateCount, decision.Score, missingSignals)
```
EndlessSeeker: P2（非阻断）：sampled debug 日志目前只有候选数、最终分数和聚合 `missing_signals`，没有指出哪些固定 signal 可用，也没有 winner/tie/fallback 等低基数选择原因；`SignalValue.Reason` 同样未被生产或消费。

触发条件是 `debug.sampleRate > 0` 且命中采样。它不影响调度正确性，但不易按 Design #4346 的 observability 约定诊断某项 scorer 缺失或选择原因。建议后续记录固定 signal 的可用计数/位图和固定选择原因枚举，同时继续避免 endpoint、body、prompt、token 或动态错误文本。

johnlanni (changes requested):
## Review summary

Reviewed exact head `1e7502978a100127ec3f2949e653bf2f02dc8d94` independently. The overall plugin structure, fixed-denominator scoring, fail-open lifecycle, override-success-only recording, race safety, and build/test hygiene are strong. Two issues should be fixed before merge.

### P1: deeply nested valid JSON causes quadratic CPU work

`prefixcache/locality.go` recursively canonicalizes arrays via `writeCanonicalJSON` → `forEachArrayValue` → `scanJSONValue`. At every nesting level, `scanJSONValue` rescans the full remaining subtree and allocates a delimiter stack proportional to depth. The pseudo-token/byte caps bound output, but not this scan work.

A focused PoC using a valid extra metadata field in a Chat message produced:

```text
depth=  5,000  ~10 KB   178 ms
depth= 10,000  ~20 KB   676 ms
depth= 20,000  ~40 KB   3.08 s
depth= 40,000  ~80 KB   14.14 s
```

A roughly 1 MB / 500,000-depth body remained CPU-bound for more than 4.5 minutes before termination. This is approximately quadratic scaling. Since the WASM request callback runs on an Envoy worker thread, a small crafted request can stall traffic on that worker.

Please add a strict JSON nesting-depth bound in `scanJSONValue` / canonical traversal, map over-complex input to prefix-unavailable or fail-open, and add an adversarial regression test that demonstrates bounded runtime.

### P1: valid batched token-ID Completion prompts bypass the whole scheduler

OpenAI/vLLM Completion requests may use batched token prompts such as:

```json
{"model":"m","prompt":[[1,2],[3,4]]}
```

In `extractCompletion`, the first nested array selects number mode and is passed to `parseUint32`, producing `prompt token ID is invalid`. `main.go` treats any extraction error as full fail-open, so queue/KV scheduling is skipped entirely. This differs from unsupported multimodal Chat input, where only prefix locality is disabled and queue/KV scheduling continues.

Please either:

- support one prefix chain per inner token array; or
- classify nested token batches as prefix-unsupported (`supported=false`, no error) so the remaining scorers still run.

Add a regression test for flat tokens, batched token arrays, and mixed/invalid prompt shapes.

### Non-blocking follow-ups

- `prefixcache/index.go` records blocks head-to-tail with `PushFront`, so capacity eviction removes chain heads before suffixes. `Score` stops at the first missing head, leaving suffix entries that consume capacity but can never match. Consider recording blocks in reverse so eviction is tail-first and retained entries always form a usable prefix.
- Invalid metadata or metrics on one host currently fails open the entire cluster decision; skipping only the bad host would preserve scheduling for healthy peers.
- `SetCapacity(address, 0)` for unhealthy hosts maps back to the default capacity rather than clearing their cache.

### Verification completed

```text
go test ./...                         PASS
go test -race ./...                   PASS
go vet ./...                          PASS
gofmt -l .                            clean
go mod tidy -diff                     clean
GOOS=wasip1 GOARCH=wasm go build ...  PASS (valid 12.7 MB WASM)
prefixcache benchmarks                PASS
```

The worktree remained clean and the exact PR head was rechecked. Design/process references: https://github.com/higress-group/higress/issues/4296, https://github.com/higress-group/higress/issues/4346, https://github.com/higress-group/higress/issues/4347.

plugins/wasm-go/extensions/ai-endpoint-picker/main.go
EndlessSeeker: 已在 425d498fc79f3a8ea96aa0807da0047ac2602b47 修复：Decision 现在使用固定低基数 reason（max_score/random_tie），并记录固定 signal availability mask、skipped signal mask/count；采样日志仍不包含 endpoint、body、prompt、token 或动态错误文本。对应契约测试和完整 runtime 证据见 https://github.com/higress-group/higress/pull/4352#issuecomment-5344123332 。

johnlanni (approved):
LGTM

PR #4375, opened 2026-08-08
bugfix: fix unsafe type assertion in ai-proxy retryCall retryCount context

### LINKED ISSUE #4373: Unsafe type assertion in ai-proxy retryCall retryCount context
## Description

The `retryCall` function in `plugins/wasm-go/extensions/ai-proxy/provider/retry.go` uses a bare type assertion on `ctx.GetContext(ctxRetryCount).(int)` that can cause a Wasm plugin panic if the context value is missing or has an unexpected type.

### Affected file

`plugins/wasm-go/extensions/ai-proxy/provider/retry.go` — `retryCall` (line 83):
```go
retryCount := ctx.GetContext(ctxRetryCount).(int)
```

### Root Cause

The `ctxRetryCount` context key is set to `1` (int) in `retryFailedRequest` before calling `sendRetryRequest`, which eventually triggers `retryCall`. However, if the context value is corrupted or missing (e.g. by a plugin interaction or race condition), the bare `.(int)` assertion panics.

This is especially dangerous because `retryCall` is called from an HTTP callout callback — if it panics, the retry mechanism fails silently and the original request hangs.

### Expected behavior

The context assertion should use the comma-ok pattern to gracefully handle missing/corrupted context values, returning an error instead of panicking.

### Actual behavior

The bare assertion panics when the context value is missing or corrupted, crashing the Wasm plugin instance and potentially hanging the original request.

### Version

v2.2.3 (main branch, commit f21b9fa5)

wc4440222:
## I. What does this PR do

Fixes an unsafe type assertion in the `retryCall` function of `plugins/wasm-go/extensions/ai-proxy/provider/retry.go` that can cause a Wasm plugin panic if the context value is missing or has an unexpected type.

### Root Cause

The `retryCall` function uses a bare type assertion on `ctx.GetContext(ctxRetryCount).(int)`. While `ctxRetryCount` is set to `1` (int) in `retryFailedRequest` before calling `sendRetryRequest`, the bare assertion panics if the context value is corrupted or missing (e.g. by a plugin interaction or race condition).

This is especially dangerous because `retryCall` is called from an HTTP callout callback — if it panics, the retry mechanism fails silently and the original request may hang.

### Fix

Replace the bare assertion with the comma-ok pattern:
```go
retryCount, ok := ctx.GetContext(ctxRetryCount).(int)
if !ok {
    log.Errorf("retryCount context value missing or invalid, aborting retry")
    proxywasm.ResumeHttpResponse()
    return
}
```

When the context value is missing or invalid, the function logs the error and resumes the HTTP response (returning the original upstream response to the client) instead of panicking.

## II. Fixes Issue

Fixes #4373

## III. Why no test cases

The retry path requires a full Wasm runtime environment with HTTP callout callbacks. The fix is a straightforward comma-ok pattern replacement that prevents panics in edge cases. Existing provider tests continue to pass.

## IV. How to verify

- `cd plugins/wasm-go/extensions/ai-proxy && GOOS=wasip1 GOARCH=wasm go build . ./provider/` — compiles successfully
- `cd plugins/wasm-go/extensions/ai-proxy && go test ./provider/ -count=1` — all tests pass

- Same defensive pattern as approved in #4226 (WAF plugin) and #4224 (ai-load-balancer)
- When the context value is missing, the function calls `proxywasm.ResumeHttpResponse()` to return the original upstream response to the client, rather than hanging
- The `ok` variable is scoped to the `retryCall` function and does not conflict with any existing declarations

- [x] Regular modification scenario: AI analyzed the retryCall callback flow to identify the unsafe type assertion

### DIFF
--- a/plugins/wasm-go/extensions/ai-proxy/provider/retry.go
+++ b/plugins/wasm-go/extensions/ai-proxy/provider/retry.go
@@ -7,9 +7,9 @@ import (
 	"net/http"
 
 	"github.com/alibaba/higress/plugins/wasm-go/extensions/ai-proxy/util"
+	"github.com/higress-group/proxy-wasm-go-sdk/proxywasm"
 	"github.com/higress-group/wasm-go/pkg/log"
 	"github.com/higress-group/wasm-go/pkg/wrapper"
-	"github.com/higress-group/proxy-wasm-go-sdk/proxywasm"
 	"github.com/tidwall/gjson"
 )
 
@@ -81,9 +81,14 @@ func (c *ProviderConfig) retryCall(
 	ctx wrapper.HttpContext, activeProvider Provider,
 	apiName ApiName, statusCode int, responseHeaders http.Header, responseBody []byte,
 	retryClient *wrapper.ClusterClient[wrapper.RouteCluster],
-	apiTokenInUse string, apiTokens []string) {
-
-	retryCount := ctx.GetContext(ctxRetryCount).(int)
+	apiTokenInUse string, apiTokens []string,
+) {
+	retryCount, ok := ctx.GetContext(ctxRetryCount).(int)
+	if !ok {
+		log.Errorf("retryCount context value missing or invalid, aborting retry")
+		proxywasm.ResumeHttpResponse()
+		return
+	}
 	log.Infof("Sent retry request: %d/%d", retryCount, c.retryOnFailure.maxRetries)
 
 	if statusCode == 200 {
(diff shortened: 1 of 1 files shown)


johnlanni (approved):
Reviewed exact head `42f0c8ec5e2d661506eae6f3d990a22f965eff83`.

`ctxRetryCount` is initialized before every live `retryCall` entry, so this is defensive hardening for an invariant violation. If the guard ever fires, `ResumeHttpResponse()` mirrors the existing terminal retry-failure branches: the response is paused, the function returns immediately, and there is no double-resume or state mutation afterward.

The focused/full ai-proxy tests and Wasm build pass. LGTM.

PR #4377, opened 2026-08-08
bugfix: fix unsafe type assertion in ai-agent toolsCall ToolCallsCount context

### LINKED ISSUE #4376: Unsafe type assertion in ai-agent toolsCall ToolCallsCount context
## Description

The `toolsCall` function in `plugins/wasm-go/extensions/ai-agent/main.go` uses a bare type assertion on `ctx.GetContext(ToolCallsCount).(int)` that can cause a Wasm plugin panic if the context value is missing or has an unexpected type.

### Affected file

`plugins/wasm-go/extensions/ai-agent/main.go` — `toolsCall` (line ~394):
```go
count := ctx.GetContext(ToolCallsCount).(int)
```

### Root Cause

The `ToolCallsCount` context key is set to `0` (int) in `onHttpRequestBody`. However, `toolsCall` is called from HTTP callout callbacks (the LLM Post callback in `toolsCallResult` → `toolsCall`). If the callback executes without `onHttpRequestBody` having been called first (e.g. due to protocol routing or context corruption), the context value is `nil` and the bare `.(int)` assertion panics.

This is especially dangerous because `toolsCall` is called from an HTTP callout callback — if it panics, the agent's tool-calling loop fails silently and the original request hangs.

### Expected behavior

The context assertion should use the comma-ok pattern to gracefully handle missing/corrupted context values, returning a safe default instead of panicking.

### Actual behavior

The bare assertion panics when the context value is missing or corrupted, crashing the Wasm plugin instance.

### Version

wc4440222:
## I. What does this PR do

Fixes an unsafe type assertion in the `toolsCall` function of `plugins/wasm-go/extensions/ai-agent/main.go` that can cause a Wasm plugin panic if the context value is missing or has an unexpected type.

### Root Cause

The `toolsCall` function uses a bare type assertion on `ctx.GetContext(ToolCallsCount).(int)`. While `ToolCallsCount` is set to `0` (int) in `onHttpRequestBody`, `toolsCall` is called from HTTP callout callbacks (the LLM Post callback in `toolsCallResult` → `toolsCall`). If the callback executes without `onHttpRequestBody` having been called first, the context value is `nil` and the bare `.(int)` assertion panics.

This is especially dangerous because `toolsCall` is called from an HTTP callout callback — if it panics, the agent's tool-calling loop fails silently and the original request hangs.

### Fix

Replace the bare assertion with the comma-ok pattern:
```go
count, ok := ctx.GetContext(ToolCallsCount).(int)
if !ok {
    count = 0
}
```

When the context value is missing or invalid, the function uses `0` as the default — the same value set in `onHttpRequestBody` initialization.

## II. Fixes Issue

Fixes #4376

## III. Why no test cases

The `toolsCall` path requires a full Wasm runtime environment with HTTP callout callbacks and LLM integration. The fix is a straightforward comma-ok pattern replacement that prevents panics in edge cases. Existing tests continue to pass.

## IV. How to verify

- `cd plugins/wasm-go/extensions/ai-agent && GOOS=wasip1 GOARCH=wasm go build ./...` — compiles successfully
- `cd plugins/wasm-go/extensions/ai-agent && go test ./... -count=1` — all tests pass

- Same defensive pattern as approved in #4226 (WAF plugin) and #4224 (ai-load-balancer)
- When the context value is missing, the default `0` is used — same as the initialization value in `onHttpRequestBody`
- The `ok` variable is scoped to the `toolsCall` function and does not conflict with any existing declarations
- Note: competitor PR #4311 by yuluo-yx touches the same file (ai-agent/main.go) but addresses a different issue (OpenAPI server config validation), so there should be no merge conflict

- [x] Regular modification scenario: AI analyzed the toolsCall callback flow to identify the unsafe type assertion

### DIFF
--- a/plugins/wasm-go/extensions/ai-agent/main.go
+++ b/plugins/wasm-go/extensions/ai-agent/main.go
@@ -396,7 +396,10 @@ func toolsCall(ctx wrapper.HttpContext, llmClient wrapper.HttpClient, llmInfo LL
 	if action == "Final Answer" {
 		return types.ActionContinue, actionInput
 	}
-	count := ctx.GetContext(ToolCallsCount).(int)
+	count, ok := ctx.GetContext(ToolCallsCount).(int)
+	if !ok {
+		count = 0
+	}
 	count++
 	log.Debugf("toolCallsCount:%d, config.LLMInfo.MaxIterations=%d", count, llmInfo.MaxIterations)
 	// 函数递归调用次数，达到了预设的循环次数，强制结束


johnlanni (approved):
Reviewed exact head `48ccd78043cd80a09d3b1006a9922cacba5befc2`.

The missing `ToolCallsCount` state is reachable when request-body parsing exits before initialization but a non-empty upstream response still enters `toolsCall`. The guard removes the panic and uses the same initial value (`0`). It does not create an unbounded loop: after the first increment the context is initialized, and execution remains capped by `MaxIterations` (default 15).

Focused PoC reproduced the baseline panic and clean pass-through on this head; package tests and Wasm build pass. LGTM.

Non-blocking follow-up: `rawResponse.Choices[0]` immediately above this path may deserve a separate defensive check for choice-less responses.

PR #4379, opened 2026-08-08
bugfix: fix panic in mcp-server FixedQueryToken when credential keys are missing

### LINKED ISSUE #4378: Panic in mcp-server FixedQueryToken when credential keys are missing
## Description

The `FixedQueryToken` function in `plugins/golang-filter/mcp-server/registry/remote.go` uses bare type assertions on map lookups that discard the `ok` flag, causing a panic when the credential map is missing the expected keys.

### Affected file

`plugins/golang-filter/mcp-server/registry/remote.go` — `FixedQueryToken` (line ~47-50):
```go
func FixedQueryToken(cred *CredentialInfo, h *HttpRemoteCallHandle) {
    key, _ := cred.Credentials[FIX_QUERY_TOKEN_KEY]
    value, _ := cred.Credentials[FIX_QUERY_TOKEN_VALUE]
    h.Query[key.(string)] = value.(string)
}
```

### Root Cause

The `ok` flag from the map lookup is discarded (`_ :=`). If `cred.Credentials` does not contain the `"key"` or `"value"` entries (e.g. due to a misconfigured MCP server credential), `key` and/or `value` will be `nil`, and the subsequent `nil.(string)` assertion panics.

This function is called directly from `HandleToolCall` in the same file, which is the per-request handler for MCP tool calls with `fixed-query-token` credential type. There is no `recover` in the call chain, so the panic crashes the tool call and potentially the Envoy worker.

### Expected behavior

The function should check the `ok` flag and return an error (or skip the credential injection) instead of panicking when the credential map is missing required keys.

### Actual behavior

The bare assertion panics when credential keys are missing, crashing the MCP tool call.

### Version

wc4440222:
## I. What does this PR do

Fixes a panic in the `FixedQueryToken` function of `plugins/golang-filter/mcp-server/registry/remote.go` that occurs when the credential map is missing the expected `key` or `value` entries.

### Root Cause

The `FixedQueryToken` function discards the `ok` flag from map lookups:
```go
key, _ := cred.Credentials[FIX_QUERY_TOKEN_KEY]
value, _ := cred.Credentials[FIX_QUERY_TOKEN_VALUE]
h.Query[key.(string)] = value.(string)
```

If `cred.Credentials` does not contain the `"key"` or `"value"` entries (e.g. due to a misconfigured MCP server credential), `key` and/or `value` will be `nil`, and the subsequent `nil.(string)` assertion panics.

This function is called directly from `HandleToolCall` in the same file, which is the per-request handler for MCP tool calls with `fixed-query-token` credential type. There is no `recover` in the call chain, so the panic crashes the tool call.

### Fix

1. Changed `FixedQueryToken` signature to return `error` instead of being void
2. Added `ok` flag checks on both map lookups with descriptive error messages
3. Added type assertion checks (`key.(string)` / `value.(string)`) with comma-ok pattern
4. Updated `getHttpCredentialHandle` return type to match
5. Updated `HandleToolCall` to propagate the error from `credentialHandle`

## II. Fixes Issue

Fixes #4378

## III. Why no test cases

The MCP server registry code requires a full Envoy golang-filter runtime environment with MCP protocol integration to test the tool call path. The fix is a straightforward defensive programming change (checking `ok` flags and using comma-ok pattern for type assertions).

## IV. How to verify

- `cd plugins/golang-filter/mcp-server/registry && go build .` — compiles successfully

- This is a golang-filter plugin (not wasm-go), so the panic is not recovered by the wasm-go runtime — it directly crashes the Envoy worker
- The fix changes the function signature of `FixedQueryToken` and `getHttpCredentialHandle` to return `error`, which is a breaking change for any external callers — but since these are internal functions in the mcp-server package, there should be no external callers
- The error messages include the specific credential key that is missing, making it easy to diagnose configuration issues
- Same defensive pattern as approved in #4226 (WAF plugin) and #4224 (ai-load-balancer), but applied to a golang-filter instead of wasm-go

- [x] Regular modification scenario: AI analyzed the FixedQueryToken credential handling flow to identify the discarded ok flag and bare type assertions

### DIFF
--- a/plugins/golang-filter/mcp-server/registry/remote.go
+++ b/plugins/golang-filter/mcp-server/registry/remote.go
@@ -23,7 +23,7 @@ const (
 	DEFAULT_HTTP_PATH     = "/"
 )
 
-func getHttpCredentialHandle(name string) (func(*CredentialInfo, *HttpRemoteCallHandle), error) {
+func getHttpCredentialHandle(name string) (func(*CredentialInfo, *HttpRemoteCallHandle) error, error) {
 	if name == "fixed-query-token" {
 		return FixedQueryToken, nil
 	}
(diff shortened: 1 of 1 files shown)


johnlanni (approved):
Reviewed exact head `275412ab9e3ab129dc165b5572daa92fd89e4af0`.

This is a real config-driven panic path: registry credential values come from arbitrary JSON, so missing or non-string `key`/`value` entries can reach the assertions. The fix validates both fields before mutating the request and propagates a per-call JSON-RPC error. It is fail-closed, does not partially mutate `h.Query`, and does not expose credential values in the error.

The handler signature change is internal and all callers are updated. Build, focused tests, vet, and formatting checks pass. LGTM.

A small unit test for this pure function would still be useful, but it is not a semantic blocker.

PR #4383, opened 2026-08-08
bugfix: fix unsafe type assertions in certmgr OnEvent that can panic certmagic callback

### LINKED ISSUE #4382: Unsafe type assertions in certmgr OnEvent can panic certmagic callback
## Description

The `OnEvent` function in `pkg/cert/certmgr.go` uses bare type assertions on map lookups without checking key existence, causing a panic when the certmagic event map is missing expected keys.

### Affected file

`pkg/cert/certmgr.go` — `OnEvent` (~line 217-227):
```go
domain := data["identifier"].(string)
isRenew := data["renewal"].(bool)
privateKeyPath := data["private_key_path"].(string)
certificatePath := data["certificate_path"].(string)
```

### Root Cause

All four assertions assume the `data` map (from certmagic's event callback) always contains these keys with the correct types. If certmagic changes its event payload structure or a key is missing (e.g. during error/fallback events), `nil.(string)` or `nil.(bool)` panics the reconcile goroutine.

Additionally, the `Storage.Load` errors on the following lines are silently overwritten — if the first `Load` fails, `privateKey` may be nil but the code continues to `parseCertsFromPEMBundle` which could also fail.

### Expected behavior

The code should use comma-ok pattern for map lookups and check errors before proceeding.

### Version

wc4440222:
## I. What does this PR do

Fixes unsafe type assertions in the `OnEvent` function of `pkg/cert/certmgr.go` that can cause a panic when the certmagic event map is missing expected keys or has unexpected types.

### Root Cause

The `OnEvent` callback uses bare type assertions on `data` map lookups:
```go
domain := data["identifier"].(string)
isRenew := data["renewal"].(bool)
privateKeyPath := data["private_key_path"].(string)
certificatePath := data["certificate_path"].(string)
```

If certmagic changes its event payload structure or a key is missing, `nil.(string)` or `nil.(bool)` panics the reconcile goroutine.

Additionally, `Storage.Load` errors were silently overwritten — the first `Load` error was lost when the second `Load` was called.

### Fix

1. Replaced all 4 bare type assertions with comma-ok pattern
2. `identifier`, `private_key_path`, `certificate_path` — return nil (skip event) if missing with error log
3. `renewal` — default to `false` if missing (non-critical field)
4. Added error checks on both `Storage.Load` calls before proceeding

## II. Fixes Issue

Fixes #4382

## III. Why no test cases

The `OnEvent` callback is invoked by certmagic during certificate lifecycle events, requiring a full certmagic + K8s integration environment. The fix is a defensive programming change (comma-ok pattern + error checking). Syntax verified with `gofmt -e`.

## IV. How to verify

- `gofmt -e pkg/cert/certmgr.go` — syntax OK
- Full compilation: `go build ./pkg/cert/` (requires Istio submodules)

- The `renewal` field defaults to `false` when missing because it is non-critical — it only affects whether the update is logged as a renewal vs. a new cert
- The `identifier`, `private_key_path`, and `certificate_path` fields are critical — if any is missing, the event is logged and skipped (return nil) rather than panicking
- Added error checks on `Storage.Load` calls that were previously silently overwritten

- [x] Regular modification scenario: AI analyzed the OnEvent callback flow to identify all unsafe type assertions and missing error checks

### DIFF
--- a/pkg/cert/certmgr.go
+++ b/pkg/cert/certmgr.go
@@ -238,10 +238,12 @@ func (s *CertMgr) OnEvent(ctx context.Context, event string, data map[string]any
 		}
 		privateKey, err := s.cfg.Storage.Load(context.Background(), privateKeyPath)
 		if err != nil {
+			CertLog.Errorf("failed to load private key for domain %s from path %s: %v", domain, privateKeyPath, err)
 			return err
 		}
 		certificate, err := s.cfg.Storage.Load(context.Background(), certificatePath)
 		if err != nil {
+			CertLog.Errorf("failed to load certificate for domain %s from path %s: %v", domain, certificatePath, err)
 			return err
 		}
 		certChain, err := parseCertsFromPEMBundle(certificate)
--- a/pkg/cert/certmgr.go
+++ b/pkg/cert/certmgr.go
@@ -217,12 +217,33 @@ func (s *CertMgr) OnEvent(ctx context.Context, event string, data map[string]any
 	*/
 	if event == EventCertObtained {
 		// obtain certificate and update secret
-		domain := data["identifier"].(string)
-		isRenew := data["renewal"].(bool)
-		privateKeyPath := data["private_key_path"].(string)
-		certificatePath := data["certificate_path"].(string)
+		domain, ok := data["identifier"].(string)
+		if !ok {
+			CertLog.Errorf("missing or invalid identifier in cert event data: %+v", data)
+			return nil
+		}
+		isRenew, ok := data["renewal"].(bool)
+		if !ok {
+			isRenew = false
+		}
+		privateKeyPath, ok := data["private_key_path"].(string)
+		if !ok {
+			CertLog.Errorf("missing or invalid private_key_path in cert event data: %+v", data)
+			return nil
+		}
+		certificatePath, ok := data["certificate_path"].(string)
+		if !ok {
+			CertLog.Errorf("missing or invalid certificate_path in cert event data: %+v", data)
+			return nil
+		}
 		privateKey, err := s.cfg.Storage.Load(context.Background(), privateKeyPath)
+		if err != nil {
+			return err
+		}
 		certificate, err := s.cfg.Storage.Load(context.Background(), certificatePath)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


johnlanni (commented):
Reviewed exact head `fc000852cb77546956299de99fc3a006351d1f2a`.

The type-assertion guards are safe defensive hardening, and checking both `Storage.Load` errors fixes a real issue: the previous code could continue toward a Secret update with missing key/certificate data.

One small observability improvement is needed before I would approve: please log the two `Storage.Load` failures before returning them, including the domain and which artifact failed, without logging certificate or key contents. `certmagic.Config.emit` discards the callback's returned error, so the current `return err` paths can fail silently and make certificate renewal failures difficult to diagnose.

For example, add an `Errorf` before each return for:
- private key load failure;
- certificate load failure.

A focused in-package unit test is also feasible for missing event fields and storage load failures; it does not require a full Kubernetes environment unless exercising `secretMgr.Update`.

johnlanni (approved):
Reviewed exact head `fc000852cb77546956299de99fc3a006351d1f2a` again against the pinned certmagic fork.

The assertion guards are safe defensive hardening, and the `Storage.Load` checks fix a real latent corruption bug: on the base code, a private-key load failure could still proceed to `secretMgr.Update` with an empty `tls.key`, potentially replacing a working renewal Secret. This head skips the update instead and is strictly safer for every traced failure permutation.

The unlogged `Storage.Load` returns are a worthwhile non-blocking observability follow-up because the two `cert_obtained` emit sites discard callback errors. However, this silent-return pattern already exists in the adjacent base code, and it is not a regression introduced by this PR. Please consider adding `CertLog.Errorf` and focused unit tests in a follow-up.

LGTM.

johnlanni (approved):
Re-reviewed exact head `a37d9788426e286d7f479046cf122d1abdc43b76`.

The only change since the previously approved revision adds the requested `CertLog.Errorf` calls before the private-key and certificate `Storage.Load` error returns. The messages identify the domain and failed artifact without logging key or certificate contents. This closes the observability follow-up without changing the safer failure semantics.

LGTM.

PR #4637, opened 2026-09-02
fix(ext-auth): avoid double-forward on fail-open call dispatch errors

### LINKED ISSUE #4622: bug: ext-auth double-forwards requests when ext-auth server call fails synchronously with failure_mode_allow
- [x] I have searched the [issues](https://github.com/higress-group/higress/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description

When the ext-auth plugin's call to the ext-auth server (`DispatchHttpCall`) fails **synchronously** and `failure_mode_allow: true` is configured, the original downstream request is continued **twice** and can reach the application backend as two concurrent requests under the same parent request ID.

This issue is filed from the analysis of #4614, where the behavior was observed in production with the `ext-auth:2.0.0` plugin image. The mechanism has been verified against the current `main` branch (commit `c24b0036`) and is **pre-existing** — it is not introduced or exposed by #4560, which only added header-presence match rules.

Severity framing: this is a **correctness/reliability defect, not an authentication bypass**. The fail-closed path is unaffected, and fail-open is an operator-selected mode. However, when fail-open is selected, a single downstream request can be delivered to the backend twice, which is dangerous for non-idempotent operations such as LLM inference, image/video generation, and payment-related calls.

### Ⅱ. Describe what happened

In `checkExtAuth` (`plugins/wasm-go/extensions/ext-auth/main.go`), the ext-auth server is called at [main.go:122-139](https://github.com/higress-group/higress/blob/c24b00360e9b46f6c3bba7cdc7d2fed9b4e777ce/plugins/wasm-go/extensions/ext-auth/main.go#L122-L139). When `DispatchHttpCall` fails synchronously — for example when the ext-auth service's Envoy cluster is missing from the gateway configuration, returning `error status returned by host: bad argument` — `Client.Call` returns a non-nil error, the asynchronous callback never runs, and the request was never paused (the `return pauseAction` at [main.go:147](https://github.com/higress-group/higress/blob/c24b00360e9b46f6c3bba7cdc7d2fed9b4e777ce/plugins/wasm-go/extensions/ext-auth/main.go#L147) was never reached).

The synchronous error branch then does **both** of the following:

```go
// plugins/wasm-go/extensions/ext-auth/main.go:141-146
if err != nil {
    log.Errorf("failed to call ext auth server: %v", err)
    // Since the handling logic for call errors and HTTP status code 500 is the same, we directly use 500 here.
    callExtAuthServerErrorHandler(cfg, http.StatusInternalServerError, nil, nil)
    return types.ActionContinue
}
```

`callExtAuthServerErrorHandler` reuses the **asynchronous** error handler, which resumes the (never-paused) request when fail-open is enabled:

```go
// plugins/wasm-go/extensions/ext-auth/main.go:199-206
func callExtAuthServerErrorHandler(config config.ExtAuthConfig, statusCode int, extAuthRespHeaders http.Header, responseBody []byte) {
	if statusCode >= http.StatusInternalServerError && config.FailureModeAllow {
		if config.FailureModeAllowHeaderAdd {
			_ = proxywasm.ReplaceHttpRequestHeader(HeaderFailureModeAllow, "true")
		}
		proxywasm.ResumeHttpRequest()   // main.go:204 — resumes a request that was never paused
		return
	}
```

So on the synchronous failure path with `failure_mode_allow: true`:

1. `callExtAuthServerErrorHandler` is invoked in-callback ([main.go:144](https://github.com/higress-group/higress/blob/c24b00360e9b46f6c3bba7cdc7d2fed9b4e777ce/plugins/wasm-go/extensions/ext-auth/main.go#L144)) and executes `proxywasm.ResumeHttpRequest()` ([main.go:204](https://github.com/higress-group/higress/blob/c24b00360e9b46f6c3bba7cdc7d2fed9b4e777ce/plugins/wasm-go/extensions/ext-auth/main.go#L204)) — the first continue;
2. control returns to `checkExtAuth`, which returns `types.ActionContinue` ([main.go:145](https://github.com/higress-group/higress/blob/c24b00360e9b46f6c3bba7cdc7d2fed9b4e777ce/plugins/wasm-go/extensions/ext-auth/main.go#L145)) — the second continue.

Both entry points are affected because both call `checkExtAuth`: the header phase ([main.go:80](https://github.com/higress-group/higress/blob/c24b00360e9b46f6c3bba7cdc7d2fed9b4e777ce/plugins/wasm-go/extensions/ext-auth/main.go#L80)) and the body phase ([main.go:103](https://github.com/higress-group/higress/blob/c24b00360e9b46f6c3bba7cdc7d2fed9b4e777ce/plugins/wasm-go/extensions/ext-auth/main.go#L103)).

As observed in #4614, the backend received two requests sharing one gateway parent request ID, with two different backend request IDs, two different upstream TCP source ports, the same start timestamp, and diverging statuses (e.g. HTTP 200 and HTTP 400/500) — the two copies also race for the downstream request body.

The asynchronous paths are correct: an async 5xx with fail-open resumes exactly once via the handler; an async successful authorization resumes exactly once at [main.go:137](https://github.com/higress-group/higress/blob/c24b00360e9b46f6c3bba7cdc7d2fed9b4e777ce/plugins/wasm-go/extensions/ext-auth/main.go#L137); fail-closed sends the local error response via `util.SendResponse` ([main.go:224](https://github.com/higress-group/higress/blob/c24b00360e9b46f6c3bba7cdc7d2fed9b4e777ce/plugins/wasm-go/extensions/ext-auth/main.go#L224)) and does not forward.

The request should be continued exactly once in every fail-open scenario:

- synchronous `DispatchHttpCall` error: return `ActionContinue` **without** calling `ResumeHttpRequest`;
- asynchronous ext-auth HTTP 5xx: call `ResumeHttpRequest` exactly once;
- successful authorization response: call `ResumeHttpRequest` exactly once;
- fail-closed mode: send the configured local error response and do not forward the request to the application backend.

1. Enable the ext-auth plugin for a test route (current `main` or `ext-auth:2.0.0`).
2. Set `failure_mode_allow: true` (optionally `status_on_error: 503`).
3. Configure `http_service.endpoint.service_name` so that the corresponding Envoy cluster is absent (e.g. a service that is not pushed to the gateway), making `DispatchHttpCall` fail synchronously with `bad argument`.
4. Send a POST request with a unique `X-Request-Id` and a JSON body.
5. Confirm the gateway log shows the synchronous error `failed to call ext auth server: error status returned by host: bad argument`.
6. Count the requests reaching the backend for the same `X-Request-Id`.

Expected backend request count: `1`. Observed backend request count: `2` (racing for the same request body).

Lineage, verified with `git log -S` on `plugins/wasm-go/extensions/ext-auth/main.go`:

- #1103 introduced the ext-auth plugin. The synchronous error path already returned `types.ActionContinue` after calling the error handler, but the fail-open branch of the handler did **not** call `ResumeHttpRequest`, so the synchronous path continued exactly once.
- #1694 reworked the match rules and kept this structure (synchronous error → shared error handler → `ActionContinue`; fail-open branch without a resume call).
- #1705 ("fix: ext-auth crash bugfix") removed the `defer proxywasm.ResumeHttpRequest()` from the asynchronous callback and added `proxywasm.ResumeHttpRequest()` to the fail-open branch of `callExtAuthServerErrorHandler` (now [main.go:204](https://github.com/higress-group/higress/blob/c24b00360e9b46f6c3bba7cdc7d2fed9b4e777ce/plugins/wasm-go/extensions/ext-auth/main.go#L204)). Because the synchronous error branch reuses this handler **and** returns `types.ActionContinue`, this is the commit that introduced the double-continue on the synchronous failure path.
- #4560 only added header-presence match rules ([main.go:58-99](https://github.com/higress-group/higress/blob/c24b00360e9b46f6c3bba7cdc7d2fed9b4e777ce/plugins/wasm-go/extensions/ext-auth/main.go#L58-L99)); it did not touch `checkExtAuth` or `callExtAuthServerErrorHandler`, so it neither introduced nor exposed this defect.

Suggested fix: in the synchronous error branch ([main.go:141-146](https://github.com/higress-group/higress/blob/c24b00360e9b46f6c3bba7cdc7d2fed9b4e777ce/plugins/wasm-go/extensions/ext-auth/main.go#L141-L146)), do not reuse the resuming handler:

- fail-open: optionally add the `x-envoy-auth-failure-mode-allowed` header (respecting `failure_mode_allow_header_add`) and return `types.ActionContinue`, without calling `ResumeHttpRequest`;
- fail-closed: send the local error response (`status_on_error`) and return.

The asynchronous path stays unchanged. Equivalently, `callExtAuthServerErrorHandler` can take a `requestPaused` flag (as suggested in #4614) and only resume when the request was actually paused: the synchronous `Client.Call` error path passes `requestPaused=false`, the asynchronous response callback passes `requestPaused=true`.

Suggested regression coverage (from #4614):

1. synchronous `BadArgument` + fail-open: returned action is `ActionContinue`, resume count is 0, backend request count is 1;
2. asynchronous HTTP 503 + fail-open: initial action pauses the request, resume count is 1, backend request count is 1;
3. successful authorization: resume count is 1;
4. fail-closed: backend request count is 0.

### Ⅵ. Environment

- Higress version: current `main` (verified at commit `c24b0036`); first observed with plugin image `ext-auth:2.0.0`
- Component: `plugins/wasm-go/extensions/ext-auth` (wasm-go plugin)
- Deployment: Kubernetes
- OS: Linux containers

References: #4614 (original report and reproduction), #1103 / #1694 / #1705 (lineage), #4560 (confirmed unrelated).

yx9o:
## I. Summary

Fix ext-auth fail-open handling when the outbound authorization call fails before the request is paused.

Previously, the synchronous `Client.Call` error path reused the same failure handler as asynchronous callout responses. With `failure_mode_allow` enabled, that handler always called `ResumeHttpRequest()`. For synchronous call dispatch errors, the plugin returns `ActionContinue` and the request was never paused, so resuming it can double-forward the request.

This change tracks whether the request has actually been paused before resuming it. Asynchronous fail-open callout responses still resume paused requests. Synchronous call dispatch failures only add the configured failure-mode header and continue.

No configuration or compatibility migration is required.

## II. Related issue

Fixes #4622

## III. Change type

- [x] Bug fix
- [ ] Feature or enhancement
- [ ] Documentation or configuration only
- [x] Refactoring, tests, or maintenance

## IV. Agent participation and issue-spec gate

- [ ] **No agent participation:** This PR is human-only.
- [ ] **Trivial agent-use exception:** Agent use was limited to spelling,
      punctuation, whitespace, or formatting with no substantive choice or
      behavioral effect; the explanation is below.
- [ ] **Verified maintainer/administrator exception:** Material agent
      participation occurred, and authenticated `gh` verification confirmed
      a `role_name` of `maintain` or `admin` on `higress-group/higress`, and the
      verified login matches this PR's GitHub author; record the required
      evidence and rationale below.
- [x] **Material agent participation:** An AI or coding agent materially
      participated in analysis, design, implementation, testing, or PR
      preparation.

For material participation that does not use the verified maintainer/administrator
exception, check each timed obligation:

- [ ] **Before implementation began:** A Higress maintainer had approved the
      Proposal and Design Issues, and authorized implementation TASKs linked the
      work to the applicable SPECs.
- [ ] **Before verification began:** The Design contained a concrete
      Verification Plan.
- [ ] **Before requesting maintainer review or acceptance:** Applicable
      implementation and verification TASKs were `done` with exact commands,
      results, evidence links, and hashes.

Then choose exactly one overall gate status:

- [ ] Complete: all three timed obligations above were satisfied.
- [x] Noncompliant: one or more obligations were not satisfied; explain below.
- [ ] Verified exception: material agent participation used the
      verified-maintainer/administrator exception; provide the required evidence
      and rationale below.
- [ ] N/A because the PR is human-only or qualifies for the trivial exception.

**Gate-status explanation (or N/A):**

Material agent participation occurred before approved Proposal/Design/TASK issue-spec artifacts were available. No retroactive approval is claimed.

**Trivial-exception explanation (or N/A):**

N/A.

**Verified maintainer/administrator exception evidence (or N/A):**

N/A.

**Approved Proposal Issue URL (or N/A with explanation):**

N/A. No approved Proposal Issue was available before implementation.

**Approved Design Issue URL (or N/A with explanation):**

N/A. No approved Design Issue was available before implementation.

**Maintainer approval link(s) (or N/A with explanation):**

N/A. No maintainer approval link is available for the issue-spec gate.

**Authorized implementation TASK link(s) (or N/A with explanation):**

N/A. No authorized implementation TASK was available before implementation.

**Verification Plan and verification TASK/evidence link(s) (or N/A with explanation):**

N/A. Verification was performed locally with Go tests, but no approved issue-spec Verification Plan or verification TASK is available.

## V. Testing and verification

**Test and deterministic rerun commands (or N/A with explanation):**

```text
cd plugins/wasm-go/extensions/ext-auth
GOTOOLCHAIN=auto GOPROXY=https://goproxy.cn,direct GOSUMDB=sum.golang.org go test -count=1 -run 'TestCallExtAuthServerErrorHandler|TestOnHttpRequestHeaders/failure_mode_allow' ./...
GOTOOLCHAIN=auto GOPROXY=https://goproxy.cn,direct GOSUMDB=sum.golang.org go test -count=1 ./...

cd ../../../../
git diff --check

### DIFF
--- a/plugins/wasm-go/extensions/ext-auth/main.go
+++ b/plugins/wasm-go/extensions/ext-auth/main.go
@@ -45,6 +45,12 @@ const (
 	HeaderFailureModeAllow = "x-envoy-auth-failure-mode-allowed"
 )
 
+var (
+	replaceHttpRequestHeader = proxywasm.ReplaceHttpRequestHeader
+	resumeHttpRequest        = proxywasm.ResumeHttpRequest
+	sendResponse             = util.SendResponse
+)
+
 // Currently, x-forwarded-xxx headers only apply for forward_auth.
 const (
 	HeaderOriginalMethod   = "x-original-method"
@@ -123,7 +129,8 @@ func checkExtAuth(ctx wrapper.HttpContext, cfg config.ExtAuthConfig, body []byte
 		func(statusCode int, responseHeaders http.Header, responseBody []byte) {
 			if statusCode != http.StatusOK {
 				log.Errorf("failed to call ext auth server, status: %d", statusCode)
-				callExtAuthServerErrorHandler(cfg, statusCode, responseHeaders, responseBody)
+				// The callback only runs after the request has been paused for the external auth response.
+				callExtAuthServerErrorHandler(cfg, statusCode, responseHeaders, responseBody, true)
 				return
 			}
 
@@ -141,7 +148,8 @@ func checkExtAuth(ctx wrapper.HttpContext, cfg config.ExtAuthConfig, body []byte
 	if err != nil {
 		log.Errorf("failed to call ext auth server: %v", err)
 		// Since the handling logic for call errors and HTTP status code 500 is the same, we directly use 500 here.
-		callExtAuthServerErrorHandler(cfg, http.StatusInternalServerError, nil, nil)
+		// Dispatch errors return ActionContinue below, so there is no paused request to resume.
+		callExtAuthServerErrorHandler(cfg, http.StatusInternalServerError, nil, nil, false)
 		return types.ActionContinue
 	}
 	return pauseAction
@@ -196,12 +204,14 @@ func buildExtAuthRequestHeaders(ctx wrapper.HttpContext, cfg config.ExtAuthConfi
 	return extAuthReqHeaders
 }
 
-func callExtAuthServerErrorHandler(config config.ExtAuthConfig, statusCode int, extAuthRespHeaders http.Header, responseBody []byte) {
+func callExtAuthServerErrorHandler(config config.ExtAuthConfig, statusCode int, extAuthRespHeaders http.Header, responseBody []byte, requestPaused bool) {
 	if statusCode >= http.StatusInternalServerError && config.FailureModeAllow {
 		if config.FailureModeAllowHeaderAdd {
-			_ = proxywasm.ReplaceHttpRequestHeader(HeaderFailureModeAllow, "true")
+			_ = replaceHttpRequestHeader(HeaderFailureModeAllow, "true")
 		}
-		proxywasm.ResumeHttpRequest()
+		if requestPaused {
+			resumeHttpRequest()
+		}
 		return
 	}
 
@@ -221,5 +231,5 @@ func callExtAuthServerErrorHandler(config config.ExtAuthConfig, statusCode int, 
 	if statusCode >= http.StatusInternalServerError {
 		statusToUse = int(config.StatusOnError)
 	}
-	_ = util.SendResponse(uint32(statusToUse), "ext-auth.unauthorized", respHeaders, responseBody)
+	_ = sendResponse(uint32(statusToUse), "ext-auth.unauthorized", respHeaders, responseBody)
 }
--- a/plugins/wasm-go/extensions/ext-auth/main_test.go
+++ b/plugins/wasm-go/extensions/ext-auth/main_test.go
@@ -16,8 +16,11 @@ package main
 
 import (
 	"encoding/json"
+	"net/http"
 	"strings"
 	"testing"
+
+	"ext-auth/config"
 
 	"github.com/higress-group/proxy-wasm-go-sdk/proxywasm/types"
 	"github.com/higress-group/wasm-go/pkg/test"
@@ -159,6 +162,25 @@ var failureModeConfig = func() json.RawMessage {
 	return data
 }()
 
+var failureModeDispatchErrorConfig = func() json.RawMessage {
+	data, _ := json.Marshal(map[string]interface{}{
+		"http_service": map[string]interface{}{
+			"endpoint_mode": "forward_auth",
+			"endpoint": map[string]interface{}{
+				"service_name":   "ext-auth.backend.svc.cluster.local",
+				"service_port":   8090,
+				"path":           "%zz",
+				"request_method": "POST",
+			},
+			"timeout": 1000,
+		},
+		"failure_mode_allow":            true,
+		"failure_mode_allow_header_add": true,
+		"status_on_error":               500,
+	})
+	return data
+}()
+
 // 测试配置：带 allowed_properties 的配置
 var allowedPropertiesConfig = func() json.RawMessage {
 	data, _ := json.Marshal(map[string]interface{}{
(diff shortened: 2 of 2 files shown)


johnlanni (changes requested):
Independent review verdict: **the fix is correct** — recommend merge once the following merge-gates are cleared.

**Mechanism independently reproduced by code reading** (not from the PR description): pre-patch, a synchronous `HttpCall` failure (e.g. `url.Parse` rejection or `DispatchHttpCall` returning an error because the ext-auth cluster is missing) means the callout ID was never assigned, so the registered callback can never fire; yet `callExtAuthServerErrorHandler` unconditionally called `proxywasm.ResumeHttpRequest()` on that path, while `checkExtAuth` also returned `types.ActionContinue` — two continues for one downstream request. The submitted fix gates on whether the request was actually paused, which closes exactly that path with no tracked state to desync (call-site-derived constant). The fail-closed path, async fail-open resume, and 200-accept flow are behaviorally unchanged.

Two small non-blocking asks:
1. A one-line comment at the two boolean sites explaining why the sync-error path is `false` (swapping the literals would pass every test today; a comment locks the semantics in).
2. Please drop the stale HTML comment in the PR body claiming #4622 "appears unrelated" — it is exactly this bug.

**Merge-gates remaining:**
- **DCO**: the commit lacks `Signed-off-by` — please amend with `git commit -s --amend` and force-push (or add a new signed commit).
- **CI**: the changed-plugin and conformance jobs are still pending with the branch behind base — rebase onto latest main once DCO is fixed.
- **Governance**: the PR self-declares the issue-spec gate as Noncompliant for material agent participation. Recording the maintainer waiver here explicitly: per the verified-maintainer/administrator exception policy, the maintainer accepts this submission given the incident-driven fix (#4622, mechanism maintainer-verified on 8/31) and the independent review above. No retroactive Proposal/Design is required for this change.

johnlanni (approved):
LGTM

PR #4689, opened 2026-09-11
feat: Add Qdrant as a vector search provider

anush008:
## Description

This PR adds Qdrant as a vector backend for Higress RAG and tool-search.

[Qdrant](https://qdrant.tech/) is an open-source vector search engine built for high-performance and massive scale.

Until now both MCP servers only worked with Milvus. If you already run Qdrant, you had to stand up Milvus too. After this, Qdrant will be a supported alternative.

Milvus continues to be the default. No change for existing users.

### Testing

I've unit tested this integration against a local Qdrant instance.

### Setup

You can run Qdrant with 

```bash
docker run -p 6333:6333 -p 6334:6334 qdrant/qdrant
```

The dashboard is accessible at http://localhost:6333/dashboard.

In the RAG server vectordb block, we can then set:

```yaml
vectordb:
  provider: qdrant
  host: localhost
  port: 6334
  collection: rag_collection
```

In the tool-search vector block:

```yaml
vector:
  type: qdrant
  host: localhost
  port: 6334
  tableName: apig_mcp_tools
```

We using v1.15.2 of the Qdrant client on purpose. Newer clients want Go 1.24+ and would bump the whole `golang-filter` module.


plugins/golang-filter/mcp-server/servers/tool-search/milvus.go
anush008, line 1: `tool-search` used to own its own Milvus client and reject every `vector.type` except `milvus`.
It now goes through the RAG vector factory. `vector.type` can be `milvus` or `qdrant`.

PR #4697, opened 2026-09-14
feat(ingress): make the ECDH curves of proxied upstreams configurable

ThxCode-Chen:
> Rebased on review feedback: the option is now a per-registry field
> (`McpBridge.spec.registries[].ecdhCurves`, next to `sni`) instead of a global
> `higress-config` item.

## I. Summary

A service registered in `McpBridge` and bound to a proxy server
(`spec.registries[].proxyName`) gets its cluster rebuilt in
`constructProxyEnvoyFilters()`: the cluster is replaced by a `STATIC` one pointing at the local
proxy listener, and its `transport_socket` is rebuilt carrying only the SNI of the upstream.

Everything Istio would derive from the `DestinationRule` / mesh configuration for that cluster is
therefore dropped — including `meshConfig.tlsDefaults.ecdhCurves` — and Envoy falls back to the
defaults of its linked TLS library, which are `X25519` and `P-256` for non-FIPS BoringSSL builds
(see `TlsParameters.ecdh_curves` in `envoy/extensions/transport_sockets/tls/v3/common.proto`).

An upstream that only presents a certificate on another curve is then unreachable: under TLS 1.2
the server must not use a certificate whose curve is missing from the client's `supported_groups`
(RFC 8422, §5.1.1), so it aborts the handshake and the gateway returns 503 with
`SSLV3_ALERT_HANDSHAKE_FAILURE` / `HANDSHAKE_FAILURE_ON_CLIENT_HELLO`.

This PR adds an optional `ecdhCurves` field to `McpBridge.spec.registries[]`, next to the existing
`sni` field, so the curves can be declared per upstream:

```yaml
spec:
  registries:
    - name: provider-1
      type: dns
      domain: api.example.com
      port: 443
      protocol: https
      proxyName: provider-1-proxy
      ecdhCurves: ["X25519", "P-256", "P-384"]   # omitted = current behavior
```

* unset → the generated TLS context is left exactly as before (no behavior change);
* names that Envoy cannot use are dropped with a warning (see §VI, they would otherwise make the
  transport socket fail to initialize and reject the whole cluster configuration);
* the field only affects registries bound to a proxy server, which is the scope of the affected
  code path.

Compatibility: additive CRD field; older control planes ignore it.

## II. Related issue

N/A — no issue was filed for this; the failure was diagnosed from a production 503 (see §V).

## III. Change type

- [ ] Bug fix
- [x] Feature or enhancement
- [ ] Documentation or configuration only
- [ ] Refactoring, tests, or maintenance

## IV. Agent participation and issue-spec gate

- [ ] No agent participation
- [ ] Trivial agent-use exception
- [ ] Verified maintainer/administrator exception
- [x] **Material agent participation**

For material participation that does not use the verified maintainer/administrator exception,
check each timed obligation:

- [ ] Before implementation began: a Higress maintainer had approved the Proposal and Design
      Issues, and authorized implementation TASKs linked the work to the applicable SPECs.
- [ ] Before verification began: the Design contained a concrete Verification Plan.
- [x] Before requesting maintainer review or acceptance: the applicable implementation and
      verification evidence is recorded in §V (unit tests plus the reproduction below).

Overall gate status:

- [ ] Complete
- [x] **Noncompliant** — see explanation
- [ ] Verified exception
- [ ] N/A because the PR is human-only or qualifies for the trivial exception.

**Gate-status explanation:**

The change was implemented before any Proposal/Design Issue existed, so the first two timed
obligations were not satisfied. The author is aware of the policy in
`docs/developers/agent-assisted-contributions.md` and is not claiming an exception: no authenticated
`gh` check was run, and the author's role on this repository is `read`, so the
verified-maintainer/administrator exception does not apply.

Requested action: because the change is a scoped, opt-in CRD field with no behavior change when
unset, the author asks the maintainers to waive the Proposal/Design requirement on the record and
evaluate the change on its merits (§VI lists the follow-up work that would address the root cause).

**Trivial-exception explanation (or N/A):** N/A — the change is material.

**Verified maintainer/administrator exception evidence (or N/A):** N/A — the exception is not used.

**Approved Proposal Issue URL (or N/A with explanation):** N/A — no Proposal Issue exists; see the
gate-status explanation.

**Approved Design Issue URL (or N/A with explanation):** N/A — no Design Issue exists; see the
gate-status explanation.

**Maintainer approval link(s) (or N/A with explanation):** N/A — implementation predates any
approval.

**Authorized implementation TASK link(s) (or N/A with explanation):** N/A — no issue-spec TASKs
exist.

**Verification Plan and verification TASK/evidence link(s) (or N/A with explanation):** N/A — no
Design Issue exists; the verification performed is recorded in §V.

## V. Testing and verification

**Test and deterministic rerun commands:**

```text
go build ./...
go test ./pkg/ingress/config/ ./pkg/ingress/kube/configmap/
# regeneration of the API artifacts (api/gen.sh requires buf + the istio codegen plugins):
cd api && buf generate --path networking
```

**Environment and configuration:**

```text
Go 1.26.0 (linux/amd64)
protoc-gen-go v1.31.0, buf v1.28.1, protoc-gen-golang-deepcopy / -jsonshim (istio.io/tools)
kubernetes control plane not required: the tests are unit tests of the generated EnvoyFilter
```

**Exact tested revision, version, image tag/digest, manifests, and values:**

```text
branch set_ecdh_curves, commit <to be filled in after the force-push>
```

**Baseline reproduction evidence for bug fixes (or N/A with explanation):**

Before the change, against an upstream that presents a single ECDSA P-384 certificate (host redacted):

```text
$ openssl s_client -connect <host>:443 -servername <host> -tls1_2 -curves X25519:P-256 -brief
139985965287232:error:14094410:SSL routines:ssl3_read_bytes:sslv3 alert handshake failure:
ssl/record/rec_layer_s3.c:1544:SSL alert number 40

$ openssl s_client -connect <host>:443 -servername <host> -tls1_2 -curves X25519:P-256:P-384 -brief
CONNECTION ESTABLISHED
Ciphersuite: ECDHE-ECDSA-AES256-GCM-SHA384
```

The corresponding ClientHello captured on the gateway's egress carries
`supported_groups = [X25519(0x001d), secp256r1(0x0017)]` (no `secp384r1`), and the upstream answers
with a fatal alert 40 while the handshake is still at the ClientHello. This is not an SNI problem:
an unknown SNI produces `unrecognized_name(112)`, not `handshake_failure(40)`.

**Fixed-version verification evidence for bug fixes (or N/A with explanation):**

Unit tests (same commit):

```text
$ go test ./pkg/ingress/config/ ./pkg/ingress/kube/configmap/
ok  	github.com/alibaba/higress/v2/pkg/ingress/config
ok  	github.com/alibaba/higress/v2/pkg/ingress/kube/configmap
```

`pkg/ingress/config/proxy_envoyfilter_test.go` asserts, for the proxy-bound registry:

* `ecdhCurves` unset → no `common_tls_context` is emitted, and the SNI plus the local proxy
  endpoint (`127.0.0.1:<listenerPort>`) are unchanged;
* `ecdhCurves: ["X25519","P-256","P-384","X25519Kyber768Draft00","bogus"]` → exactly
  `["X25519","P-256","P-384"]` is emitted, with the same SNI and endpoint assertions.

Live end-to-end verification on a running gateway (a request against an upstream with a P-384
certificate, with and without the new field) has not been performed; the author can run it in a
disposable cluster on request.

**Wasm source revision and SHA-256 (or N/A with explanation):** N/A — control-plane change only.

## VI. Special notes for reviewers

* Scope: only registries bound to a proxy server (`spec.registries[].proxyName`) are affected;
  everything else keeps the Istio-generated TLS context, where `meshConfig.tlsDefaults.ecdhCurves`
  continues to apply.
* The accepted curve names are intentionally narrower than `security.ValidECDHCurves`
  (`P-256`, `P-384`, `P-521`, `X25519`): names such as `X25519Kyber768Draft00` or
  `X25519MLKEM768` are not implemented by every TLS library, and an unsupported name makes the
  transport socket fail to initialize, which rejects the whole cluster configuration and would
  break every service behind the cluster — the exact failure this field exists to avoid.
* `X25519` is not FIPS approved; on a gateway built for FIPS 140-2 only `P-256` / `P-384` should be
  listed. This is noted in the API comment.
* Generated artifacts: `api/networking/v1/mcp_bridge.pb.go` was regenerated with
  `protoc-gen-go v1.31.0` + `buf v1.28.1`; the two CRD manifests were updated by hand because
  `cue-gen` is no longer published with `istio.io/tools` in this environment (the added schema
  entry follows the exact shape of the neighbouring `repeated string` fields). Please re-run
  `make gen-api` when convenient to confirm the artifacts match.
* Root cause / follow-up (not attempted in this PR): the real fix is for
  `constructProxyEnvoyFilters()` to merge with the Istio-generated TLS context instead of replacing
  `transport_socket` wholesale; that would let `meshConfig` / `DestinationRule` settings flow
  naturally and would make this field unnecessary. It requires changes in the internal control-plane
  TLS generation (the packet-level EnvoyFilter semantics of the rewritten cluster, and how the
  Istio-generated TLS context is produced) that the author is not familiar with, so the author
  suggests the maintainers own it. The analysis and reproduction recorded in §V/§VI are available as
  input, and the author will assess the size of the change and open a tracking issue separately
  rather than proposing a hand-rolled internal change. Two candidate directions, for reference only:
  (a) emit the relevant `meshConfig.tlsDefaults` values into the TLS context generated for proxied
  clusters; (b) stop replacing the socket wholesale, which first requires confirming at the Envoy
  level how a `MERGE` patch treats the `Any`-typed `typed_config`.

## VII. AI coding disclosure

**Tool(s) used:** Claude Code (Anthropic), driving an agent that read the repository, produced the
patch, and ran the build and unit tests locally.

### Prompts or instructions

Summarized and redacted (internal hostnames, addresses and credentials removed):

1. "An AI gateway route returns 503 with
   `upstream_transport_failure_reason: ... SSLV3_ALERT_HANDSHAKE_FAILURE |
   HANDSHAKE_FAILURE_ON_CLIENT_HELLO`; the upstream is reached through a proxy configured in
   `McpBridge`. Explain why."
2. "The failure is reproducible against the upstream with `openssl s_client -tls1_2 -curves
   X25519:P-256` (alert 40) and disappears when `P-384` is added to the client's curves. Add a way
   to configure the curves of the TLS context that Higress generates for proxy-bound registries."
3. Review feedback incorporated: "move `ecdhCurves` from the global `higress-config` ConfigMap to
   `McpBridge.spec.registries[]`, next to `sni`", narrow the accepted curve names, correct the
   comment that claimed a default was applied, note the X25519/FIPS caveat, and assert the
   SNI/endpoint invariants in the configured case as well.

### AI-assisted work summary

Key decisions and considerations:

* The mechanism (REMOVE+ADD cluster rewrite → dedicated `UpstreamTlsContext` carrying only the SNI
  → library default curves) was confirmed against a packet capture of the gateway's egress and
  against the generated configuration, before any code was written.
* The option is opt-in and per-registry: an unset field leaves the generated TLS context byte-for-
  byte as before, so no existing deployment changes behavior.
* Unsupported curve names are dropped with a warning rather than propagated, to avoid a transport
  socket initialization failure that would NACK the cluster configuration.
* Limitations: no live end-to-end run on a gateway was performed (see §V); the CRD manifests were
  updated by hand because `cue-gen` could not be installed in this environment.


johnlanni (changes requested):
# Review: request changes — engineering is solid; two changes requested (one is a design re-home)

First, the good news: the mechanism analysis in this PR is excellent and independently verified. The `REMOVE`+`ADD` cluster rewrite in `constructProxyEnvoyFilters()` (`ingress_config.go:1996-2073`) genuinely discards the Istio-generated cluster — the only place `meshConfig.tlsDefaults.ecdhCurves` and DestinationRule TLS settings are applied — and the rebuilt `UpstreamTlsContext` carries only the SNI. I also verified the curves land on the correct hop: the local proxy listener uses HTTP CONNECT tunneling (`%REQUESTED_SERVER_NAME%`), so the ClientHello reaches the real upstream and `ecdh_curves` on the rewritten cluster governs the actual handshake. The opt-in semantics, the drop-unsupported-names-with-warning behavior, and both test assertions are correct.

## Requested change 1: re-home the config from global ConfigMap to per-registry

Please move `ecdhCurves` from the `higress-config` ConfigMap to **`McpBridge.spec.registries[]`**, next to the existing `sni` field (`api/networking/v1/mcp_bridge.pb.go` RegistryConfig):

- **Scope match**: the affected surface is exactly the registries bound via `proxyName` — a per-upstream property. A global list forces operators into a union of all upstreams' needs, which weakens the handshake posture of every other proxied service (more curves = larger attack surface).
- **Precedent**: `sni` is precisely "per-registry upstream TLS parameter" — `ecdhCurves` belongs beside it, and the emission point (`ingress_config.go:2035-2041`) already reads registry fields one by one, so the code motion is small.
- **Coverage**: Ingress annotations would miss pure-McpBridge registrations (non-K8s domains); RegistryConfig covers both.
- Unset on a registry = current behavior (no `tls_params` emitted) — same opt-in guarantee, better granularity.

## Requested change 2: fill the PR template's agent-participation section

The commit carries `Co-Authored-By: Claude` (material agent participation), but the PR body skips §IV/§VII entirely. Per AGENTS.md, the issue-spec gate applies — and since the verified-maintainer exception requires `maintain`/`admin` (your current role on this repo is `read`), the normal path applies to this PR: please fill in §IV (choose the applicable participation checkbox), §VII (AI disclosure), and link any Proposal/Design issue. For a scoped, opt-in config knob like this one, a full Proposal is likely overkill — a maintainer can waive on the record once the disclosure is filled.

## Minor items (non-blocking, feel free to fold into the next push)

- `ingress_config.go:1912-1913`: the comment "the default set is then applied" is inaccurate — nothing is applied; the context is left untouched (the comment at `:1924-1925` is the correct one).
- `ingress_config.go:1932`: Istio's `ValidECDHCurves` accepts names BoringSSL may reject (e.g. `X25519Kyber768Draft00`); a rejected name fails transport-socket init → cluster NACK → 503, the exact failure this knob exists to prevent. Consider narrowing the accepted set or documenting the risk.
- `global.go:85-86`: worth noting X25519 is not FIPS-approved — the PR's own example list would fail on a FIPS gateway.
- `global.go:176-179`: `deepCopyGlobal` drops the new field (not functional today, but a trap for the next consumer).
- `proxy_envoyfilter_test.go`: SNI/endpoint preservation is only asserted in the unset case — add the configured case.
- Docs: `helm/core/values.yaml` / README don't mention the knob (fine to omit from defaults for opt-in, but worth documenting — adjust for the new location).

## Follow-up (not this PR)

The real root fix is for `constructProxyEnvoyFilters()` to **merge** with the Istio-generated TLS context instead of replacing `transport_socket` wholesale — then meshConfig/DestinationRule settings would flow naturally and this knob could eventually be retired. Would you like to open that issue, or shall we?

## A friendly note on AI-assisted contributions

To be clear: **AI-assisted contributions are welcome here** — a good number of recent changes in this repo were developed with AI assistance, and the quality bar this PR demonstrates (the mechanism analysis, the reproduction, the tests) is exactly what we hope for. The governance template (§IV/§VII) isn't friction against AI use — it exists so that the review can account for how a change was produced, and so agent-driven work gets the same review rigor as human-only work. Fill it in and you're all set; if anything in the process is unclear, ask away. We'd love to see more contributions from you.

PR #4702, opened 2026-09-15
docs: use block_regexp_urls in request_block regex example

89799969:
## What

In the Chinese README section for blocking regex-matched request URL paths, replace `block_exact_urls` with `block_regexp_urls`.

## Why

The sample uses the pattern `.*swagger.*`, which is a regular expression. In the same file, the configuration table documents `block_regexp_urls` as the regex field and `block_exact_urls` as exact match; the plugin source compares exact URLs with equality and regexes with a regex matcher. The current sample therefore describes a configuration that can never match.

## Agent participation

- Agent participation: yes (mechanical documentation/config-name correction; no code behavior change)
- issue-spec: not required for mechanical documentation corrections under the current policy
- AI coding disclosure: prepared with an AI coding assistance tool.

## Verification

- Cross-checked against the configuration table in the same README and `plugins/wasm-cpp/extensions/request_block/plugin.cc` field usage.


89799969:
Checking in — still relevant and ready for review. Happy to rebase or adjust if maintainers prefer.

PR #4743, opened 2026-09-22
fix(frontend-gray): handle 404 fallback without Content-Type

### LINKED ISSUE #4354: [frontend-gray] HTML fallback fails when 404 response lacks Content-Type
- [x] I have searched the issues and open pull requests of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description

The frontend-gray HTML fallback assumes every upstream 404 response contains a `content-type` header.

### Ⅱ. Describe what happened

When HTML fallback is configured and the upstream returns only `:status: 404`, the response hook indexes an empty header slice. The host reports an index-out-of-range panic, and the fallback response headers are not produced correctly.

The plugin should safely change the status to 200 and add `content-type: text/html`, even when that header was absent upstream.

### Ⅳ. How to reproduce it

1. Configure frontend-gray with a non-empty `html` fallback.
2. Process an index request.
3. Pass a response containing `:status: 404` and no `content-type`.
4. Observe the recovered panic and missing fallback header.

### Ⅴ. Environment

- Higress version: main at `f21b9fa50cbfc57ded4b8db31b31d169f2426b3b`
- Reproduced with the wasm-go plugin test host

maoruiqi-hub:
When an upstream returns 404 without Content-Type and frontend-gray has an HTML fallback configured, the plugin indexes an empty header slice and panics. The callback's recovery leaves the response at 404 instead of applying the fallback headers.

This replaces the two slice-index writes with whole-key assignments. The fallback sets status 200 and a single Content-Type: text/html whether the upstream header is absent or present. The existing Content-Length removal and body-buffering logic are unchanged. Only main.go and main_test.go are changed.

Fixes #4354

### Verification

- Added regression cases for a 404 without Content-Type and a 404 with an existing non-HTML Content-Type. They assert status 200 and exactly one text/html header; the missing-header case also asserts removal of Content-Length. Added an unchanged Content-Type assertion to the existing no-fallback case.
- The implementation run reproduced the missing-header failure on baseline a4248234 in both Go and compiled-Wasm modes: index-out-of-range panic and status remaining 404.
- Independently rebuilt the current Wasm artifact and reran `go test -run TestOnHttpResponseHeader -count=1 .`: PASS in Go and Wasm modes. Wasm SHA-256: `3250c16c6c786675fccaeef73c38e79a43f156dfaf4b546fbe67563ec5038477`.
- `go test ./... -count=1`: main and config packages PASS. On Windows, util/TestPrefixFileRewrite fails because filepath produces backslashes; util is unchanged by this PR. This is not reported as an all-green full-suite run.
- `git diff --check HEAD^ HEAD`: PASS.

Wasm build command (POSIX shell):
```sh
cd plugins/wasm-go/extensions/frontend-gray
GOOS=wasip1 GOARCH=wasm go build -buildmode=c-shared -o main.wasm .
go test -run TestOnHttpResponseHeader -count=1 -v .
```

The compiled Wasm tests use wazero, not a live Envoy gateway. Live Envoy verification has not been run. Duplicate upstream Content-Type values are collapsed by the whole-key assignment, but there is no dedicated duplicate-input regression test.

### Related work and AI disclosure

The earlier fix in #4363 is closed and unmerged; this PR addresses the still-open issue on a4248234 with the regression coverage above.

Implementation and its initial verification were assisted by ZCode (GLM); code review, independent test reruns, and PR preparation were assisted by OpenAI Codex. The instructions were to fix the missing Content-Type panic, preserve existing fallback behavior, add regression coverage, and review and submit this narrowly scoped fix. No approved Proposal/Design preceded implementation; this PR does not claim issue-spec gate compliance. Commits retain the implementation's AI attribution. No human Signed-off-by has been added by an agent.


maoruiqi-hub:
@johnlanni Could you help check or retrigger `yy-sensitive-scan` on this PR? On the current head `c71bf1d8b485061077002653af25d61cd62110d9`, its status is `error` with the description: "An internal error occurred during scan." There is no finding identifying sensitive content or a file/line to fix.

DCO and CLA are now passing, as are the build, plugin tests, conformance tests, and CodeQL checks. The Codecov comment confirms coverage of all modified coverable lines; its warnings concern a missing base coverage report and the repository's Codecov app setup.

I haven't changed the code in response to the scanner error because it provides no actionable finding. If the internal report identifies anything we need to address, please let me know.

